PR #9, opened 2021-12-04
port store v2 changes back

austbot:


thlorenz:
I restarted the `Token Metadata` workflow due to it failing for wrong reasons most likely, but the metaplex workflow failed due to a real issue afaik. Please have a look @austbot https://github.com/metaplex-foundation/metaplex-program-library/runs/4418502598?check_suite_focus=true

thlorenz (approved):
LGTM

PR #19, opened 2021-12-07
add security policy

aheckmann:


.github/SECURITY.md
```diff
@@ -0,0 +1,12 @@
+## Supported Versions
+
+The following versions are currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 1.x     | :white_check_mark: |
+| < 1.0   | :x:                |
```
aheckmann, line 8: @thlorenz what makes sense here?

thlorenz: Not sure ATM .. this is somewhat early for this. We probably will be releasing a `1.x` version of each contract once we encourage devs to use them,  so this makes sense in order to prevent _early_ security fixes coming in.

However we prolly also should indicate in the readme then that this is still a WIP somewhat.

thlorenz (approved):
For now this makes sense .. readme change could be a separate PR, so I'm fine with this.

.github/SECURITY.md
aheckmann: readme PR https://github.com/metaplex-foundation/metaplex-program-library/pull/24

PR #25, opened 2021-12-09
Allow is_mutable to be flipped false via update_metadata_accounts_v2

samuelvanderwaal:
Ported from [metaplex repo:](https://github.com/metaplex-foundation/metaplex/pull/984). 

> Currently there appears to be no way to set is_mutable to be false after an NFT metadata account has been created. Many NFTs, such as any created with default settings by candy_machine_cli, are created with the flag to true. It is desirable to have the ability to flip it to false at a later date, once any metadata issues are resolved or the community makes a decision to do so, etc.

I'll be looking at adding tests to this next before it gets merged. @thlorenz would like to discuss tests with you (it appears the tests for `token-metadata` are currently broken).

### DIFF
--- a/token-metadata/program/src/error.rs
+++ b/token-metadata/program/src/error.rs
@@ -304,6 +304,10 @@ pub enum MetadataError {
     // In the legacy system the reservation needs to be of size one for cpu limit reasons
     #[error("In the legacy system the reservation needs to be of size one for cpu limit reasons")]
     ReservationArrayShouldBeSizeOne,
+
+    /// Is Mutable can only be flipped to false
+    #[error("Is Mutable can only be flipped to false")]
+    IsMutableCanOnlyBeFlippedToFalse,
 }
 
 impl PrintProgramError for MetadataError {
--- a/token-metadata/program/src/instruction.rs
+++ b/token-metadata/program/src/instruction.rs
@@ -20,6 +20,16 @@ pub struct UpdateMetadataAccountArgs {
 
 #[repr(C)]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone)]
+/// Args for update call
+pub struct UpdateMetadataAccountArgsV2 {
+    pub data: Option<Data>,
+    pub update_authority: Option<Pubkey>,
+    pub primary_sale_happened: Option<bool>,
+    pub is_mutable: Option<bool>,
+}
+
+#[repr(C)]
+#[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone)]
 /// Args for create call
 pub struct CreateMetadataAccountArgs {
     /// Note that unique metadatas are disabled for now.
@@ -58,6 +68,11 @@ pub enum MetadataInstruction {
     ///   0. `[writable]` Metadata account
     ///   1. `[signer]` Update authority key
     UpdateMetadataAccount(UpdateMetadataAccountArgs),
+
+    /// Update a Metadata with is_mutable as a parameter
+    ///   0. `[writable]` Metadata account
+    ///   1. `[signer]` Update authority key
+    UpdateMetadataAccountV2(UpdateMetadataAccountArgsV2),
 
     /// Register a Metadata as a Master Edition V1, which means Editions can be minted.
     /// Henceforth, no further tokens will be mintable from this primary mint. Will throw an error if more than one
@@ -304,6 +319,33 @@ pub fn update_metadata_accounts(
             data,
             update_authority: new_update_authority,
             primary_sale_happened,
+        })
+        .try_to_vec()
+        .unwrap(),
+    }
+}
+
+// update metadata account v2 instruction
+pub fn update_metadata_accounts_v2(
+    program_id: Pubkey,
+    metadata_account: Pubkey,
+    update_authority: Pubkey,
+    new_update_authority: Option<Pubkey>,
+    data: Option<Data>,
+    primary_sale_happened: Option<bool>,
+    is_mutable: Option<bool>,
+) -> Instruction {
+    Instruction {
+        program_id,
+        accounts: vec![
+            AccountMeta::new(metadata_account, false),
+            AccountMeta::new_readonly(update_authority, true),
+        ],
+        data: MetadataInstruction::UpdateMetadataAccountV2(UpdateMetadataAccountArgsV2 {
+            data,
+            update_authority: new_update_authority,
+            primary_sale_happened,
+            is_mutable,
         })
         .try_to_vec()
         .unwrap(),
--- a/token-metadata/program/src/processor.rs
+++ b/token-metadata/program/src/processor.rs
@@ -61,6 +61,17 @@ pub fn process_instruction<'a>(
                 args.primary_sale_happened,
             )
         }
+        MetadataInstruction::UpdateMetadataAccountV2(args) => {
+            msg!("Instruction: Update Metadata Accounts");
+            process_update_metadata_accounts_v2(
+                program_id,
+                accounts,
+                args.data,
+                args.update_authority,
+                args.primary_sale_happened,
+                args.is_mutable,
+            )
+        }
         MetadataInstruction::DeprecatedCreateMasterEdition(args) => {
             msg!("Instruction: Deprecated Create Master Edition");
             process_deprecated_create_master_edition(program_id, accounts, args.max_supply)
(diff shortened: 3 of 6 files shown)


thlorenz:
I will have a look tomorrow afternoon (out until after lunch).

> I'll be looking at adding tests to this next before it gets merged. @thlorenz would like to discuss tests with you (it appears the tests for token-metadata are currently broken).

There are currently no token-metadata tests on master (I'm working on them on a separate branch).
What you maybe referring to is a `main` script under the `tests` folder which currently doesn't build as it is mainly here to inform writing integration tests driven by the contract's JS SDK.

However for your PR all checks passed so you're good for now.

Tests aren't needed until I finished the [amman](https://github.com/metaplex-foundation/amman) test utils that will help you with that. Also I'll merge in the tests I added to token-metadata so far to serve as a guidance for you.

thlorenz:
Just realizing you may not be aware that there are Rust tests for token-metadata already which you see passing [here](https://github.com/metaplex-foundation/metaplex-program-library/runs/4464698849?check_suite_focus=true).
They are defined [here](https://github.com/metaplex-foundation/metaplex-program-library/tree/master/token-metadata/program/tests) and you should add tests for your feature add here as well.

However as mentioned integration tests will have to come a bit later.

samuelvanderwaal:
> Just realizing you may not be aware that there are Rust tests for token-metadata already which you see passing [here](https://github.com/metaplex-foundation/metaplex-program-library/runs/4464698849?check_suite_focus=true). They are defined [here](https://github.com/metaplex-foundation/metaplex-program-library/tree/master/token-metadata/program/tests) and you should add tests for your feature add here as well.
> 
> However as mentioned integration tests will have to come a bit later.

Yeah I was looking at the `test` folder at the outer level with `main.rs` that doesn't build.

Not trying to rush you, just wanted to make sure I knew what I should be doing for tests. I'll add Rust tests for my new update function tomorrow and then we can add integration tests once you've got the framework all complete.

Thanks!

thlorenz (approved):
Awesome, this is looking good.
I like that you added tests for the invalid input cases as well.

Also thanks for fixing the `test/main.rs` as part of this.

Do we need to update the SDK next to be able to call the new instruction?
If so I should have the testing foundation by EOW so you could do this with an integration test included .. all in a separate PR though :)

samuelvanderwaal:
> Awesome, this is looking good. I like that you added tests for the invalid input cases as well.
> 
> Also thanks for fixing the `test/main.rs` as part of this.
> 
> Do we need to update the SDK next to be able to call the new instruction? If so I should have the testing foundation by EOW so you could do this with an integration test included .. all in a separate PR though :)

Yeah, I can look at doing that next week after your testing foundation is ready.

I'll let Austin review this PR and then will merge it if he doesn't find any blockers.

novaknole:
Hi folks,

This feature is extra handy. Currently, Updating `isMutable` from js library is not possible. But I went ahead and quickly add it, though, tx immediatelly gets failed with `BoschError` and couldn't detect why.

Is this contract already updated on metaplex for devnet and mainnet as well ? I tested it for devnet.

Would appreciate your inputs on this.

samuelvanderwaal:
> Hi folks,
> 
> This feature is extra handy. Currently, Updating `isMutable` from js library is not possible. But I went ahead and quickly add it, though, tx immediatelly gets failed with `BoschError` and couldn't detect why.
> 
> Is this contract already updated on metaplex for devnet and mainnet as well ? I tested it for devnet.
> 
> Would appreciate your inputs on this.

This fix should be deployed on devnet currently and will be deployed onto mainnet likely next week.

samuelvanderwaal:
> @samuelvanderwaal When was it deployed on `devnet` ? I created candy machine 2 days ago with devnet and passing `isMutable:0` still fails with `BoschError`.

I just tried this on devnet and it worked. How are you calling the handler?

truedandyman0617:
> > @samuelvanderwaal When was it deployed on `devnet` ? I created candy machine 2 days ago with devnet and passing `isMutable:0` still fails with `BoschError`.
> 
> It's definitely deployed to testnet because I just called the handler and it worked. The JS library is probably not updated yet, though.

If it has deployed on devnet already, perhaps it looks like made another issue. When I try to update metadata and set to isMutable: true without any extra setting, I get the error "Is Mutable can only be flipped to false".
And when I try to set this value to false in CLI code, it is updated without error but it sets to immutable.
I am not sure if it's not bug but in this case we can only update it one time.

samuelvanderwaal:
> > > @samuelvanderwaal When was it deployed on `devnet` ? I created candy machine 2 days ago with devnet and passing `isMutable:0` still fails with `BoschError`.
> > 
> > 
> > It's definitely deployed to testnet because I just called the handler and it worked. The JS library is probably not updated yet, though.
> 
> If it has deployed on devnet already, perhaps it looks like made another issue. When I try to update metadata and set to isMutable: true without any extra setting, I get the error "Is Mutable can only be flipped to false". And when I try to set this value to false in CLI code, it is updated without error but it sets to immutable. I am not sure if it's not bug but in this case we can only update it one time.

This is how it's supposed to work. You can't set `isMutable: true`. If it's already true, you don't need to set it, and if it's false, you can't set it to be mutable again. (Can only be flipped false.) You can set it to false but not back to mutable again. 

Let me know if that makes sense.

samuelvanderwaal:
The idea here is that an NFT being mutable: false (immutable) is a guarantee that the NFT can't be changed any more. Once you set an NFT to mutable: false, that's it, it can't be updated anymore and the owner has a guarantee that no one can mess with their NFT. If we allowed is mutable to be set back and forth there would be no point in having that field at all.

truedandyman0617:
When I use CLI, I don't set that flag but the updateMetadata function set it to true always in [Here](https://github.com/metaplex-foundation/metaplex/blob/master/js/packages/cli/src/commands/mint-nft.ts#L250).
And we should be able to set it to true anytime if current value is true.
How this could be an error?

samuelvanderwaal:
> When I use CLI, I don't set that flag but the updateMetadata function set it to true always in [Here](https://github.com/metaplex-foundation/metaplex/blob/master/js/packages/cli/src/commands/mint-nft.ts#L250). And we should be able to set it to true anytime if current value is true. How this could be an error?

I see, you're using the JS library. Can you open an issue on that repo describing the bug? This PR was for updating the *on-chain* Solana program.

truedandyman0617:
Well, I think current js library is ok since isMutable is optional and it can has the value True as default.
The problem is, when I try update Metadata for the mutable nft, the TokenMetadata program bans it with the error "Is Mutable can only be flipped to false". It looks like we can update Metadata only 1 time, after that NFT should be immutable always.
This is the problem now and that's why I opened issue in here.

samuelvanderwaal:
In the on-chain Solana program, `is_mutable` is an `Option` in the Rust code. This means you can pass in either a `None` value or `Some(true)` or `Some(false)`. If you pass in `None`, nothing happens and the value stays the same. If you pass in `Some(true)` you'll get an error because you're saying you want to change it to be `true` and you can't change it to be `true` because it can only be flipped to `false`. If you pass in `Some(false)` that will set it to `false` (or keep it that way if it's already `false`).

Unless you're trying to change the `is_mutable` value you should always pass in a `None`. Typically this is a `null` value in JS libraries. If our JS libraries do not allow passing in a `null` then that's a bug in the JS libraries, but as far as I can tell the on-chain program is still working as expected. 

If you can show me a specific test you're running where you think the on-chain program is not behaving correctly, that would be helpful. Otherwise, if this is an issue specifically with our JS library you can open an issue directly there.

@aheckmann pinging you for context because I think this is about how the JS library handles the `updateMetadataAccountsV2` command.

truedandyman0617:
> In the on-chain Solana program, `is_mutable` is an `Option` in the Rust code. This means you can pass in either a `None` value or `Some(true)` or `Some(false)`. If you pass in `None`, nothing happens and the value stays the same. If you pass in `Some(true)` you'll get an error because you're saying you want to change it to be `true` and you can't change it to be `true` because it can only be flipped to `false`. If you pass in `Some(false)` that will set it to `false` (or keep it that way if it's already `false`).
> 
> Unless you're trying to change the `is_mutable` value you should always pass in a `None`. Typically this is a `null` value in JS libraries. If our JS libraries do not allow passing in a `null` then that's a bug in the JS libraries, but as far as I can tell the on-chain program is still working as expected.
> 
> If you can show me a specific test you're running where you think the on-chain program is not behaving correctly, that would be helpful. Otherwise, if this is an issue specifically with our JS library you can open an issue directly there.
> 
> @aheckmann pinging you for context because I think this is about how the JS library handles the `updateMetadataAccountsV2` command.

I understood exactly, Yes it looks like JS library code issue. Thanks.

PR #27, opened 2021-12-09
Move nft-packs and fusion programs

VadimGrozinok:
## What ?
Move nft-packs and fusion programs to Metaplex-foundation repository from [Metaplex/smart-contracts](https://github.com/metaplex/smart-contracts)


thlorenz (changes requested):
I'm not sure if the anchor setup works this way. @austbot mentioned something about it being a bit tricky.
Also could you please add a github workflow to check that it builds and that tests pass, similar to [this](https://github.com/metaplex-foundation/metaplex-program-library/actions/runs/1559282195/workflow) please?

Additionally we should include a badge in the [readme](https://github.com/metaplex-foundation/metaplex-program-library#metaplex-contracts) at least for the _Program_.

austbot (changes requested):
Tests Dir structure

.github/workflows/program-nft-packs.yml
```diff
@@ -0,0 +1,53 @@
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ./rust/target
+          key: ${{ env.cache_id }}-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ env.RUSTC_HASH }}
+
+      # Build deps
+      - uses: ./.github/actions/build-token-vault
+      - uses: ./.github/actions/build-metaplex
+      - uses: ./.github/actions/build-token-metadata
```
thlorenz, line 45: You may not need this as when you build metaplex it'll build the other two since they're dependencies:
```
mpl-token-vault = { path = "../../token-vault/program", features = [ "no-entrypoint" ], version="0.0.1" }
mpl-token-metadata = { path = "../../token-metadata/program", features = [ "no-entrypoint" ], version="0.0.1" }
```
Therefore  all `.so` files will be produced so you could also use them for integration tests once you add those.

Building the other two separately could also mess with caching of the target folder (each action tries to cache it's result).

As a side-note I'm still not entirely sure how to cache this cleanly but we'll address this later.

VadimGrozinok: Oh ok thanks
Also have question about another one dependency - random-oracle program
NFT-packs depends on that one program but it's from different [repository](https://github.com/metaplex-foundation/randomness-oracle)
How can I build it in GH workflow as dependency ?

thlorenz: We'll have to setup an action to build this somehow so you can use it in your workflow.

Basically it would entail either cloning the repo and building the program needed or somehow pulling down the `.so` file by other means.

Anchor.toml
```diff
@@ -0,0 +1,19 @@
...
+[registry]
+url = "https://anchor.projectserum.com"
+
+[provider]
+cluster = "localnet"
+wallet = "~/.config/solana/id.json"
+
+[[test.genesis]]
+address = "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s"
+program = "fusion/program/tests/deps/mpl_token_metadata.so"
```
thlorenz, line 13: This doesn't seem correct as the built `.so` files get written to `./target/deploy` folder in the root, see [here](https://github.com/metaplex-foundation/metaplex-program-library/pull/33/files#diff-8fbc7010039c384550aad716b7faf9354fee2d7b8bb84c200bfff581587ad331R5)

thlorenz: So you should not somehow put it into `test/deps`, but use the one we build.

VadimGrozinok:
Hey @tomjohn1028
Looks like some of the fusion program tests don't pass
Could you look at it

tomjohn1028:
Screen Shot 2021-12-17 at 7 50 22 PM


@VadimGrozinok fixed locally but I don't have write access to the repo. cc @austbot

thlorenz:
Hi thanks for all your work! However I'm realizing that you're adding two contracts via one PR.

I'd prefer if we do one contract per PR instead. This makes it easier to review as well as manage reverts, etc. in the future.
Additionally it would most likely allow us to merge part of this already (nft packs) and work out the issues with the fusion contract tests separately.

Could you please split those out into two PRs? You don't have to provide them both at once. So let's start with a PR that only adds nft-packs and while that is in review we can prepare the fusion PR.
I'd greatly appreciate that!

VadimGrozinok:
I created new PR with nft-packs program moving: https://github.com/metaplex-foundation/metaplex-program-library/pull/50
Can we close this one so we will open new PR for fusion program moving ?

PR #35, opened 2021-12-15
chore: add auction-house contract

thlorenz:
## Summary

This moves over the Rust code of the auction-program from @metaplex-foundation/metaplex at
commit #22526991.

Additionally it adds a skeleton `./js` folder which at this points contains the following:

- `./src/` with _standin_ entry file
- `./scripts/api-gen-ts.js` which generates the IDL JSON file via anchor and will later
  generate the TypeScript SDK from that IDL
- `./idl` folder which contains the generated IDL
- `./package.json` which among the usual contract SDK package scripts also has an `api:gen`
  script which performs the API generation mentioned above

## Anchor Config

An `Anchor.toml` file was added which at this point considers only `auction-house` and its
dependencies.

## Github Workflows

A github workflow for both the Rust program and the JS SDK were added as well and the related
badges included in the Readme.

## Going Forward

When the TypeScript generator tool is complete I will add it here as a dependency and use it to
generate the entire TypeScript SDK as part of `./scripts/api-gen.ts`.

## Related PRs pulling in other Anchor Contracts

Please use this PR as a guide on how to pull in other contracts from
`@metaplex-foundation/metaplex`.  Make sure to include the commit hash of the metaplex repo at the time the contract was moved over like I did in the summary above. This will make it easier to later merge in changes that were applied there.

You should follow roughly the above steps when preparing the PR i.e.:

- include Rust code and adapt it to the naming scheme used here, this also will require udpates
  to the `Cargo.toml` of that contract
- include the contract in the main `Cargo.toml` `workspaces` and update the `Anchor.toml` to
  include the contract
- add a JS skeleton derived from the one included with this PR inside `auction-house/js` and
  ensure that `yarn api:gen` works and generates `idl/<program>.json` (this will require
  updating `js/scripts/api-gen-ts.js`
- make sure everything builds
- add github actions for both the Rust Program and the JS SDK of the added contract
- update the readme to include the relevant badges
- ideally add at least one Rust test for the contract which others can build up from

## Stabilizing added Contracts

- once the generator is providing SDK we can add integration tests driven from TypeScript
- please use the `tape` testing framework currently used in the `token-metadata/js` tests in
  order to use one tool for this job across the entire repo
- from then on each contract should have integration tests added and any new feature/fix should
  come with an associated test
- I will demonstrate how to write those for auction-house, but they're going to be similar in
  idea to the once added to `token-metadata/js` already


auction-house/js/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Developing
+
+In order to update the generated SDK when the rust contract was updated please run:
+
+```
+yarn gen:api
+```
+
+NOTE: at this point this only generates the IDL json file but will generate TypeScript
```
aheckmann, line 13:
```suggestion
NOTE: at this point this only generates the IDL json file but later we will generate TypeScript
```

austbot (dismissed):

auction-house/js/README.md
```diff
@@ -0,0 +1,18 @@
...
+definitions and SDK code as well, derived from that IDL.
+
+## LICENSE
+
+Apache v2.0
```
aheckmann, line 18: The license file in this PR is MIT so this should match. The license files in the other contracts in this repo are also MIT. 
Note that the [License file in the root of the repo is Apache 2.0](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/LICENSE). Separately we'll want to iron this out.

thlorenz:
@aheckmann fixed license and added _later_ to readme https://github.com/metaplex-foundation/metaplex-program-library/pull/35/commits/2b1b1e4a031510a9c3cd8c0e7ce446fee6be2c00

The _we_ was not correct since it is talking about what the command above is doing (not us).

aheckmann (approved):

PR #41, opened 2021-12-16
Fix limited edition instant sale

### LINKED ISSUE #40: Limited Edition Auction Place bid transaction fails after ~80 bids
Expected behavior: When placing a bid on a limited edition auction with instant sale the program should see that the bid is an instant sale and instantly push bid to stack. If stack is full, then bail. 

Actual behavior: When placing a bid on a limited edition auction with instant sale, the program will loop through all the bids and only insert at the highest point. This is not ideal because the program will run out of compute budget after ~80 checks to the array. https://github.com/metaplex-foundation/metaplex-program-library/blob/master/auction/program/src/processor.rs#L567

This makes limited edition instant sale auctions over 80 copies impossible. 

Note: This is error happens while using AuctionV2 as well.

bmeeder22:
Fixes #40

### DIFF
--- a/auction/program/src/processor.rs
+++ b/auction/program/src/processor.rs
@@ -562,6 +562,15 @@ impl BidState {
             BidState::EnglishAuction { ref mut bids, max } => {
                 match bids.last() {
                     Some(top) => {
+                        match instant_sale_price {
+                            Some(instant_price) => {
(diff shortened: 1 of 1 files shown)


VadimGrozinok:
We already have fix for this: https://github.com/metaplex-foundation/metaplex/pull/1102

PR #50, opened 2021-12-20
Move nft packs program

VadimGrozinok:
## What ?
Move nft-packs program to Metaplex-foundation repository from Metaplex/smart-contracts


.github/workflows/program-nft-packs.yml
```diff
@@ -0,0 +1,51 @@
...
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ./rust/target
+          key: ${{ env.cache_id }}-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ env.RUSTC_HASH }}
+
+      # Build deps
+      - uses: ./.github/actions/build-metaplex
+      - uses: ./.github/actions/build-token-metadata
```
thlorenz, line 44: Token metadata should already build when you build metaplex, so no need to build it again.

Anchor.toml
```diff
@@ -20,4 +20,4 @@ program = "target/deploy/mpl_auction_house.so"
 members = [
   "token-metadata/program",
   "auction-house/program"
-] 
+]
```
thlorenz, line 23: Maybe not necessary, but I'd have expected that you'd add nft-packs here?
Does `anchor build` work and create the `target/idl/nft_packs.json`?

package.json
```diff
@@ -34,6 +34,19 @@
     "eslint-plugin-prettier": "^4.0.0",
     "lerna": "^4.0.0",
     "prettier": "^2.4.1",
-    "typescript": "^4.5.2"
+    "ts-node": "^10.4.0",
```
thlorenz, line 37: These dependencies should not be added here, but instead in the JS folder you should create for the nft-packs since they're needed specifically for this project.
We want to structure dependencies in such a manner that they are self contained and we could theoretically move each contract folder like `nft_packs` into a separate repo. Thus as much as possible they should declare what they depend on instead of the root package.json having those deps.

As you see currently those deps are only the ones needed to lint and build.

package.json
```diff
@@ -34,6 +34,19 @@
...
+    "@types/chai": "^4.2.21",
+    "@types/mocha": "^9.0.0",
```
thlorenz, line 39: We use `tape` in this repository to write tests, please do not introduce another testing framework. Nor do we need `chai`.
Also I do not see tests anywhere, so not sure why this would be needed?

thlorenz (changes requested):
Hi thanks for separating out this PR.
I added a few notes.

The main issue I see is that there is no JS folder. Even for anchor contracts we will generate an SDK (I'm currently working on the needed tooling) and thus we create a JS folder for each contracts we merge in.

I outlined this in detail in the description of this PR https://github.com/metaplex-foundation/metaplex-program-library/pull/35 which also serves as an example on which folders to create and what they should contain until the generated code is added.

Please update this PR following those instructions.

Thanks.

.github/workflows/program-nft-packs.yml
VadimGrozinok: When we build metaplex program it doesn't produce necessary *.so file so we need to build token-metadata
It happens because programs are building as dynamic libraries

Anchor.toml
VadimGrozinok: NFT-packs isn't an anchor program so it will not generate idl file

thlorenz:
Per call we agreed that since this is a non-anchor contract we it doesn't make sense to add an empty JS folder for now.
Instead we need to manually add the JS SDK via a separate PR.

thlorenz (dismissed):
Thanks this is perfect now!

thlorenz (approved):

PR #64, opened 2022-01-08
(feat) make actions contitional on the path that changed +core

austbot:
Problem:
All the checks run allways

Solution: 
Make them conditional


aheckmann (commented):
LFG

PR #65, opened 2022-01-08
(feat) adds candy machine

austbot:
Adds Rust program from metaplex at 43ab97affa8fd018570b18229cd822c4bd955319.

Adapted it to build and included as part of Anchor setup.

Solita not compatible yet

Added candy-machine github actions for both Rust and JS SDK .. note that none of those have
any tests yet.

Will include CI badges in readme as well.


nft-candy-machine/js/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "author": "Metaplex Maintainers <contact@metaplex.com>",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@metaplex-foundation/beet": "0.0.0",
+    "@metaplex-foundation/beet-solana": "0.0.0",
+    "@solana/web3.js": "^1.31.0"
+  },
+  "devDependencies": {
+    "eslint": "^8.3.0",
+    "prettier": "^2.5.1",
```
thlorenz, line 47: WRT those (including eslint) they are also devDeps in the root package.json and in order to keep versions of those compatible we should probably not include them per package.

.github/workflows/program-token-metadata.yml
```diff
@@ -1,4 +1,4 @@
-name: Program Token Metadata
+name: Program Candy Machine
```
thlorenz, line 1: This should be a new workflow instead of changing an existing one.

thlorenz (changes requested):
LGTM except for those mentioned nits (prettier and related repo wide devdeps we can figure out later)

nft-candy-machine/js/package.json
austbot: but they fail, something is up with the workspace resolution

thlorenz: OK weird this worked before ... maybe it's due to switching yarn version?

For now let's keep this and figure out later.

.github/workflows/program-candy-machine.yml
```diff
@@ -0,0 +1,70 @@
...
+  CARGO_TERM_COLOR: always
+  SOLANA_VERSION: 1.8.5
+  RUST_TOOLCHAIN: stable
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
```
thlorenz, line 19: Needs to be `core`

.github/workflows/program-candy-machine.yml
```diff
@@ -0,0 +1,70 @@
...
+      id: filter
+      with:
+        filters: |
+          core:
+            - 'core/**'
+          package:
+            - 'nft-candy-machine/**'
+  build-and-test-candy-machine:
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
```
thlorenz, line 33: Needs to be `if: ${{ needs.changes.outputs.core == 'true' || needs.changes.outputs.package == 'true' }}`

.github/workflows/sdk-candy-machine.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
```
thlorenz, line 14: Ditto in program yml (`core`)

.github/workflows/sdk-candy-machine.yml
```diff
@@ -0,0 +1,36 @@
...
+      id: filter
+      with:
+        filters: |
+          core:
+            - 'core/**'
+          package:
+            - 'nft-candy-machine/**'
+  build-lint-and-test-auction:
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
```
thlorenz, line 28: Ditto in program yml (incorrect `if` )

thlorenz (changes requested):
Still some issues in the path filter of the workflows.

thlorenz (approved):

PR #90, opened 2022-01-12
Set NonFungible token standard for new mints

Flawm:
set TokenStandard to NonFungible for mints that come in that aren't edition infos or fungible assets.

### DIFF
--- a/token-metadata/program/src/utils.rs
+++ b/token-metadata/program/src/utils.rs
@@ -906,6 +906,7 @@ pub fn process_create_metadata_accounts_logic(
     )?;
 
     let mint_decimals = get_mint_decimals(mint_info)?;
+    let mint_supply   = get_mint_decimals(mint_info)?;
 
     metadata.mint = *mint_info.key;
     metadata.key = Key::MetadataV1;
@@ -919,6 +920,8 @@ pub fn process_create_metadata_accounts_logic(
     if add_token_standard {
         let token_standard = if is_edition {
             TokenStandard::NonFungibleEdition
+        } else if mint_supply == 1 && mint_decimals == 0 {
+            TokenStandard::NonFungible
         } else if mint_decimals == 0 {
             TokenStandard::FungibleAsset
         } else {


Flawm:
@austbot so what about solana monkey business? https://explorer.solana.com/address/GyohWe2b4yJeLQgAGP9ef2GGEbNBTAFAJ2j4aS8DqsvR

Flawm:
any thoughts on the matter of checking tokens that have null freeze & mint authorities with no master edition infos?

roederw:
> @austbot so what about solana monkey business? https://explorer.solana.com/address/GyohWe2b4yJeLQgAGP9ef2GGEbNBTAFAJ2j4aS8DqsvR
> 
> additionally, maybe we can alter the check to ensure mint authority is none, that code might look something like this
> 
> ```rust
> let no_mint_auth =  mint.mint_authority.is_none();
> ```

Solana Money Business pre-dates the Metaplex NFT Standard.

roederw:
Solana Monkey Business pre-dates the Metaplex NFT Standard, I'm pretty sure. I'd argue you need to have a `Master Edition` set moving forward to set `NonFungible`.

roederw:
@austbot any last minute comments?

Flawm:
I'd argue that 99.9% of mints don't leverage the power that is lent from master edition info and as such add needless complexity & 
a waste of gas. Any reason why it's so baked into the standard? Can't we check for null mint & freeze authority as well?

roederw:
It enforces that an NFT is actually non-fungible. You can create a `Master Edition` and even if it hasn't been minted yet, you can guarantee it will be unique. A `Mint Account` that hasn't been minted yet, doesn't enforce that it will remain non-fungible until after it's been minted and nulled out.

Flawm:
Right, but for tokens that do exist you can check to see if they have a fixed supply of 1 with decimals 0 with a null mint authority, thus guaranteeing they're Non-fungible & unique without any master edition info context. I'm not sure if SMB set a null mint authority, but with this check they could be grandfathered into the 1.1.0 standard as well as other tokens.

roederw:
Moving forward we want `Master Editions` to be leveraged when creating an NFT.

roederw:
We could re-evaluate this in a future standard version bump, but for now we're not ready to make this change. Thanks for opening up this PR though.

Flawm:
yeah, maybe possible to grand-father in old tokens in 1.1.1 or otherwise. It just seems weird to classify 'old' nfts as 'Fungible Tokens' as of right now

roederw:
I made a note to discuss this internally.

roederw:
I'll tag here if we come to an updated conclusion.

Flawm:
right on, appreciate you.

btw I just checked, smb doesn't have a null mint authority so theoretically a burned token could be reminted to a new user -- thus they can't be guaranteed to be NonFungible even if there were some grand-father check.

I'm not sure if they're aware of this issue, but I'm sure they'd be be interested in fixing this if brought to their attention - assuming they have the original mint authority signing keys (probably not)?

Flawm:
any thoughts or updates on the matter from the team at large?

Flawm:
Looks like this was inadvertently added into 1.3 - thanks :)
image

PR #106, opened 2022-01-18
(feat) adds collection authority system

### LINKED ISSUE #87: Add Collection Use Authority
# Problem

On Chain minting tools need an authority semantic by which an collection owner can delegate the right to add something to a collection.

# Solution

Add on chain collection delegated authority. Simular to how UseAuthority works, this would allow any number of collection authorities for an nft collection.

austbot:
Closes #87 
Allows collection authority to be delegated to another key.

- [ ] check data on verify or un verify of record pda

### DIFF
--- a/token-metadata/program/tests/approve_use_authority.rs
+++ b/token-metadata/program/tests/approve_use_authority.rs
@@ -342,15 +342,6 @@ mod approve_use_authority {
 
         let (burner, _) = find_program_as_burner_account();
 
-        let thing = context
-            .banks_client
-            .get_account(test_meta.token.pubkey())
-            .await
-            .unwrap()
-            .unwrap();
-
-        println!("{:?}", Account::unpack_from_slice(&thing.data).unwrap());
-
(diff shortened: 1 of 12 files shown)


token-metadata/program/tests/approve_use_authority.rs
```diff
@@ -342,15 +342,6 @@ mod approve_use_authority {
...
         let (burner, _) = find_program_as_burner_account();
 
-        let thing = context
-            .banks_client
-            .get_account(test_meta.token.pubkey())
-            .await
-            .unwrap()
-            .unwrap();
-
-        println!("{:?}", Account::unpack_from_slice(&thing.data).unwrap());
```
roederw, line 352: We needed this. Please bring it back /s

token-metadata/program/tests/approve_use_authority.rs
```diff
@@ -364,6 +355,8 @@ mod approve_use_authority {
             1,
         );
 
+
```
roederw, line 358: nit

roederw (approved):

PR #119, opened 2022-01-20
Add createMarket instruction, tests (js)

olgkv:
* update create store instruction (js)
* update tests for Store account (js)
* add create market instruction (js)
* add tests for Market account (js)

### DIFF
--- a/membership-token/js/test/transactions/create-market.ts
+++ b/membership-token/js/test/transactions/create-market.ts
@@ -1,0 +1,75 @@
+import { Connection, Keypair, PublicKey, Transaction } from '@solana/web3.js';
+import * as beet from '@metaplex-foundation/beet';
+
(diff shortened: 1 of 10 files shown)


kurpav (approved):

membership-token/js/test/transactions/create-market.ts
```diff
@@ -0,0 +1,75 @@
...
+      treasyryOwnerBump,
+      mutable,
+      price,
+      piecesInOneWallet,
+      startDate,
+      endDate,
+    },
+  );
+
+  const marketTx = new Transaction();
```
zaxozhu, line 65: This code below seems to be similar from place to place. 
What about moving transaction creation and signing to utils?
Same for **init-selling-resouce** and **create-store**

zaxozhu (approved):

zaxozhu (approved):

austbot (approved):

PR #121, opened 2022-01-20
Nft packs/remove random oracle

MetaMykola:


nft-packs/program/src/utils.rs
```diff
@@ -248,27 +248,24 @@ pub fn empty_account_balance(
...
-
-    if clock.slot.error_sub(slot)? > MAX_LAG_SLOTS {
-        return Err(NFTPacksError::RandomOracleOutOfDate.into());
-    }
-
-    // Hash random value from the oracle with current slot and proving process data and receive new random u16
+    // Hash slot, current timestamp and value from last slothash and proving process data and receive new random u16
     let mut hasher = DefaultHasher::new();
-    hasher.write(oracle_random_value.as_ref());
-    hasher.write(proving_process.try_to_vec()?.as_ref());
```
VadimGrozinok, line 267: It shouldn't be deleted because there can couple of this instructions in one slot

VadimGrozinok (changes requested):

nft-packs/program/src/utils.rs
MetaMykola: fixed

nft-packs/program/src/processor/request_card_to_redeem.rs
```diff
@@ -166,8 +170,13 @@ pub fn request_card_for_redeem(
         return Err(NFTPacksError::UserRedeemedAllCards.into());
     }
 
-    let random_value =
-        get_random_oracle_value(randomness_oracle_account, &proving_process, &clock)?;
+    // get slot hash 
+    let data = recent_slothashes_info.data.borrow();
+    let most_recent = array_ref![data, 8, 8];
+    let index = u64::from_le_bytes(*most_recent);
```
VadimGrozinok, line 176: We shouldn't convert hash into u64
Use recent slot hash as it is

nft-packs/program/src/utils.rs
```diff
@@ -248,27 +248,26 @@ pub fn empty_account_balance(
...
+    
+    // last slothash
+    hasher.write_u64(index);
```
VadimGrozinok, line 261: And based on previous comment here also will be some change(we will use write() method instead of write_u64())

VadimGrozinok (changes requested):

VadimGrozinok:
Not yet
Working on it
We already doing manual testing

PR #152, opened 2022-01-25
rename 'nft-candy-machine' to 'candy-machine' for consistency

samuelvanderwaal:
@thlorenz per our Slack conversation. This PR renames the `nft-candy-machine` program to `candy-machine` for naming consistency. This means that the actual package name will be `mpl-candy-machine` following the naming convention: `mpl-token-metadata`, etc.


thlorenz (approved):
LGTM, thanks!

PR #179, opened 2022-02-02
(fix) migrate bump seeds on UARs

austbot:
# Problem
Existing UseAuthorityRecords would become unusable if  1.2.0 was shipped
#Solution
As a part of the 1.2.0 changes we identified a migration case we could accomplish to avoid breaking the 100s of use Authority records in the wild.
We did not implement this migration for the CollectionAuthority records as they are not in use yet

### DIFF
--- a/token-metadata/program/src/state.rs
+++ b/token-metadata/program/src/state.rs
@@ -207,10 +207,6 @@ impl CollectionAuthorityRecord {
         let ca: CollectionAuthorityRecord = try_from_slice_checked(b, Key::CollectionAuthorityRecord, COLLECTION_AUTHORITY_RECORD_SIZE)?;
         Ok(ca)
     }
-
-    pub fn bump_empty(&self) -> bool {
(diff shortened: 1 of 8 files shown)


token-metadata/program/src/state.rs
```diff
@@ -174,15 +174,17 @@ impl UseAuthorityRecord {
...
     }
 
     pub fn from_bytes(b: &[u8]) -> Result<UseAuthorityRecord, ProgramError> {
         let ua: UseAuthorityRecord = try_from_slice_checked(b, Key::UseAuthorityRecord, USE_AUTHORITY_RECORD_SIZE)?;
-
         Ok(ua)
     }
+
+    pub fn bump_empty(&self) -> bool {
+       return self.bump == 0 && self.key == Key::UseAuthorityRecord;
```
samuelvanderwaal, line 186: Is this supposed to be 

```self.key == Key::UseAuthorityRecord```

or

```self.key != Key::UseAuthorityRecord```

samuelvanderwaal (dismissed):

PR #181, opened 2022-02-03
Receipt Accounts

kespinola:
# Goal

Provide a receipt of trade state that documents the trade state seed, easier to query with gPA, and processable with account db.

## Approach
Add instructions for writing a listing, bid, or purchase receipt based on the previous instruction in the transaction.

The print purchase receipt creates a purchase receipt and also sets on the bid and listing receipt the associated purchase receipt. When listing receipt has a purchase_receipt it is associated with a completed sale. When the canceled_at is set the bid or listing was canceled by the user.

## Accounts

```rust
pub struct BidReceipt {
    pub trade_state: Pubkey,
    pub bookkeeper: Pubkey,
    pub auction_house: Pubkey,
    pub buyer: Pubkey,
    pub metadata: Pubkey,
    pub token_account: Option<Pubkey>,
    pub purchase_receipt: Option<Pubkey>,
    pub price: u64,
    pub token_size: u64,
    pub bump: u8,
    pub trade_state_bump: u8,
    pub created_at: i64,
    pub canceled_at: Option<i64>,
}
```

`print_bid_receipt` documents either a public or private sale. When the sale is private the token_account is set on the bid receipt.

```rust
pub struct ListingReceipt {
    pub trade_state: Pubkey,
    pub bookkeeper: Pubkey,
    pub auction_house: Pubkey,
    pub seller: Pubkey,
    pub metadata: Pubkey,
    pub purchase_receipt: Option<Pubkey>,
    pub price: u64,
    pub token_size: u64,
    pub bump: u8,
    pub trade_state_bump: u8,
    pub created_at: i64,
    pub canceled_at: Option<i64>,
}
```

```rust
pub struct PurchaseReceipt {
    pub buyer: Pubkey,
    pub seller: Pubkey,
    pub auction_house: Pubkey,
    pub metadata: Pubkey,
    pub token_size: u64,
    pub price: u64,
    pub bump: u8,
    pub created_at: i64,
}
```


auction-house/program/src/lib.rs
```diff
@@ -966,6 +967,96 @@ pub mod auction_house {
...
+
+        assert_derivation(
+            &id(),
+            &trade_state.to_account_info(),
+            &ts_seeds,
+        )?;
+
+        if trade_state.data_is_empty() {
+            return Err(ErrorCode::TradeStateDoesntExist.into());
+        }
```
kespinola, line 1046: This needs to be dropped since the tradestate is closed out in the cancel and execute_sale.

auction-house/program/src/lib.rs
```diff
@@ -966,6 +967,96 @@ pub mod auction_house {
...
+
+        let receipt_lamp =  receipt.to_account_info().lamports();
+        **receipt.to_account_info().lamports.borrow_mut() = 0;
+
+        **bookkeeper.to_account_info().lamports.borrow_mut() = receipt
+            .to_account_info()
+            .lamports()
+            .checked_add(receipt_lamp)
+            .ok_or(ErrorCode::NumericalOverflow)?;
```
kespinola, line 1055: This should be closing out the receipt and send the funds to the wallet that created the receipt. CR here please.

kespinola (commented):
Logic is in place. I'll be adding specs this weekend. need to move to my linux box. M1 doesn't like 1.8.x of solana. WEN 1.9.5 :)

auction-house/program/src/lib.rs
```diff
@@ -965,6 +966,96 @@ pub mod auction_house {
...
+            &receipt.buyer_price.to_le_bytes(),
+            &receipt.token_size.to_le_bytes(),
+            &[trade_state_bump],
+        ];
+
+        assert_derivation(
+            &id(),
+            &trade_state.to_account_info(),
+            &ts_seeds,
+        )?;
```
kespinola, line 1009: This should be ensuring that receipts are only saved when the contents of the receipt equate to the seed of the tradestate ie only make receipts that match tradestates.

auction-house/program/src/lib.rs
kespinola: need to zero out data.

kespinola:
The use of receipts allows for tracking offers and listings but doesn't suffice for tracking completed sales (ie execute_sale) because both cancel and execute_sale close out the trade states. Closing the trade states is the appropriate action but since both actions result in the same outcome it's not possible to denote a sale from a cancelation.

The only active idea I have for distinguishing a sale from cancel is using the slot number and that is if all instructions assigned to a transaction (even CPI) are assigned to the same slot number. If this is true then a sale would be a zeroing out of the offer and listings receipts at the same slot number as a token transfer from the seller to the buyer.

https://docs.rs/solana-accountsdb-plugin-interface/1.9.6/solana_accountsdb_plugin_interface/accountsdb_plugin_interface/trait.AccountsDbPlugin.html#method.update_account

This is only feasible for teams tracking the state of the auction house program using account db where the slot number is provided in the on account change hook. 

The more accessible solution is to provide a new version of execute_sale that produces a purchase_receipt which represents a sale and exists on-chain permanently. It could be added to the current execute_sale action if folks are fine with using the remainder of accounts passed to the instruction. However, I believe this bypasses the validations and deserialization provided by anchor.

The execute_sale_v2 would accept the same accounts as the v1 with the addition of a purchase_receipt. The body of execute_sale_v2 will invoke execute_sale and then book the sale on the purchase_receipt. The purchase_receipt will remain on-chain as a record of the sale

The final approach I'm happy to explore is building this version of execute_sale into another program and using CPI to invoke execute_sale on auction house.

Let me know what you all think.

auction-house/program/src/lib.rs
```diff
@@ -1388,6 +1498,29 @@ pub struct AuctionHouse {
...
+1; // bump 
+
+
+#[account]
+pub struct Receipt {
+    pub trade_state: Pubkey,
+    pub bookkeeper: Pubkey,
+    pub auction_house: Pubkey,
+    pub wallet: Pubkey,
+    pub token_account: Pubkey,
```
kespinola, line 1518: Not tracking the metdata address for nft associated to the receipt because it can be derived from the mint of the token_account.

kespinola: https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/pda.rs#L29

kespinola:
> The use of receipts allows for tracking offers and listings but doesn't suffice for tracking completed sales (ie execute_sale) because both cancel and execute_sale close out the trade states. Closing the trade states is the appropriate action but since both actions result in the same outcome it's not possible to denote a sale from a cancelation.

This addition to this PR resolves issue with tracking execute_sale using an account.

https://github.com/holaplex/metaplex-program-library/pull/1

auction-house/program/src/lib.rs
```diff
@@ -965,6 +966,89 @@ pub mod auction_house {
...
+        let trade_state = &ctx.accounts.trade_state;
+        let receipt = &mut ctx.accounts.receipt;
+        let auction_house = &ctx.accounts.auction_house;
+        let token_account = &ctx.accounts.token_account;
+        let bookkeeper = &ctx.accounts.bookkeeper;
+        let wallet = &ctx.accounts.wallet;
+
+        receipt.trade_state = trade_state.key();
+        receipt.auction_house = auction_house.key();
+        receipt.token_account = token_account.key();
```
kespinola, line 985: Please audit me but not saving the metadata address because it can be looked up based on the mint of the token account.

auction-house/program/src/lib.rs
```diff
@@ -567,13 +567,136 @@ pub mod auction_house {
...
+            auction_house_fee_account: auction_house_fee_account.clone(),
+            auction_house_treasury: auction_house_treasury.clone(),
+            program_as_signer: program_as_signer.clone(),
+            token_program: token_program.clone(),
+            system_program: system_program.clone(),
+            ata_program: ata_program.clone(),
+            rent: rent.clone(),
+        };
+
+        let context = Context::new(ctx.program_id, &mut accounts, ctx.remaining_accounts);
```
austbot, line 686: Why not encapsulate logic and just call the actual code?

auction-house/program/src/lib.rs
```diff
@@ -1064,6 +1366,111 @@ pub struct Sell<'info> {
...
+    #[account(mut)]
+    bookkeeper: Signer<'info>,
+    system_program: Program<'info, System>,
+    rent: Sysvar<'info, Rent>,
+    clock: Sysvar<'info, Clock>,
+}
+
+#[derive(Accounts)]
+pub struct CloseListingReceipt<'info> {
+    token_account: Account<'info, TokenAccount>,
```
austbot, line 1389: Are you actually closing the reciept looks like the account stays out there forever

auction-house/program/src/lib.rs
```diff
@@ -1064,6 +1366,111 @@ pub struct Sell<'info> {
...
+    trade_state: UncheckedAccount<'info>,
+    #[account(mut, seeds=[PUBLIC_BID_PREFIX.as_bytes(), trade_state.key().as_ref()], bump=receipt.bump)]
+    receipt: Account<'info, PublicBid>,
+    system_program: Program<'info, System>,
+    clock: Sysvar<'info, Clock>,
+}
+
+#[derive(Accounts)]
+#[instruction(purchase_receipt_bump: u8, escrow_payment_bump: u8, free_trade_state_bump: u8, program_as_signer_bump: u8, buyer_price: u64, token_size: u64)]
+pub struct ExecuteSaleWithReceipt<'info> {
```
austbot, line 1432: I dont know about this pass through, what value does it add over the extra instruction pattern

auction-house/program/src/lib.rs
```diff
@@ -1266,6 +1673,84 @@ pub struct AuctionHouse {
...
+32 + // token_mint
+8 + // price
+8 + // token_size
+1 + // bump
+1 + // trade_state_bump
+1 + 8 + // activated_at
+1 + 8; // closed_at
+
+#[account]
+pub struct PublicBid {
```
austbot, line 1690: Is this a reciept or id you add to my public tratestate stuff

austbot: I think its a receipt but the name confuses me, lets put Receipt in there

austbot (changes requested):
Review:
1. Why do we need a new execute sale, you may have told me but I  forgot.
2. Can we iterativley improve the code organization, can you move this our of the megafile of death and into packages like I did with bid.

In the future all new anchor instruction I want the Accounts struct right on top of the processing method and in the same file so that the logic can be grokked since its shared between the accounts struct and the processor,

auction-house/program/src/lib.rs
kespinola: The account do stay alive forever so the record of the bids, offers, and purchases as accounts. closing out would clear out the data contents. The close actions set the closed_at timestamp to signal the cancel or completion of a sale.

auction-house/program/src/lib.rs
kespinola: To write the purchase receipt need to accept another account to the instruction. Can this be achieved without another instruction?

kespinola: We can match an incoming account against what we have indexed and if its data is cleared write the closed_at as the current time of processing but that causes some DR issues where a loss of the db makes it very difficult to rebuild the index based on the state of accounts on chain. As of right now we can blow our index away and rebuild it based on the active accounts on-chain.

auction-house/program/src/lib.rs
kespinola: We can match an incoming account against what we have indexed and if its data is cleared write the closed_at as the current time of processing but that causes some DR issues where a loss of the db makes it very difficult to rebuild the index based on the state of accounts on chain. As of right now we can blow our index away and rebuild it based on the active accounts on-chain.

auction-house/program/src/lib.rs
kespinola: yep. done. added suffix to all receipt-related accounts.

auction-house/program/src/lib.rs
kespinola: migrated to checking previous instruction so don't need proxy anymore.

auction-house/program/src/lib.rs
kespinola: keeping open but renamed to cancel to not mislead caller expecting the account to be closed out.

auction-house/program/src/pda.rs
```diff
@@ -95,3 +95,34 @@ pub fn find_public_bid_trade_state_address(
         &id(),
     )
 }
+
+/// Return public bid receipt `Pubkey` address and bump seed.
```
kespinola:
```suggestion
/// Return bid receipt `Pubkey` address and bump seed.
```

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    let mut receipt = ListingReceipt::try_deserialize(&mut receipt_data_slice)?;
+
+    receipt.canceled_at = Some(clock.unix_timestamp);
+
+    let mut data = receipt_data;
+    let dst: &mut [u8] = &mut data;
+    let mut cursor = std::io::Cursor::new(dst);
+
+    receipt.try_serialize(&mut cursor)?;
```
kespinola, line 226: This pattern pulled from anchor_lang.

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    let mut buffer = &prev_instruction.data[8..];
+    let buy_data = Buy::deserialize(&mut buffer)?;
+
+    let bid_type = assert_program_bid_instruction(&prev_instruction.data[..8])?;
+
+    let token_account = match bid_type {
+        BidType::PrivateSale => Some(token_account.pubkey),
+        BidType::PublicSale => None,
+    };
```
kespinola, line 272: Handles writing the token_account when a private sale. None for public.

auction-house/program/src/utils.rs
```diff
@@ -168,6 +169,30 @@ pub fn assert_keys_equal(key1: Pubkey, key2: Pubkey) -> ProgramResult {
...
 
+pub enum BidType {
+    PublicSale,
+    PrivateSale,
+}
+
+pub fn assert_program_bid_instruction(sighash: &[u8]) -> Result<BidType, ErrorCode> {
+    match sighash {
+        [169, 84, 218, 35, 42, 206, 16, 171] => Ok(BidType::PublicSale),
+        [102, 6, 61, 18, 1, 218, 235, 234] => Ok(BidType::PrivateSale),
```
kespinola, line 180: Matches call signature for public_buy or buy instruction.

auction-house/program/src/pda.rs
kespinola: done.

auction-house/program/src/lib.rs
kespinola: It can when checking the previous instruction. Ty for the tip.

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    let prev_instruction = get_instruction_relative(-1, instruction_account)?;
+    let prev_instruction_accounts = prev_instruction.accounts;
+
+    let wallet = &prev_instruction_accounts[0];
+    let auction_house = &prev_instruction_accounts[4];
+    let seller_trade_state = &prev_instruction_accounts[6];
+    let metadata = &prev_instruction_accounts[2];
+
+    let mut buffer = &prev_instruction.data[8..];
```
kespinola, line 127: The 1st 8 bytes are for the ID of the instruction so drop those bytes out before trying to deserialize the instruction data.

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+    let sell_data = Sell::deserialize(&mut buffer)?;
+
+    assert_program_instruction_equal(
+        &prev_instruction.data[..8],
+        [51, 230, 133, 164, 1, 127, 131, 173],
+    )?;
```
kespinola, line 133: Anchor adds 8 byte ID to the instruction that can be located by running.

```
cargo rustc --profile=check -- -Zunpretty=expanded | grep -C 10 InstructionData
```

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+#[derive(Accounts)]
+#[instruction(receipt_bump: u8)]
+pub struct PrintBidReceipt<'info> {
+    #[account(mut)]
+    receipt: UncheckedAccount<'info>,
+    #[account(mut)]
+    bookkeeper: Signer<'info>,
+    system_program: Program<'info, System>,
+    rent: Sysvar<'info, Rent>,
+    instruction: UncheckedAccount<'info>,
```
kespinola, line 240: No Sysvar is available for instructions so passing the program id as an unchecked account. Let me know if there is an alternative.

austbot:
```suggestion
    #[account(address = sysvar::instructions::id())]
    pub instructions: UncheckedAccount<'info>,
```

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,510 @@
...
+    let receipt_data = receipt_info.try_borrow_mut_data()?;
+    let mut receipt_data_slice: &[u8] = &receipt_data;
+
+    let mut receipt = ListingReceipt::try_deserialize(&mut receipt_data_slice)?;
+
+    receipt.canceled_at = Some(clock.unix_timestamp);
+
+    let mut data = receipt_data;
+    let dst: &mut [u8] = &mut data;
+    let mut cursor = std::io::Cursor::new(dst);
```
austbot, line 226: Whats the purpose of cursor?

auction-house/program/src/receipt/mod.rs
```diff
@@ -0,0 +1,510 @@
...
+    let receipt_data = receipt_info.try_borrow_mut_data()?;
+    let mut receipt_data_slice: &[u8] = &receipt_data;
+
+    let mut receipt = BidReceipt::try_deserialize(&mut receipt_data_slice)?;
+
+    receipt.canceled_at = Some(clock.unix_timestamp);
+
+    let mut data = receipt_data;
+    let dst: &mut [u8] = &mut data;
+    let mut cursor = std::io::Cursor::new(dst);
```
austbot, line 366: Same as Above

austbot (commented):
Just a few questions on the cursor stuff.

auction-house/program/src/receipt/mod.rs
kespinola: removed by reusing existing borrowed reference.

https://github.com/metaplex-foundation/metaplex-program-library/pull/181/commits/5494a6af2ad0e456262d87e925ae8083e33e0a7e

auction-house/program/src/receipt/mod.rs
kespinola: dropped by reusing borrow reference used to deserialize.

https://github.com/metaplex-foundation/metaplex-program-library/pull/181/commits/5494a6af2ad0e456262d87e925ae8083e33e0a7e

auction-house/program/src/receipt/mod.rs
kespinola: https://github.com/metaplex-foundation/metaplex-program-library/pull/181/commits/570736061c93ed266ea743876c9db52705c5e410

austbot (approved):

PR #183, opened 2022-02-03
Refactor set_store_index to handle offset == len.

ray-kast:
`set_store_instruction` has a couple argument validation errors, chiefly several instances of an off-by-one bug for comparing `offset` to the index length and an incorrect error response for a missing `below_cache`.  This makes it impossible to append a cache to the end of a store index (i.e. `offset = length`) and may possibly cause undesired behavior when inserting an element at `offset = length - 1`.

Work left to do:
 - [x] Test changes against various use cases
   - Empty index list
   - Non-empty, inserting 0th element
   - Non-empty, inserting (n - 1)th element
   - Non-empty, appending new last (nth) element
   - Non-empty, inserting in the middle
 - [x] Handle validation of `positional_info_2` when it is not used
 - [x] Add additional `msg!` calls to improve transaction log forensics in the wild


metaplex/program/src/processor/set_store_index.rs
```diff
@@ -143,12 +105,61 @@ pub fn process_set_store_index<'a>(
...
+    let above_key = indexer.auction_caches.get(offset);
+    let below_key = offset
+        .checked_sub(1)
+        .and_then(|i| indexer.auction_caches.get(i));
+
+    let (above_cache_info, below_cache_info) = if above_key.is_some() {
+        (positional_info_1, positional_info_2)
+    } else {
+        // When above is not required, below becomes the first and only account argument here
+        (None, positional_info_1)
```
ray-kast, line 121: It might be desirable to assert that `positional_info_2` is `None` here to reduce user confusion, but this would likely be a breaking change compared to current behavior.

kespinola (dismissed):

PR #186, opened 2022-02-04
Adds Freeze Authority

austbot:
https://github.com/metaplex-foundation/metaplex-program-library/pull/171

I had to re make the PR so I could fix it.

We have been building a protocol that enabled rentals for new tokens and were planning to try to enable this functionality for existing NFTs (whose freeze_authority is held by metaplex-token-metadata)

Our proposal is: freeze/unfreeze instructions that proxy instruction through to spl_token_program IF the signer is the delegate of the single token in the token_account that they are trying to freeze.

In other words, if someone delegated you the NFT in a token account, you can freeze that specific token account. Essentially this opens up a world of possibilities around conditional ownership, rentals, time-revoking tokens etc. where you as the holder can claim it but agree by signing away delegate rights that this other program can manage / revoke it from you.

I can't think of any reasons this would cause issues. The only interesting thing is that if I delegate tokens to someone, they could always freeze my account forever, but they also could always just take the tokens so I think thats okay because the user has signed away their rights to the token by delegating it

@austbot
One thing I just thought of was that if this is your ATA, freeze might cause you to not be able to accept new tokens. This should be a non-issue with NFTs of course unless there is a specific scenario where a semi-fungible token has an edition but dont think thats possible

### DIFF
--- a/token-metadata/program/src/error.rs
+++ b/token-metadata/program/src/error.rs
@@ -350,7 +350,7 @@ pub enum MetadataError {
     #[error("This Collection Authority Record Already Exists.")]
     CollectionAuthorityRecordAlreadyExists,
 
-    #[error("This Collection Authoritty Record Does Not Exist.")]
+    #[error("This Collection Authority Record Does Not Exist.")]
     CollectionAuthorityDoesNotExist,
 
     #[error("This Use Authority Record is invalid.")]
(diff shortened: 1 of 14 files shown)


Amjad-W:
Can we please have an example/documentation of some sort? i'm struggling to make this work calling `freeze_delegated_account`
Thank you

PR #195, opened 2022-02-05
feat: add findByOwnerV3 method

fersilva16:
I've tested it with an account with 11 tokens and it's around 10-25 times faster than `findByOwnerV2`. Not sure if it's still faster with wallets that have way more tokens.

Here is the [benchmark file](https://gist.github.com/fersilva16/b34cc997e952d8c4d4573b5e30e24927).
Some results:
| findByOwnerV2  | findByOwnerV3 | Times faster |
| -------------- | ------------- | ------------ |
| `23.220s`      | `840.56ms`    | `26.62x`     |
| `16.619s`      | `1.586s`      | `9.47x`      |
| `15.926s`      | `1.604s`      | `8.92x`      |
| `20.315s`      | `969.688ms`   | `19.95x`     |
| `16.426s`      | `1.596s`      | `9.29x`      |
| `20.174s`      | `1.032s`      | `18.54x`     |

### DIFF
--- a/token-metadata/js/src/accounts/Metadata.ts
+++ b/token-metadata/js/src/accounts/Metadata.ts
@@ -242,6 +242,12 @@ export class Metadata extends Account<MetadataData> {
     }
   }
 
+  static async findByMint(connection: Connection, mint: AnyPublicKey) {
+    const pda = await this.getPDA(mint);
+
+    return this.load(connection, pda);
+  }
+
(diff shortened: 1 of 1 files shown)


austbot (dismissed):

zaxozhu:
findDataByOwner looks the same, the only difference it returns deserialized data. It would be great to reuse the same functionality in it then.

austbot (approved):

PR #199, opened 2022-02-07
feat: fixed-price-sale - CreateSecondaryMetadataCreators

zaxozhu:
* Adds CreateSecondaryMetadataCreators instruction
* Adds SecondaryMetadataCreators & Creator accounts
* Creator should be removed when mpl-token-metadata starts to use Beet
* Update file naming to use single convention


fixed-price-sale/js/src/instructions/createSecondaryMetadataCreators.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+    {
+      pubkey: web3.SystemProgram.programId,
+      isWritable: false,
+      isSigner: false,
+    },
+  ];
+
+  const ix = new web3.TransactionInstruction({
+    programId: new web3.PublicKey('SaLeTjyUa5wXHnGuewUSyJ5JWZaHwz3TxqUntCE9czo'),
```
kurpav, line 72: please use program id const

kurpav (changes requested):
what is the point to change file names from kebab case to camel case?

olgkv (dismissed):

zaxozhu:
> what is the point to change file names from kebab case to camel case?

Solita generates files in camelCase, it's better to stay with the rules it dictates, so we won't be changing file names every time we generate APIs.

olgkv (approved):

tbiedukhin (approved):
Good job

kurpav (approved):

olgkv (approved):

austbot (approved):

PR #201, opened 2022-02-07
fix candy machine19

austbot:
- (fix) ensure correct metadata program
- (fix) fixes candymachine deprecated sysvar

### DIFF
--- a/candy-machine/program/src/lib.rs
+++ b/candy-machine/program/src/lib.rs
@@ -27,10 +27,13 @@ use {
     spl_token::state::Mint,
     std::{cell::RefMut, ops::Deref, str::FromStr},
 };
+use solana_program::sysvar::SysvarId;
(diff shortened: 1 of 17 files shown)


candy-machine/program/src/lib.rs
```diff
@@ -27,10 +27,13 @@ use {
     spl_token::state::Mint,
     std::{cell::RefMut, ops::Deref, str::FromStr},
 };
-anchor_lang::declare_id!("cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ");
+use solana_program::sysvar::SysvarId;
+anchor_lang::declare_id!("6EbJsELBbYMa9DL828YJijPvmRH5j9Y483yoxZVgh9pM");
```
roederw, line 31: Everything looks good here. Did you want to generate a cool public key (like above?). If not, you're good to check in.

roederw (dismissed):

roederw (dismissed):

PR #203, opened 2022-02-07
Allow limited editions to have a different metadata uri than their master edition

CalebEverett:
Allows limited editions to have a different uri value in their metadata accounts. Edition name gets set as master edition name with edition number appended if new uri provided.

### DIFF
--- a/token-metadata/cli/src/main.rs
+++ b/token-metadata/cli/src/main.rs
@@ -1,39 +1,36 @@
-
 use solana_client::rpc_request::TokenAccountsFilter;
 
-use {
-    clap::{crate_description, crate_name, crate_version, App, Arg, ArgMatches, SubCommand},
-    mpl_token_metadata::{
-        instruction::{
-            create_master_edition, create_metadata_accounts,
-            mint_new_edition_from_master_edition_via_token, puff_metadata_account,
-            update_metadata_accounts,
-        },
-        state::{
-            get_reservation_list, Data, Edition, Key, MasterEditionV1, MasterEditionV2, Metadata,
-            EDITION, MAX_NAME_LENGTH, MAX_SYMBOL_LENGTH, MAX_URI_LENGTH, PREFIX,
-        },
+use clap::{crate_description, crate_name, crate_version, App, Arg, ArgMatches, SubCommand};
+use mpl_token_metadata::{
+    instruction::{
+        create_master_edition, create_metadata_accounts,
+        mint_new_edition_from_master_edition_via_token, puff_metadata_account,
+        update_metadata_accounts,
     },
-    solana_clap_utils::{
-        input_parsers::pubkey_of,
-        input_validators::{is_url, is_valid_pubkey, is_valid_signer},
+    state::{
+        get_reservation_list, Data, Edition, Key, MasterEditionV1, MasterEditionV2, Metadata,
+        EDITION, MAX_NAME_LENGTH, MAX_SYMBOL_LENGTH, MAX_URI_LENGTH, PREFIX,
     },
-    solana_client::rpc_client::RpcClient,
-    solana_program::{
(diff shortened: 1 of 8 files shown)


CalebEverett:
Sorry about all the extra formatting changes. I think you guys pushed a change to master that I tried to merge in before creating the pull request, but it looks like it got the formatting wrong.

samuelvanderwaal:
Hey Caleb, thanks for the PR. I know we were talking about something related to this on Discord but can you elaborate on your goals and motivations for this change? The changes as you've currently constructed them aren't backwards compatible as you can't just add fields to a struct without changing the account size (unless there's padding built-in). In addition, this changes the design and purpose of Limited Editions which are currently designed to be carbon copies of Master Editions. As I understand these changes, they would require a significant change to the token metadata standard and probably a full version bump along with everything that entails (updating tooling, publicizing changes to stakeholders like wallets and marketplaces).

CalebEverett:
Hey Sam, thanks for checking it out.

#### Motivation
This is a way to accurately represent arguably the most common implementation of a collection of tokens where you have some metadata that is common to all tokens in the collection, some metadata that is specific to each token and a fixed supply of tokens. It seems this pattern is most often implemented by having a different metadata uri that points to a different json file for each token that usually has the same description and symbol, but different artwork and traits and a name that is the same for all, but with the number of the token appended to the end. In this set up, token holders trust issuers to limit supply to the advertised amount.

This pull request makes a minor modification to the Limited Edition concept, I believe only to the program instructions and not to the structure of the account data, to cover this most common use case.

#### Limited Edition
I get that the original intent of the Limited Edition was to be a carbon copy of the Master Edition.  Although my guess is that that use case is a lot less common than case described above, the carbon copy use case would still be possible with this change by passing `None` for the optional `uri` argument.

It does seem like there may be overlap in concept now between  a `FungibleAsset` and a `NonFungibleEdition`. If the desire is to have a carbon copy of a Master Edition, it seemed like it might be more efficient to have something that points to the same Metadata account as opposed to creating a new Metadata account for each edition populated with exactly the same data as the Master Edition metadata account.

However, this PR uses the fact that the current implementation already creates a new Metadata account to not require any changes to the size or structure of the current accounts. It just takes a uri as an argument and instead of duplicating all of the Master Edition metadata, it uses the new uri and changes the name to include the number of the edition. Neither the `Edition` or `EditionMarker` structs are changed. I think this would only be breaking to the extent that it would require passing `None` for the uri argument when creating editions.

#### Collection
The idea of tying tokens to another token seems like it has a lot of utility, but adds a decent amount of complexity, requiring the creation of an additional `NonFungible` token and including it as an argument in the creation of member tokens. More importantly, additional logic is still required to limit the supply of tokens in a collection. This PR uses all of the logic that already exists to support fixed supply of limited editions to avoid having to create an additional token or any additional supply logic. For more complex arrangements, i.e., a families of collections, being able to add a collection token would work really well and this PR, enabling fixed supply of collection tokens, would make that set up better.

samuelvanderwaal:
Thanks for explaining your thoughts and elaborating. I'll review the code more thoroughly this week.

CalebEverett:
@austbot  - not knocking the current collections functionality at all - lots of added utility and ergonomically designed. I was just saying that if you had an edition, you could tell whether it was derived from a particular master edition directly from its pda. Also, you could make editions mutable, but you don't have to either. In the implementation here, I opted for not, keeping `is_mutable` false.

roederw:
Hey @CalebEverett I'm also popping in to see what you've built. Although before I begin, I'd like to mention that because of nature of importance contained in these programs, we'd prefer that individuals first propose requests, before submitting PRs. That sparks conversation/collaboration within the community before adding new functionality (from what I've read from @samuelvanderwaal and @austbot there seems to potentially be a backwards compatibility conflict).

We will have a better mechanism for these sorts of proposals/requests in the future.

CalebEverett:
You missed it

PR #204, opened 2022-02-09
Fix/fixed sale price claim

bezrazli4n0:
## Notes
- Fixed critical bug in 'ClaimResource' instruction, that revert transaction if user use native SOL

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -265,22 +265,6 @@ dependencies = [
 
 [[package]]
 name = "anchor-client"
-version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cac39d0e96187b40a943361120733299fad52550df837ffc4366a261a89114cb"
-dependencies = [
- "anchor-lang 0.19.0",
- "anyhow",
- "regex",
- "serde",
- "solana-client",
- "solana-sdk",
- "thiserror",
- "url",
-]
-
-[[package]]
-name = "anchor-client"
 version = "0.20.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "05ae5e7cf28684fc1c3d1e1671a37ac1aa9b01e53f35982d219a88759d9bd044"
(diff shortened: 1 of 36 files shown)


austbot (approved):

fixed-price-sale/program/src/lib.rs
```diff
@@ -786,8 +786,21 @@ pub mod fixed_price_sale {
             }
         }
 
+        let is_native = market.treasury_mint == System::id();
```
austbot, line 789: There is a spl native mint id, this is what should be used in stead of system

fixed-price-sale/program/src/lib.rs
```diff
@@ -786,8 +786,21 @@ pub mod fixed_price_sale {
...
+
+        let treasury_holder_amount = if is_native {
+            treasury_holder.lamports()
```
austbot, line 792: should this not be try_borrow_lamports() ?

austbot (changes requested):
Some comments

fixed-price-sale/program/src/lib.rs
bezrazli4n0: Here we use `System::id()` to indicate native SOL's, but spl native mint provide wrapped spl SOLs

fixed-price-sale/program/src/lib.rs
bezrazli4n0: This PR is broken, actual resumed PR [here](https://github.com/metaplex-foundation/metaplex-program-library/pull/207)

bezrazli4n0:
This PR is broken, actual resumed PR [here](https://github.com/metaplex-foundation/metaplex-program-library/pull/207)

fixed-price-sale/program/src/lib.rs
bezrazli4n0: I'm not sure that this required in current context, because we don't need check dynamically borrow rules.

PR #208, opened 2022-02-10
[Proposal] append mint_number to metadata uri

HaoCherHong:
As mentioned [here](https://github.com/metaplex-foundation/metaplex/issues/1331)
I've added mint_number in the tail of metadata uri. This makes more sense because nft issuer doesn't have to update every metadata uri for each mint to reveal, which will be cumbersome and wasting transaction fees.

I understand that updating candy machine program will involve many changes to docs and address changes for related programs. But here is the proposal and please let me know what should I do to contribute. Many Thanks.

# Changes
- Append mint_number to metadata uri
- Update program id

### DIFF
--- a/candy-machine/program/src/lib.rs
+++ b/candy-machine/program/src/lib.rs
@@ -28,7 +28,8 @@ use {
     std::{cell::RefMut, ops::Deref, str::FromStr},
 };
 use solana_program::sysvar::SysvarId;
-anchor_lang::declare_id!("cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ");
+anchor_lang::declare_id!("8VRwdyLCnj3KYuKpsEuWZqyaqgTMQ5ESAZEohS5vwBDN");
+// anchor_lang::declare_id!("cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ");
 
 const EXPIRE_OFFSET: i64 = 10 * 60;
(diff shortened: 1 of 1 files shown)


samuelvanderwaal:
Adding the mint number to the end of the URI will just result in invalid URIs that wallets and explorers won't be able to properly parse. You could add it as a hash at the end of the URI without affecting the URI's ability to resolve but that is kind of a hacky solution. However, this doesn't add any additional information to the NFT as the number is already in the name.

The issue here is that you seem to misunderstand the purpose of the hash in the hidden settings. The hidden settings work like this:

Offline, you create a file that maps between the mint number and the actual metadata URI, something like this:

```json
[ 
  {
     "mint_num": 1,
     "uri": "https://arweave.net/1234abcdef"
   },
  {
     "mint_num": 2,
     "uri": "https://arweave.net/1235abcdef"
   },
]
```

You then take a [hash](https://en.wikipedia.org/wiki/Hash_function) of the file and this value is what you put in the config file to be stored on-chain. Now you proceed with the mint and everyone gets a different numbered NFT but with the same URI/metadata. After the mint is over, you now update all the NFTs with the new URIs from your file using Metaboss or some other tool that lets you update NFTs. 

What's the purpose of the hash then? You can publish this file you created to the users somewhere and they can download it, check the mint number they received against the expected URI and the actual URI their NFT has. They can also hash the file and compare the hash against what's stored on chain. If the hashes match they know you didn't cheat them.

PR #216, opened 2022-02-12
auction house publicbuy

austbot:
#Adds Public Bids

Adds a new instruction that allows a trade state to be created that does not have the seller token account in it. 
This allows the bid to be valid even if the NFT changes hands before the sale is complete.

## Example
see `execute_sale` tests for an example.

Cancel, Execute Sale, Buy , Public buy support both trade state formats. It is also resilient to `switcharoo` attack since you cannot use the new tradestade as the seller trade state and steal a bidders funds.

It also includes a 0 data check so that a tradestate cannot be used to execute sale after it has been emptied by the program.


roederw:
This looks to me. How long are we going to have this bake in `devnet`?

roederw (approved):

PR #221, opened 2022-02-15
fix(token-metadata): add missing account

samuelvanderwaal:
add missing account to revoke_collection_authority instruction

### DIFF
--- a/token-metadata/Cargo.lock
+++ b/token-metadata/Cargo.lock
@@ -1510,7 +1510,7 @@ dependencies = [
 
 [[package]]
 name = "mpl-token-metadata"
-version = "1.2.4"
+version = "1.2.5"
 dependencies = [
  "arrayref",
  "borsh",
--- a/token-metadata/js/package.json
+++ b/token-metadata/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metaplex-foundation/mpl-token-metadata",
-  "version": "1.2.4",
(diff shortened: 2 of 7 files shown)


samuelvanderwaal:
> This will break the tests, why is this needed, the tests are passing to revoke the authority

I'll look at the tests but the Rust instruction is wrong. Currently it [looks like this](https://github.com/metaplex-foundation/metaplex-program-library/blob/d2adfc68dd05c9294efd5d6054604b0295ecab1c/token-metadata/program/src/instruction.rs#L1035):

Selection_669

image

It's missing the new/delegate collection authority from the accounts list. The [TypeScript bindings](https://github.com/metaplex-foundation/metaplex-program-library/blob/d2adfc68dd05c9294efd5d6054604b0295ecab1c/token-metadata/js/src/transactions/RevokeCollectionAuthority.ts#L18) correctly have the fifth account:

Selection_667

Presumably they're generated from the [`processor.rs` file](https://github.com/metaplex-foundation/metaplex-program-library/blob/d2adfc68dd05c9294efd5d6054604b0295ecab1c/token-metadata/program/src/processor.rs#L1177) because that also has the fifth account:

Selection_668

If you call the Rust instruction as is, it will fail with "not enough accounts" error. If you add the new collection authority as the fifth account, it works. 

I'll look at and update the revoke tests later today or tomorrow as part of this PR.

austbot (approved):

PR #222, opened 2022-02-15
WIP: create entangled pair shouldn't require edition accounts

levicook:
This pull request introduces a second version of create entangled pair within the token-entangler program. 

- The initial version doesn't _really_ use these accounts.
- Dropping this constraint allows us to entangle pre-metaplex standard NFTs


levicook:
closing in favor of #213

PR #225, opened 2022-02-16
PAckage IMprovements

austbot:
Updates solita, fixes lint nits, bumps versions and generates a CM mpl sdk


auction-house/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metaplex-foundation/mpl-auction-house",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
thlorenz, line 3: did you actually publish a new version and tag it?
Please do that only once we merged into master since once we squash the PR that tagged commit is no longer there.
Workflow should be:

1. PR changes
2. from master `npm version && npm pub` which tags properly, adds a commit and pushes it + tag

candy-machine/js/.gitignore
```diff
@@ -1,2 +1,2 @@
 idl/*
-!idl/auction_house.json
\ No newline at end of file
+!idl/*.json
```
thlorenz, line 2: We actually _do_ wanna check in the IDL so that other people can derive their SDKs from it if needed.

candy-machine/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metaplex-foundation/mpl-candy-machine",
-  "version": "0.0.1",
+  "version": "3.0.1",
```
thlorenz, line 3: Same here, I cannot see that tag anywhere [here](https://github.com/metaplex-foundation/metaplex-program-library/tags). And as said it would disappear once the PR is squashed.

We want devs to be able to easily find the code corresponding to each npm version (that's why I added those `npm pre version and post publish` scripts.

thlorenz (approved):
IMHO this PR is doing way too much.. should've done this one package at a time ... and definitely don't version or publish until it's in master. 
It's super difficult to review ....
I understand though since deps were updated that some changes affected multiple packs.

I'm approving, but hoping that we'll do more isolated PRs in the future.

auction-house/js/package.json
austbot: no I do that from master

candy-machine/js/package.json
austbot: I only depploy from master

PR #230, opened 2022-02-17
fixes prints when collection is added

austbot:
# Problem
Collections on Master Editions that are Verified were breaking Editions being minted

#Solution
Allow Creation Verification if the token standard is NonFungible Edition

### DIFF
--- a/token-metadata/program/src/processor.rs
+++ b/token-metadata/program/src/processor.rs
@@ -275,7 +275,7 @@ pub fn process_update_metadata_accounts_v2(
                 true,
             )?;
             metadata.data = compatible_data;
-            assert_collection_update_is_valid(false,&metadata.collection, &data.collection)?;
+            assert_collection_update_is_valid(false, &metadata.collection, &data.collection)?;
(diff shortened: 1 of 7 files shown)


roederw (dismissed):

token-metadata/program/src/processor.rs
```diff
@@ -275,7 +275,7 @@ pub fn process_update_metadata_accounts_v2(
                 true,
             )?;
             metadata.data = compatible_data;
-            assert_collection_update_is_valid(&metadata.collection, &data.collection)?;
+            assert_collection_update_is_valid(false,&metadata.collection, &data.collection)?;
```
roederw, line 278:
```suggestion
            assert_collection_update_is_valid(false, &metadata.collection, &data.collection)?;
```

roederw:
Yeah, this is a bit more solid than the previous check. We're kind of screwed if `isEdition` is ever not correct, but we'd have bigger problems if that were the case.

roederw (approved):

PR #243, opened 2022-02-19
feat: check for zero decimal token accounts  in `findInfoByOwner`

solberenson:
Update `findInfoByOwner` to use `getParsedTokenAccountsByOwner` instead of `getTokenAccountsByOwner`.
This allows checking for zero decimal places when filtering token accounts.


solberenson:
> @solberenson hey thanks for this PR sorry it took us so long to see it, do you still feel its needed?

It's probably no longer necessary as `findInfoByOwner` has been deprecated.

Issue #238, opened 2022-02-18
Update MPL Token Metadata Tests/Scenarios


austbot:

austbot:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Core-Engineering-61eebb497fb4920012651e19/issues/metaplex-foundation/metaplex-program-library/238?planning-poker) with ZenHub @lorisleiva

thlorenz:
Fixed by https://github.com/metaplex-foundation/metaplex-program-library/pull/314

Issue #237, opened 2022-02-18
Generate TS client for Token Metadata with Solita


austbot:
Generate TS client for Token Metadata with Solita, 

Write the script for solita, 
Create any needed wrappers for pda methods or account builders.

Issue #236, opened 2022-02-18
Shank Out Token Metadata Contract


austbot:
Using Shank to build a new IDL for TM contract.

PR #244, opened 2022-02-21
Use Shank and Solita on the Token Metadata program

lorisleiva:
This PR aims to auto-generate the main elements of the Token Metadata JS API by using Shank and Solita.

See #236 and #237.

## Changelog
- Added a relative path shank to the Token Metadata program's dependencies.
- Converted instruction comments into `ShankInstruction` macros.
- Added the `ShankAccount` macro to all accounts defined in the Token Metadata program.
- Added a `api:gen` NPM script that calls a custom script similar to the one used for the Token Vault in [this branch](https://github.com/metaplex-foundation/metaplex-program-library/tree/thlorenz/sdk/token-vault).
- Ran `api:gen` to generate the IDL and the `js/generated` folder.

## Next steps
- Change shank's relative path to a published crate
- Update tests to use the generated API (#238).

## Notes
When converting instruction comments into `ShankInstruction` macros, there were some discrepancies between the documentation and the actual accounts being required by the instructions. Whenever these discrepancies occurred, the methods creating instructions (using `AccountMeta`) were used as the source of truth.


package.json
```diff
@@ -30,7 +30,7 @@
   "license": "MIT",
   "private": true,
   "devDependencies": {
-    "@metaplex-foundation/solita": "0.0.2",
+    "@metaplex-foundation/solita": "0.1.0",
```
thlorenz, line 33: Just remove solita from here (it should be a dev dep of each contract) ... it was a mistake to add it in the first place.

thlorenz (changes requested):
Let's put this in draft mode as we won't be merging this until someone takes the time to build an SDK around it + updates/adds the SDK tests.

thlorenz:
Super-seeded by #314

PR #268, opened 2022-03-03
Espi/ah receipts

austbot:


kespinola:
https://github.com/metaplex-foundation/metaplex-program-library/pull/181

Original PR with details on the change.

PR #284, opened 2022-03-05
fix 0xbba error with whitelist token mint

stegaBOB:
Change the remaining accounts logic to match existing conditionals on frontend. without this, whitelist token mints with presale and burntoken will fail if user has no presale whitelist token


samuelvanderwaal (approved):

roederw:
Wait. This is supposed to fail.

stegaBOB:
Not sure what you mean here

roederw:
Same.

PR #285, opened 2022-03-06
Auction House Receipt JS SDK PDAs and Trade State Seed Fix

kespinola:
### Changes
- Add pda functions for auction house receipts to mpl-auction-house
- Fix seeds for trade state pda.

### DIFF
--- a/auction-house/js/src/AuctionHouseProgram.ts
+++ b/auction-house/js/src/AuctionHouseProgram.ts
@@ -3,12 +3,16 @@ import { PublicKey } from '@solana/web3.js';
 import * as errors from './generated/errors';
 import * as instructions from './generated/instructions';
 import * as accounts from './generated/accounts';
+import BN from 'bn.js';
 
 export class AuctionHouseProgram extends Program {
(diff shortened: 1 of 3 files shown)


auction-house/js/src/AuctionHouseProgram.ts
```diff
@@ -116,8 +120,8 @@ export class AuctionHouseProgram extends Program {
         auctionHouse.toBuffer(),
         treasuryMint.toBuffer(),
         tokenMint.toBuffer(),
-        Buffer.from(price, 'utf8'),
-        Buffer.from(tokenSize, 'utf8'),
+        new BN(price).toArrayLike(Buffer, 'le', 8),
+        new BN(tokenSize).toArrayLike(Buffer, 'le', 8),
```
kespinola, line 124: Normally you'd call bn.toBuffer() but is not available in browsers. Instead use toArrayLike to support browser.

It was incorrect before with strings being used. The seeds are computed based on u8 in the program so needs to be a number in js.

PR #303, opened 2022-03-11
feat: Add exported createMetadataV2 instruction

jacobcreech:
Exposing instruction for CreateMetadataV2 so users can make better packed transactions, leading to a better uX and DX


jacobcreech:
Fixed in #326

PR #306, opened 2022-03-14
fix: use correct Creator type

zaxozhu:


austbot (changes requested):
update solita and regenerate

zaxozhu:
It wasn't too old when this PR was created. Anyway, I will close this and move changes to the Gating PR.

PR #314, opened 2022-03-16
feat: token-metadata update SDK to generated version

thlorenz:
## Summary

Current SDK was moved into `./src/deprecated` and exposed under `deprecated`.

In order to pull out an IDL the Rust program was annotated with shank macros.
The new SDK was generated from that IDL and added to the `./src/generated` folder.

Tests were removed from the SDK and added to the Rust program instead as it was decided that
testing the Rust program via the SDK is not useful and that we don't need to test the SDK
itself since it is generated.

## TypeScript

An `./errors.ts` file is included which integrates `cusper` with the generated Error classes.
This integration will be generated in the future most likely.

The generated code was adapted in order to fix circular `import`. This will be fixed in
the _solita_ code generator so that in the future those edits won't be necessary.

## Rust

A check was added in the create metadata processor in order to ensure that the total number of
uses can never be less than the remaining uses and a test included to ensure that.

## CI/Testing

All existing SDK tests were removed along with the related deps and amman config.
The Integration Test CI workflow for token metadata was also removed.


lorisleiva (dismissed):
It all looks good to me!

Note that I might not be the best person to review the new rust tests but they made sense to me.

austbot (approved):

PR #318, opened 2022-03-19
Upgrade license to AGPLv3

nhanphan:
Part of Metaplex's ethos is to continue driving the growth of open-source, community-driven software in a way that encourages contributions to improve the standards and discourages fragmentation of the ecosystem. As such, we have made the decision to upgrade the software license of our open-source on-chain programs to a license that more closely aligns with these goals.

We'll be leaving this up for public comment for a period of time before considering merging.


roederw (dismissed):

cojo:
Just my two cents, but this seems unnecessarily harmful to Solana's efforts to bring on bigger builders / projects, particularly in the gaming community.

Stuff like this plus "gPA considered harmful; friends don't let friends use gPA" without providing a meaningful alternative are the reasons I think reputation of the network is dropping overall for builders with bigger ambitions.

I bear no personal I'll will towards AGPL, but as a startup founder, it is a big enough flag in my experience for future acquisition talks, etc. that I think folks in similar positions will take a look and basically feel they are forced to redo all your work rather than benefit from it.

cojo:
> > Just my two cents, but this seems unnecessarily harmful to Solana's efforts to bring on bigger builders / projects, particularly in the gaming community.
> > Stuff like this plus "gPA considered harmful; friends don't let friends use gPA" without providing a meaningful alternative are the reasons I think reputation of the network is dropping overall for builders with bigger ambitions.
> > I bear no personal ill will towards AGPL, but as a startup founder, it is a big enough flag in my experience for future acquisition talks, etc. that I think folks in similar positions will take a look and basically feel they are forced to redo all your work rather than benefit from it.
> > EDIT: I will add, just to emphasize the point, that I am specifically saying that I think this change to AGPL will result in effects directly counter to your stated goal of:
> > > discourages fragmentation of the ecosystem.
> > 
> > 
> > Instead, there will be increased fragmentation in the ecosystem as some builders - who cannot afford the risk of fighting to clearly define "derivative work" and whether they have to release their code as well or not - will simply seek an alternative (or look elsewhere entirely).
> 
> Thanks for your time in crafting this response.
> 
> I'll address GPA stuff quickly we are working on alternatives and cannot control the Solana core roadmap. I believe the tweet you are referring too was in jest but with a view to whats coming which is application-specific RPC plugins.
> 
> On AGPL we face people forking our contracts and close sourcing them, then expecting support / us to somehow know if their contract is vulnerable to things we have patched without disclosing source. With the move to AGPL our desire is that these specific core libraries remain open and all lead back to this repo for improvements. We want to give people like yourself a stable and secure foundation to build on top of.

Thanks for your reply as well - I should clarify as well that I understand this reasoning and it totally makes sense to me for the Rust / Cargo programs. 

Where I am concerned things are ambiguous and potentially harmful is with the licensing of the corresponding JS / TS client libs.

Applying AGPL to those, from my perspective, means our entire product is now subject to AGPL and we must release ALL of our source code if we utilize these client libraries (e.g. token-metadata) to speed up our development involving token accounts.

Perhaps this is not your intent? I would have no concerns if a clarification was added specifying continued Apaxhe license for client TS/JS helper libs but AGPL for all Rust / Cargo programs and core code.

nhanphan:
Thanks for your clarification @cojo. Yes our intent is to only apply AGPL to our on-chain programs. The TS clients will retain their existing permissive licenses (Apache or MIT) in their respective directories (e.g. [Candy Machine](https://github.com/metaplex-foundation/metaplex-program-library/blob/license-update/candy-machine/js/LICENSE)).

I'll update the README to clarify.

cojo:
> Thanks for your clarification @cojo. Yes our intent is to only apply AGPL to our on-chain programs. The TS clients will retain their existing permissive licenses (Apache or MIT) in their respective directories (e.g. [Candy Machine](https://github.com/metaplex-foundation/metaplex-program-library/blob/license-update/candy-machine/js/LICENSE)).
> 
> I'll update the README to clarify.

Awesome, thanks for the clarification(s)! Much appreciated. No additional concerns on my side - appreciate all of your transparency (and willingness to discuss) as a team on this!

roederw (approved):

PR #321, opened 2022-03-20
fix(auction-house): use metaplex-token-metadata lib for Metadata struct

samuelvanderwaal:
This PR replaces the `mpl-token-metadata` lib with the older version `metaplex-token-metadata` for importing the Metadata struct to avoid deserialization issues with NFTs that have corrupted metadata. 

There are a number of NFTs in the wild that have corrupted metadata in the "v2" metadata (fields after the `edition_nonce`). This causes issues with deserialization in client-side Rust tools and in Solana programs which use strict typing to do Borsh deserialization. 

I'm working on a long term fix, which requires a change to the `token-metadata` program, but that will take some time to implement. This is a very minimal change which will fix `auction-house`'s deserialization problems in the near term and which will be easy to roll back when the data is solved externally.

### DIFF
--- a/auction-house/program/Cargo.lock
+++ b/auction-house/program/Cargo.lock
@@ -1652,6 +1652,36 @@ dependencies = [
 ]
 
 [[package]]
+name = "metaplex-token-metadata"
+version = "0.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


dborstelmann (approved):

roederw:
I am okay with this as a short term fix for corrupted metadata. Although, we just need to be careful when we add more parameters to the `Token Metadata` that we CANNOT blow away.

roederw (approved):

PR #326, opened 2022-03-22
chore: upgrade shank + solita and regen token-metadata with optional accounts

thlorenz:
## Summary 

Adding latest shank and solita versions which support optional accounts and regenerated
token-metadata SDK to account for optional accounts.

## Steps

Pulled in latest shank + solita which support optional accounts.

Updated token-metadata rust code to add `optional` attribute to optional instruction accounts.

Renerated SDK.

## Side Effects

Since solita code generation overall improved especially around avoidance of circular imports
those changes were applied as part of the code regeneration.

## Optional Account Example

The [Utilize](token-metadata/js/src/generated/instructions/Utilize.ts) instruction now respects the fact that `useAuthorityRecord` and `burner` are optional accounts.

```ts
if (useAuthorityRecord != null) {
  keys.push({
    pubkey: useAuthorityRecord,
    isWritable: true,
    isSigner: false,
  });
}

if (burner != null) {
  if (useAuthorityRecord == null) {
    throw new Error(
      "When providing 'burner' then 'useAuthorityRecord' need(s) to be provided as well.",
    );
  }
  keys.push({
    pubkey: burner,
    isWritable: false,
    isSigner: false,
  });
}
```

## Repo Wide Chores

Solita and beet tooling deps were updated across the workspace in order to keep those versions as much in sync as possible.

### DIFF
--- a/token-metadata/js/idl/mpl_token_metadata.json
+++ b/token-metadata/js/idl/mpl_token_metadata.json
@@ -282,7 +282,8 @@
           "name": "reservationList",
           "isMut": true,
           "isSigner": false,
-          "desc": "(Optional) Reservation List - If present, and you are on this list, you can get an edition number given by your position on the list."
+          "desc": "Reservation List - If present, and you are on this list, you can get an edition number given by your position on the list.",
+          "optional": true
         }
       ],
       "args": [],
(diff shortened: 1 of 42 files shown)


token-metadata/js/idl/mpl_token_metadata.json
```diff
@@ -1165,13 +1166,15 @@
           "name": "useAuthorityRecord",
           "isMut": true,
           "isSigner": false,
-          "desc": "(Optional) Use Authority Record PDA If present the program Assumes a delegated use authority"
+          "desc": "Use Authority Record PDA If present the program Assumes a delegated use authority",
```
samuelvanderwaal, line 1169: Nit:

Use the same formatting as the previous description for clarity:

```
Use Authority Record PDA - If present, etc.
```

samuelvanderwaal (approved):

token-metadata/js/idl/mpl_token_metadata.json
thlorenz: This is generated from the contract description and I basically just removed `(Optional)` there.
I'm trying to keep the changes to a minimum trying to keep PRs as small as possible esp. to contract code.

We can fix those in a later PR as this one focuses solely on fixing optional accounts.

PR #329, opened 2022-03-24
Remove verified check for auction creation

justinkook:


auction/program/src/lib.rs
```diff
@@ -11,4 +11,4 @@ pub mod processor;
 pub const PREFIX: &str = "auction";
 pub const EXTENDED: &str = "extended";
 pub const BIDDER_POT_TOKEN: &str = "bidder_pot_token";
-solana_program::declare_id!("auctxRXPeJoc4817jDhf4HbjnhEcr1cCXenosMhK5R8");
+solana_program::declare_id!("GsgdMuQZnMAg4zCkc86JGSgZqRq55jTb8iBf4omiKnND");
```
austbot, line 14: I cannot merge with this program id, but I'm glad you are testing locally :)

austbot (changes requested):
@justinkook thanks so much for this addition looks like you are adding dutch auctions and allowing auctions to be made without creator verification. 

This feels problematic as it increases the chances of someone selling a fake NFT , what is your reasoning?

PR #331, opened 2022-03-25
Add collections gating feature to the fixed-price-sale market

VadimGrozinok:
## What
This PR contains collection gating feature for the fixed-price-sale program.
## Why
Some of the projects may need early sale phase for group of people who already have NFTs from specific collection in their wallets.
## How
Gating implemented using Metadata collection field. So it simply checks if users has NFT from pointed collection and if this NFT verified with collection authority or not.


brandontulsi:
Existing docs updated here: https://metaplexwiki.atlassian.net/wiki/spaces/METAPLEX/pages/2097161/Fixed-price-sale+program+tech+description

austbot (changes requested):
Good work but we have a few issues, see comments.

fixed-price-sale/program/src/processor/buy.rs
```diff
@@ -167,4 +179,96 @@ impl<'info> Buy<'info> {
 
         Ok(())
     }
+
+    fn verify_gating_token(
+        gate: &Option<GatingConfig>,
+        user_wallet: &AccountInfo<'info>,
+        remaining_accounts: &[AccountInfo<'info>],
```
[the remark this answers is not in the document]
VadimGrozinok: Thanks, the thing is that that accounts will be passed only when gating feature is activated, and as for me code will be more bulky if we pass that accounts separately

fixed-price-sale/program/src/processor/buy.rs
```diff
@@ -167,4 +179,96 @@ impl<'info> Buy<'info> {
...
+            if metadata.key() != metadata_key {
+                return Err(ErrorCode::WrongGatingMetadataAccount.into());
+            }
+
+            if user_token_acc_data.owner != user_wallet.key() {
+                return Err(ErrorCode::WrongOwnerInTokenGatingAcc.into());
+            }
+
+            // extract data from metadata data slices to save computation units
+            let is_collection_present = metadata_data[328];
```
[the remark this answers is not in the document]
VadimGrozinok: Nice catch, thank you a lot

fixed-price-sale/program/src/processor/create_market.rs
```diff
@@ -18,6 +18,7 @@ impl<'info> CreateMarket<'info> {
         pieces_in_one_wallet: Option<u64>,
         start_date: u64,
         end_date: Option<u64>,
+        gating_config: Option<GatingConfig>,
```
austbot, line 21: I dont see where you check the validity of the collection.

austbot (changes requested):
Only one thing left

fixed-price-sale/program/src/processor/create_market.rs
VadimGrozinok: Yes, missed that in CreateMarket instruction
Thanks

fixed-price-sale/program/src/processor/create_market.rs
```diff
@@ -65,6 +66,16 @@ impl<'info> CreateMarket<'info> {
...
                 }
             }
+
+            if remaining_accounts.len() != 1 {
+                return Err(ErrorCode::CollectionMintMissing.into());
+            }
+
+            let collection_mint = &remaining_accounts[0];
+
+            if collection_mint.key != &gating_data.collection || collection_mint.owner != &spl_token::id() {
```
austbot, line 76: SHouldnt the account check be optional if gating_data.collection is not there, or is collection gating required. 
Also you arent checking here if the collection is verified so it will only fail during the buy. Would it make more sense to force the verification before you create the market ?

austbot (commented):
Just a question

fixed-price-sale/program/src/processor/create_market.rs
VadimGrozinok: gating_data.collection is required field.
And not sure that we can check if collection if verified because it's collection mint itself. I mean that token doesn't belong to any collection.
Correct me please if I missed smth

austbot: Oh I understand the flow now.

austbot (approved):

PR #335, opened 2022-03-27
updates Metadata.ts

dietmerc:
error responses are getting trapped (ex. below), adds default empty array to avoid `TypeError: Cannot read property 'map' of null ...  at Function.<anonymous> ../mpl-core/dist/src/Program.js:50:18`

```js
{
  jsonrpc: '2.0',
  result: null,
  error: { code: -32602, message: 'Too many filters provided; max 4' },
  id: 'd899f230-f5e9-42f6-9446-03b06b9720d7'
}
```

### DIFF
--- a/token-metadata/js/src/deprecated/accounts/Metadata.ts
+++ b/token-metadata/js/src/deprecated/accounts/Metadata.ts
@@ -238,7 +238,9 @@ export class Metadata extends Account<MetadataData> {
         .flat()
         .map((account) => Metadata.from(account));
     } else {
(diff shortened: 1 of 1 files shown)


roederw (approved):

thlorenz:
Failing style https://github.com/metaplex-foundation/metaplex-program-library/runs/5738048396?check_suite_focus=true#step:3:120, please fix before merging.

roederw:
@dietmerc can you run a lint to fix this styling problem?

PR #338, opened 2022-03-30
Auction House Scopes and Auctioneers

samuelvanderwaal:
Refactors `auction-house` to support external auctioneer programs. 

Might be able to turn some of the unchecked accounts into `Account<'info, T>` types and then Box them. 

Needs conflicts resolved and master changes merged in as well as needs to be tested with an external Auctioneer program but the design is ready to be reviewed. 

This has **crate** API changes but the **program** API only has new additions, so is backwards compatible.


auction-house/program/src/execute_sale/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+            ata_program: a.ata_program,
+            program_as_signer: a.program_as_signer,
+            rent: a.rent,
+        }
+    }
+}
+
+/// Accounts for the [`execute_sale` handler](auction_house/fn.execute_sale.html).
+#[derive(Accounts, Clone)]
+#[instruction(escrow_payment_bump: u8, free_trade_state_bump: u8, program_as_signer_bump: u8, buyer_price: u64, token_size: u64, auctioneer_pda_bump: u8)]
```
austbot, line 112: We dont really need the bumps anymore with new versions of anchor. 
You could consider removing them.

auction-house/program/src/execute_sale/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    assert_valid_auctioneer_and_scope(
+        &auction_house.key(),
+        &auctioneer_authority.key(),
+        ah_auctioneer_pda,
+        AuthorityScope::ExecuteSale,
+    )?;
+
+    let mut accounts: InstantExecuteSale<'info> = (*ctx.accounts).clone().into();
```
austbot, line 243: due to the size of this struct I'm concerned about busting compute with the clone, is a clone needed can we not pass by reference?

auction-house/program/src/lib.rs
```diff
@@ -1030,6 +454,24 @@ pub mod auction_house {
...
 
+    pub fn buy_with_auctioneer<'info>(
+        ctx: Context<'_, '_, '_, 'info, BuyWithAuctioneer<'info>>,
+        trade_state_bump: u8,
+        escrow_payment_bump: u8,
+        buyer_price: u64,
+        token_size: u64,
+        ah_auctioneer_pda_bump: u8,
+    ) -> ProgramResult {
+        private_bid_with_auctioneer(
```
austbot, line 465: might be good to follow the same import pattern as sell and others I like it

auction-house/program/src/receipt/mod.rs
```diff
@@ -138,14 +138,11 @@ pub fn print_listing_receipt<'info>(
     let seller_trade_state = &prev_instruction_accounts[6];
```
austbot, line 138: Please ensure you update with master on this as there is a patched vuln in the receipts where the account being set is not checked, patched on chain and in master

auction-house/program/src/utils.rs
```diff
@@ -169,15 +169,65 @@ pub fn assert_keys_equal(key1: Pubkey, key2: Pubkey) -> ProgramResult {
     }
 }
 
+#[derive(Debug, Clone)]
+
 pub enum BidType {
-    PublicSale,
-    PrivateSale,
+    InstantPublicSale,
```
austbot, line 175: This could break alot of current indexing .We should keep the same names

auction-house/program/src/utils.rs
```diff
@@ -169,15 +169,65 @@ pub fn assert_keys_equal(key1: Pubkey, key2: Pubkey) -> ProgramResult {
...
+
+pub enum CancelType {
+    InstantCancel,
+    AuctionCancel,
 }
 
 pub fn assert_program_bid_instruction(sighash: &[u8]) -> Result<BidType, ErrorCode> {
     match sighash {
-        [169, 84, 218, 35, 42, 206, 16, 171] => Ok(BidType::PublicSale),
-        [102, 6, 61, 18, 1, 218, 235, 234] => Ok(BidType::PrivateSale),
```
austbot, line 180: Exhibit A on program indexing breakage

auction-house/program/src/utils.rs
```diff
@@ -577,3 +625,39 @@ pub fn assert_valid_trade_state<'a>(
         _ => Err(ErrorCode::DerivedKeyInvalid.into()),
     }
 }
+
+pub fn assert_valid_auctioneer_and_scope(
```
austbot, line 629: seems like an nice and easy method to .... dare I say it unit test

austbot (changes requested):
This looks like a PR from an Engineer who took his time, was careful and cared about quality and it shows. This is an incredible PR and I'm a better man for looking at it. 

I think we cannot change the names of the Accounts structs for existing IXs as much as I want that it will break our contract with the community to not break their processes.

auction-house/program/src/execute_sale/mod.rs
samuelvanderwaal: Is it ok to upgrade Auction House to the new Anchor version? I thought we had some reason to keep it on 0.21.

auction-house/program/src/utils.rs
samuelvanderwaal: 😁  Roger.

auction-house/program/src/bid/mod.rs
```diff
@@ -123,6 +219,40 @@ pub fn private_bid<'info>(
...
+    auction_bid_logic(
+        ctx.accounts.wallet.to_owned(),
+        ctx.accounts.payment_account.to_owned(),
+        ctx.accounts.transfer_authority.to_owned(),
+        ctx.accounts.treasury_mint.to_owned(),
+        ctx.accounts.token_account.to_owned(),
+        ctx.accounts.metadata.to_owned(),
+        ctx.accounts.escrow_payment_account.to_owned(),
+        ctx.accounts.authority.to_owned(),
+        &mut ctx.accounts.auction_house,
```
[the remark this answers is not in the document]
samuelvanderwaal: `accounts` is a `&mut T` so I can't move the auction_house value out of that, and I can't copy it because it's behind a `Box` (to avoid blowing the compute budget). 

This is the error if you remove the `&mut`: 

```
cannot move out of `ctx.accounts.auction_house` which is behind a mutable reference
move occurs because `ctx.accounts.auction_house` has type `Box<anchor_lang::prelude::Account<'_, state::AuctionHouse>>`, which does not implement the `Copy` traitrustc[E0507](https://doc.rust-lang.org/error-index.html#E0507)
```

Is there an issue with using a mutable reference here?

auction-house/program/src/cancel/mod.rs
```diff
@@ -0,0 +1,208 @@
+use anchor_lang::{prelude::*, solana_program::program::invoke, AnchorDeserialize};
+use solana_program::program_memory::sol_memset;
+
+use crate::{constants::*, errors::*, utils::*, AuctionHouse, AuthorityScope, *};
+
+/// Accounts for the [`cancel` handler](auction_house/fn.cancel.html).
+#[derive(Accounts)]
+#[instruction(buyer_price: u64, token_size: u64)]
+pub struct InstantCancel<'info> {
```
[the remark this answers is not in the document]
samuelvanderwaal: Fixed.

auction-house/program/src/utils.rs
samuelvanderwaal: Fixed.

auction-house/program/src/utils.rs
samuelvanderwaal: Not sure I follow what you mean here. I've reverted the enum names but have to add the additional sighash matches to be compatible with receipts.

auction-house/program/src/delegate/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    pub system_program: Program<'info, System>,
+}
+
+pub fn delegate_auctioneer<'info>(
+    ctx: Context<'_, '_, '_, 'info, DelegateAuctioneer<'info>>,
+    _ah_auctioneer_pda_bump: u8,
+    scopes: Box<Vec<AuthorityScope>>,
+) -> ProgramResult {
+    if scopes.len() > MAX_NUM_SCOPES {
```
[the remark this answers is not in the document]
samuelvanderwaal: You're correct. This is fixed in 8ca30d74a6634eb3f2e265602eab8310f601a80d.

auction-house/program/src/delegate/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+        return Err(ErrorCode::TooManyScopes.into());
+    }
+
+    let auction_house = &mut ctx.accounts.auction_house;
+    auction_house.has_auctioneer = true;
+
+    let auctioneer = &mut ctx.accounts.ah_auctioneer_pda;
+    auctioneer.authority = ctx.accounts.auctioneer_authority.key();
+    auctioneer.auction_house = ctx.accounts.auction_house.key();
+    auctioneer.scopes = *scopes;
```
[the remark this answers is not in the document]
samuelvanderwaal: Fixed in 8ca30d74a6634eb3f2e265602eab8310f601a80d.

auction-house/program/src/utils.rs
samuelvanderwaal: This is actually a pain to unit test because of the AccountInfo business. None of the other assert methods are unit tested so unless you really think this needs to be done I'm going to skip it and rely on integration tests.

PR #317, opened 2022-03-18
Entangle full supply

levicook:
This pull request supports entangling some pre-metaplex standard NFTs. It specifically enables swapping the entire supply of one token for another. Basically this allows communities to swap non-standard tokens with standard tokens.

Client side / cli changes can be found here: https://github.com/metaplex-foundation/metaplex/pull/2005

PR #365, opened 2022-04-08
Revert "Entangle full supply (#317)"

levicook:
This reverts commit af62203af0a31e82dfb4fb841db2c0ae8d62a666.

Unclear on why this change broke existing entanglements, but recommending we revert and investigate since there's nothing currently entangled that requires the new functionality this introduced

### DIFF
--- a/token-entangler/program/src/lib.rs
+++ b/token-entangler/program/src/lib.rs
@@ -1,14 +1,16 @@
 pub mod utils;
 
-use crate::utils::*;
-use anchor_lang::{
-    prelude::*,
-    solana_program::program::{invoke, invoke_signed},
-    AnchorDeserialize, AnchorSerialize,
-};
-use anchor_spl::{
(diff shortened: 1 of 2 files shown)


levicook:
The unexpected error is

https://github.com/project-serum/anchor/blob/51aeb08ae1c93f9f759c6e244e3fa724c9a916a7/lang/src/error.rs#L81

    #[msg("The given account is owned by a different program than expected")]
    AccountOwnedByWrongProgram,

Here's an example for reference:
https://solscan.io/tx/vC8xybdwe7Z6KChxffZZwa1e3HNbyz9whZvrSVKgbHUTg3nzugAgNm41juAHadfQAcyj2X3hHUvNY9BMWcvn3hr

roederw (approved):

roederw:
Reverting until I can figure out the problem.

PR #370, opened 2022-04-12
Updating to Solana 1.9.15 and anchor 0.24.1.

blockiosaurus:


roederw (approved):

roederw (dismissed):

blockiosaurus:
Is that the recommended workflow? I always use cargo build-bpf but can add the anchor // CHECKs if need be.

stegaBOB:
Worst case scenario we should be able to run `anchor build --skip-lint` if we want to avoid the checks, but I think the bump seed thing should still cause the builds to fail? I could be wrong though.

stegaBOB:
Yeah just checking, it looks like the build is failing for the token entangler now due to the bump target being provided. 
image

stegaBOB:
How does the cargo build-bpf succeed when the anchor build fails for that? Are the macros still being expanded, or is it panicking somewhere but not being caught? For future reference, are these builds still valid?

roederw:
@stegaBOB @blockiosaurus I'll add the `token-entangler` fixes now.

roederw:
It's an addition on the `anchor build` side of things. Not connected to `cargo build-bpf`

blockiosaurus:
Oh thanks! I guess we don't have token-entangler tests?

blockiosaurus:
I have no objections to adding CHECKs and it's the right thing to do. I'm just trying to be expedient due to other priorities.

stegaBOB:
I believe we do, but I don't think any of them use the anchor build (all probably use cargo build-bpf). Perhaps this is something that should be changed?

blockiosaurus:
Most stuff we get through anchor build is also exposed through other Anchor subcommands. My first instinct is to continue using cargo's more mature build system and explicitly call anchor's subcommands to fill in the gaps.

roederw:
I'm just going to quickly get `token-entangler` bumped and deal with the other things later. We need this bump.

roederw (approved):

roederw:
> Yeah just checking, it looks like the build is failing for the token entangler now due to the bump target being provided. image

@stegaBOB should be good to go now.

PR #372, opened 2022-04-13
chore: expose full API including errors

thlorenz:
## Expose generated API

- currently all hangs off `CandyMachineProgram` which makes less sense
  since we also export types + classes here
- also this fixes the issues of `./generated/types` not being
  exposed at all
- I kept `CandyMachineProgram` around for now, but ideally we'd remove
  that as well and switch to the approach that token-metadata uses

## Errors

- deps: add cusper dep
- api: expose errors and cusper based error API

### DIFF
--- a/candy-machine/js/package.json
+++ b/candy-machine/js/package.json
@@ -41,6 +41,7 @@
   "dependencies": {
     "@metaplex-foundation/beet": "^0.1.0",
     "@metaplex-foundation/beet-solana": "^0.1.1",
+    "@metaplex-foundation/cusper": "^0.0.2",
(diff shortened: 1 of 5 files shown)


thlorenz:
For context it now allows me to actually get the types like this: https://github.com/metaplex-foundation/js-next/blob/thlorenz/feat/candy-machine-create/src/modules/candy-machine/models/CandyMachine.ts#L2-L8

```ts
import {
  CandyMachineData,
  EndSettings,
  GatekeeperConfig,
  HiddenSettings,
  WhitelistMintSettings,
} from '@metaplex-foundation/mpl-candy-machine'
```

samuelvanderwaal (approved):

PR #373, opened 2022-04-13
Removes mut constraint from non writable account

stegaBOB:
Removes the `mut` account constraint from the merkledistributor account in the `claim` instruction, removing the writable accounts bottleneck from high volume claims.


samuelvanderwaal (approved):

samuelvanderwaal:
@stegaBOB what's the status on this?

stegaBOB:
Not entirely sure. I don't think this would result in a breaking change, and it feels like a safe change to make and should only improve performance. I think it should be merged in at some point, just not sure if now is the right time

PR #374, opened 2022-04-13
Gumdrop: remove signer constraint from base on creation

stegaBOB:
Removes unnecessary signer from NewDistributor context


roederw (approved):

stegaBOB:
We don't actually need this anymore. Closing

PR #401, opened 2022-04-21
chore: update generated api to latest program using anchor 0.24.2

thlorenz:
- chore: api:gen script adds anchor origin and version to metadata
- api: regen IDL with anchor 0.24.2
- sdk: regen API from latest anchor IDL


auction-house/js/idl/auction_house.json
```diff
@@ -179,7 +179,7 @@
         },
         {
           "name": "payer",
-          "isMut": false,
+          "isMut": true,
```
thlorenz, line 182: I'm hoping this change is intended

austbot: yes this is an anchor feature now

PR #407, opened 2022-04-26
Fix Metadata Corruption

samuelvanderwaal:
This PR adds a new custom deserialization function for the `Metadata` struct that catches invalid v1.1 data (TokenStandard, Collection, Uses) and sets their respective values to `None` if invalid data is found. This allows the `update_metadata_accounts_v2` handler to call this function and successfully deserialize NFTs with corrupted metadata allow them to then be updated.

It uses this function as a custom BorshDeserialization implementation for `Metadata` to allow client-side programs to be able to decode corrupted metadata.

In addition, it zeros out all metadata before serializing to ensure future metadata fields won't be created with corrupted data. 

Finally, it adds an extra case to the check for updating `primary_sale_happened` and `set_immutable` so that when existing values are passed up, the data stays the same and there's no error.


austbot (dismissed):

roederw (dismissed):

danenbm (approved):
Thanks for adding comments, tests, and the format cleanup!

roederw (approved):

PR #408, opened 2022-04-26
fix: Requested changes. Add ability to remove pieces limit in 'Change…

bezrazli4n0:
…Market' instruction. Changed pieces condition to correct in 'Buy' instruction.


fixed-price-sale/program/src/processor/change_market.rs
```diff
@@ -58,10 +58,7 @@ impl<'info> ChangeMarket<'info> {
             market.price = new_price;
         }
 
-        // Check is required, because we can overwrite existing value
-        if let Some(new_pieces_in_one_wallet) = new_pieces_in_one_wallet {
-            market.pieces_in_one_wallet = Some(new_pieces_in_one_wallet);
-        }
+        market.pieces_in_one_wallet = new_pieces_in_one_wallet;
```
VadimGrozinok, line 61: I think here make sense to change type of `new_pieces_in_one_wallet` variable. I mean it would be better declare it as `Option<Enum{Value(u64), Unlim}>`. It will save consistency and decrease chance accidentally change this field.

VadimGrozinok (changes requested):

fixed-price-sale/program/src/processor/change_market.rs
bezrazli4n0: Should type changing affect whole contract or just single instruction 'ChangeMarket'?

VadimGrozinok (dismissed):

fixed-price-sale/program/src/processor/change_market.rs
```diff
@@ -8,7 +13,7 @@ impl<'info> ChangeMarket<'info> {
         new_description: Option<String>,
         mutable: Option<bool>,
         new_price: Option<u64>,
-        new_pieces_in_one_wallet: Option<u64>,
+        pieces_mode: Option<PiecesMode>,
```
austbot, line 16: removing a arg is not backward compatible and will cause failures

fixed-price-sale/program/src/lib.rs
```diff
@@ -66,15 +69,10 @@ pub mod fixed_price_sale {
         new_description: Option<String>,
         mutable: Option<bool>,
         new_price: Option<u64>,
-        new_pieces_in_one_wallet: Option<u64>,
```
austbot, line 69: this is not a backward compatible change, you must leave the old argument, mark it as deprecated and announce that this is changing and give people time to switch or make a new instruction

austbot (changes requested):

fixed-price-sale/program/src/processor/change_market.rs
austbot: this will break users like @robbierogers in production if we do this without fair warning

VadimGrozinok: oh, yes, forgot about backward compatibility, thanks
so I think we'll new instruction with different arguments and mark current instruction as deprecated
such as adding new one argument will also break current front-end

VadimGrozinok:
yes, because it's an issue
@bezrazli4n0 resolve please the conflicts and update the PR

PR #414, opened 2022-04-28
Allow Metadata updates when it contains a Verified Collection

roederw:
Updating `Metadata` accounts with a `verified` collection requires that users "un verify" their collection to properly update. This fix allows for updates, as long as you maintain the verified collection state.

### DIFF
--- a/token-metadata/program/tests/update_metadata_account_v2.rs
+++ b/token-metadata/program/tests/update_metadata_account_v2.rs
@@ -130,7 +130,7 @@ mod update_metadata_account_v2 {
 
         let metadata = test_metadata.get_data(&mut context).await;
 
-        assert_eq!(metadata.data.name, puffed_updated_name,);
+        assert_eq!(metadata.data.name, puffed_updated_name);
         assert_eq!(metadata.data.symbol, puffed_symbol);
         assert_eq!(metadata.data.uri, puffed_uri);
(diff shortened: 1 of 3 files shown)


token-metadata/program/tests/update_metadata_account_v2.rs
```diff
@@ -141,6 +143,114 @@ mod update_metadata_account_v2 {
...
+
+        let tx2 = Transaction::new_signed_with_payer(
+            &[instruction::update_metadata_accounts_v2(
+                id(),
+                test_metadata.pubkey,
+                context.payer.pubkey().clone(),
+                None,
+                Some(DataV2 {
+                    name: "Test".to_string(),
+                    symbol: "TST".to_string(),
```
austbot, line 233: please change some data to be verified its changed

token-metadata/program/tests/update_metadata_account_v2.rs
```diff
@@ -378,6 +488,123 @@ mod update_metadata_account_v2 {
...
+                    uri: "uri".to_string(),
+                    creators: None,
+                    seller_fee_basis_points: 10,
+                    collection: Some(Collection {
+                        key: collection_master_edition_account.pubkey,
```
austbot, line 583: make this more clear that its like a fake collection key

austbot (dismissed):
only one test note but its optional

PR #417, opened 2022-04-29
Its time we did something about bots

austbot:
# Step 0.1 of bot destruction:

TLDR; CHARGE BOTs .01 SOL (could increase)

This is not the end solution but it will be effective in discouraging bots.

## Rationale
Candymachine has one of the lowest success rates for programs on solana
image

A very high number of those transactions are clear cases where a bot is doing something bad, trying to mint before the mint is open, minting after the candy machine is empty, minting without a WL token 
Example: 

This is an image of a CandyMachineNotLive error.
image

## Content

This PR adds a hard-coded bot tax of 0.01 sol (we may increase this) and it will charge the wallet this amount or all of their funds whatever is higher under the following conditions:
The funds are sent to the Token Metadata program account, this is to avoid any client changes needed to adopt this change. More pervasive changes are needed to fully and more accurately solve this. 
If a bot wallet does not contain enough for the bot fee and hits a failure mode from above we drain their wallet entirely. Then its up to the efficient Solana TXN system to kick out unfunded transactions.

### Specific Fee Charging Cases

1. Trying to mint when the candy machine is not live before or after
2. Trying to mint when there are no items left in the candy machine
3. Calling Candy Machine via CPI when not using gumdrop
4. Crafting a transaction where mint or set collection is not the last ix
5. Using the wrong collection id than the configured one for the candy machine
6. Setting Collection IX with a mismatched mint than what was just minted
7. Signer Payer mismatch with collection set and mint ix
8. Suspicious Transactions where disallowed programs are used
9. Trying to mint on an AllowList candy machine with no allow list token

### Cons

There is a chance that a real user could hit one of these cases, especially in # 2. But we think these will not be frequent.
This is not a fool proof fix, and it will not completely stop congestion but we belive it has a substantial enough impact to attempt it.

## Request for Feedback

Please give us feedback below, especially if you have data from your platform regarding the regularity of # 2 or if you are valid CPI user of candy machine.

### DIFF
--- a/candy-machine/program/src/lib.rs
+++ b/candy-machine/program/src/lib.rs
@@ -6,6 +6,7 @@ use {
         assert_initialized, assert_is_ata, assert_keys_equal, assert_owned_by,
         assert_valid_go_live, spl_token_burn, spl_token_transfer, TokenBurnParams,
         TokenTransferParams,
+        punish_bots,
     },
     anchor_lang::{
         prelude::*,
@@ -39,6 +40,14 @@ const EXPIRE_OFFSET: i64 = 10 * 60;
 const PREFIX: &str = "candy_machine";
 // here just in case solana removes the var
 const BLOCK_HASHES: &str = "SysvarRecentB1ockHashes11111111111111111111";
+const BOT_FEE: u64 = 1000000;
+
+pub const GUMDROP_ID: Pubkey = Pubkey::new_from_array([
+    0x0a, 0x27, 0x53, 0xf9, 0x40, 0xcb, 0xd3, 0xc6, 0xfa, 0x7d, 0xf3, 0xc0, 0x2e, 0x24, 0xb5, 0x77,
+    0x08, 0xc4, 0x87, 0xd3, 0x52, 0x15, 0x7d, 0x80, 0x4b, 0x0c, 0x5e, 0xb0, 0xec, 0x27, 0xcf, 0x41
+]);
+
+
 #[program]
 pub mod candy_machine {
     use super::*;
@@ -58,6 +67,22 @@ pub mod candy_machine {
         //Account name the same for IDL compatability
         let recent_slothashes = &ctx.accounts.recent_blockhashes;
         let instruction_sysvar_account = &ctx.accounts.instruction_sysvar_account;
+        let instruction_sysvar_account_info = instruction_sysvar_account.to_account_info();
+        let instruction_sysvar = instruction_sysvar_account_info.data.borrow();
+        let current_ix = get_instruction_relative(0, &instruction_sysvar_account_info).unwrap();
+        /// Restrict Who can call Candy Machine via CPI
+        if current_ix.program_id != candy_machine::id() || current_ix.program_id != GUMDROP_ID {
+            punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+            return Ok(());
+        }
+        let next_ix = get_instruction_relative(1, &instruction_sysvar_account_info);
+        if next_ix.is_ok() {
+            let discriminator = &next_ix.unwrap().data[0..8];
+            if discriminator != [103, 17, 200, 25, 118, 95, 125, 61] {
+                punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                return Ok(());
+            }
+        }
         if recent_slothashes.key().to_string() == BLOCK_HASHES {
             msg!("recent_blockhashes is deprecated and will break soon");
         }
@@ -66,18 +91,24 @@ pub mod candy_machine {
         {
             return Err(ErrorCode::IncorrectSlotHashesPubkey.into());
         }
+
         let mut price = candy_machine.data.price;
         if let Some(es) = &candy_machine.data.end_settings {
             match es.end_setting_type {
                 EndSettingType::Date => {
                     if clock.unix_timestamp > es.number as i64 {
                         if ctx.accounts.payer.key() != candy_machine.authority {
-                            return Err(ErrorCode::CandyMachineNotLive.into());
+                            punish_bots(ErrorCode::CandyMachineNotLive, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                            return Ok(());
                         }
                     }
                 }
                 EndSettingType::Amount => {
                     if candy_machine.items_redeemed >= es.number {
+                        if ctx.accounts.payer.key() != candy_machine.authority {
+                            punish_bots(ErrorCode::CandyMachineEmpty, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                            return Ok(());
+                        }
                         return Err(ErrorCode::CandyMachineNotLive.into());
                     }
                 }
@@ -191,7 +222,8 @@ pub mod candy_machine {
                                 if ctx.accounts.payer.key() != candy_machine.authority
                                     && !ws.presale
                                 {
-                                    return Err(ErrorCode::CandyMachineNotLive.into());
+                                    punish_bots(ErrorCode::CandyMachineNotLive, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                                    return Ok(());
                                 }
                             }
                             Some(val) => {
@@ -199,7 +231,8 @@ pub mod candy_machine {
                                     && ctx.accounts.payer.key() != candy_machine.authority
                                     && !ws.presale
                                 {
-                                    return Err(ErrorCode::CandyMachineNotLive.into());
+                                    punish_bots(ErrorCode::CandyMachineNotLive, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                                    return Ok(());
                                 }
                             }
                         }
@@ -212,7 +245,8 @@ pub mod candy_machine {
                             // A non-presale whitelist with no discount price is a forced whitelist
                             // If a pre-sale has no discount, its no issue, because the "discount"
                             // is minting first - a presale whitelist always has an open post sale.
-                            return Err(ErrorCode::NoWhitelistToken.into());
+                            punish_bots(ErrorCode::NoWhitelistToken, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                            return Ok(());
                         }
                         assert_valid_go_live(payer, clock, candy_machine)?;
                         if ws.mode == WhitelistMintMode::BurnEveryTime {
@@ -225,7 +259,8 @@ pub mod candy_machine {
                         // A non-presale whitelist with no discount price is a forced whitelist
                         // If a pre-sale has no discount, its no issue, because the "discount"
                         // is minting first - a presale whitelist always has an open post sale.
-                        return Err(ErrorCode::NoWhitelistToken.into());
+                        punish_bots(ErrorCode::NoWhitelistToken, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                        return Ok(());
                     }
                     if ws.mode == WhitelistMintMode::BurnEveryTime {
                         remaining_accounts_counter += 2;
@@ -395,9 +430,6 @@ pub mod candy_machine {
             ],
             &[&authority_seeds],
         )?;
-        let instruction_sysvar_account_info = instruction_sysvar_account.to_account_info();
-
-        let instruction_sysvar = instruction_sysvar_account_info.data.borrow();
 
         let mut idx = 0;
         let num_instructions = read_u16(&mut idx, &instruction_sysvar)
@@ -421,9 +453,11 @@ pub mod candy_machine {
                 && program_id != associated_token
             {
                 msg!("Transaction had ix with program id {}", program_id);
-                return Err(ErrorCode::SuspiciousTransaction.into());
+                punish_bots(ErrorCode::SuspiciousTransaction, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+                return Ok(());
             }
         }
+
         Ok(())
     }
 
@@ -435,13 +469,15 @@ pub mod candy_machine {
                 "Transaction had ix with program id {}",
                 &previous_instruction.program_id
             );
-            return Err(ErrorCode::SuspiciousTransaction.into());
+            punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+            return Ok(());
         }
 
         let discriminator = &previous_instruction.data[0..8];
         if discriminator != [211, 57, 6, 167, 15, 219, 35, 251] {
             msg!("Transaction had ix with data {:?}", discriminator);
-            return Err(ErrorCode::SuspiciousTransaction.into());
+            punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+            return Ok(());
         }
 
         let mint_ix_accounts = previous_instruction.accounts;
@@ -458,11 +494,13 @@ pub mod candy_machine {
                 mint_ix_cm,
                 candy_key
             );
-            return Err(ErrorCode::SuspiciousTransaction.into());
+            punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+            return Ok(());
         }
         if &mint_ix_cm != &candy_key {
             msg!("Candy Machine with pubkey {} does not match the mint ix Candy Machine with pubkey {}", mint_ix_cm, candy_key);
-            return Err(ErrorCode::SuspiciousTransaction.into());
+            punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+            return Ok(());
         }
         if mint_ix_metadata != metadata {
             msg!(
@@ -470,7 +508,8 @@ pub mod candy_machine {
                 mint_ix_metadata,
                 metadata
             );
-            return Err(ErrorCode::SuspiciousTransaction.into());
+            punish_bots(ErrorCode::SuspiciousTransaction, &ctx.accounts.payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
+            return Ok(());
         }
 
         let collection_pda = &ctx.accounts.collection_pda;
@@ -599,8 +638,8 @@ pub mod candy_machine {
                 .ok_or(ErrorCode::NumericalOverflowError)?;
             let my_position_in_vec = bit_mask_vec_start
                 + position
-                    .checked_div(8)
-                    .ok_or(ErrorCode::NumericalOverflowError)?;
+                .checked_div(8)
+                .ok_or(ErrorCode::NumericalOverflowError)?;
             let position_from_right = 7 - position
                 .checked_rem(8)
                 .ok_or(ErrorCode::NumericalOverflowError)?;
@@ -866,10 +905,10 @@ fn get_space_for_candy(data: CandyMachineData) -> core::result::Result<usize, Pr
             + (data.items_available as usize) * CONFIG_LINE_SIZE
             + 8
             + 2 * ((data
-                .items_available
-                .checked_div(8)
-                .ok_or(ErrorCode::NumericalOverflowError)?
-                + 1) as usize)
+            .items_available
+            .checked_div(8)
+            .ok_or(ErrorCode::NumericalOverflowError)?
+            + 1) as usize)
     };
 
     Ok(num)
@@ -880,7 +919,7 @@ fn get_space_for_candy(data: CandyMachineData) -> core::result::Result<usize, Pr
 #[instruction(data: CandyMachineData)]
 pub struct InitializeCandyMachine<'info> {
     /// CHECK: account constraints checked in account trait
-    #[account(zero, rent_exempt = skip, constraint = candy_machine.to_account_info().owner == program_id && candy_machine.to_account_info().data_len() >= get_space_for_candy(data)?)]
+    #[account(zero, rent_exempt = skip, constraint = candy_machine.to_account_info().owner == program_id && candy_machine.to_account_info().data_len() >= get_space_for_candy(data) ?)]
     candy_machine: UncheckedAccount<'info>,
     /// CHECK: wallet can be any account and is not written to or read
     wallet: UncheckedAccount<'info>,
@@ -952,7 +991,7 @@ pub struct RemoveCollection<'info> {
     #[account(has_one = authority)]
     candy_machine: Account<'info, CandyMachine>,
     authority: Signer<'info>,
-    #[account(mut, seeds = [b"collection".as_ref(), candy_machine.to_account_info().key.as_ref()], bump, close=authority)]
+    #[account(mut, seeds = [b"collection".as_ref(), candy_machine.to_account_info().key.as_ref()], bump, close = authority)]
     collection_pda: Account<'info, CollectionPDA>,
     /// CHECK: account checked in CPI
     metadata: UncheckedAccount<'info>,
@@ -995,7 +1034,7 @@ pub struct MintNFT<'info> {
     )]
     candy_machine: Box<Account<'info, CandyMachine>>,
     /// CHECK: account constraints checked in account trait
-    #[account(seeds=[PREFIX.as_bytes(), candy_machine.key().as_ref()], bump=creator_bump)]
+    #[account(seeds = [PREFIX.as_bytes(), candy_machine.key().as_ref()], bump = creator_bump)]
     candy_machine_creator: UncheckedAccount<'info>,
     payer: Signer<'info>,
     /// CHECK: wallet can be any account and is not written to or read
@@ -1069,7 +1108,9 @@ pub struct CandyMachine {
     // here there is a borsh vec u32 indicating number of bytes in bitmask array.
     // here there is a number of bytes equal to ceil(max_number_of_lines/8) and it is a bit mask used to figure out when to increment borsh vec u32
 }
+
 const COLLECTION_PDA_SIZE: usize = 8 + 64;
+
 /// Collection PDA account
 #[account]
 #[derive(Default, Debug)]
@@ -1150,7 +1191,7 @@ pub const CONFIG_ARRAY_START: usize = 8 + // key
     10 + // end settings
     4 + MAX_SYMBOL_LENGTH + // u32 len + symbol
     2 + // seller fee basis points
-    4 + MAX_CREATOR_LIMIT*MAX_CREATOR_LEN + // optional + u32 len + actual vec
+    4 + MAX_CREATOR_LIMIT * MAX_CREATOR_LEN + // optional + u32 len + actual vec
     8 + //max supply
     1 + // is mutable
     1 + // retain authority
@@ -1194,15 +1235,15 @@ pub fn get_good_index(
         + (items_available) * CONFIG_LINE_SIZE
         + 4
         + items_available
-            .checked_div(8)
-            .ok_or(ErrorCode::NumericalOverflowError)?
+        .checked_div(8)
+        .ok_or(ErrorCode::NumericalOverflowError)?
         + 4;
 
     while taken > 0 && index_to_use < items_available {
         let my_position_in_vec = bit_mask_vec_start
             + index_to_use
-                .checked_div(8)
-                .ok_or(ErrorCode::NumericalOverflowError)?;
+            .checked_div(8)
+            .ok_or(ErrorCode::NumericalOverflowError)?;
         /*msg!(
             "My position is {} and value there is {}",
             my_position_in_vec,
@@ -1327,6 +1368,7 @@ pub fn get_config_line<'info>(
 
 /// Individual config line for storing NFT data pre-mint.
 pub const CONFIG_LINE_SIZE: usize = 4 + MAX_NAME_LENGTH + 4 + MAX_URI_LENGTH;
+
 #[derive(AnchorSerialize, AnchorDeserialize, Debug)]
 pub struct ConfigLine {
     pub name: String,
--- a/candy-machine/program/src/utils.rs
+++ b/candy-machine/program/src/utils.rs
@@ -1,3 +1,4 @@
+use std::cmp;
 use anchor_lang::prelude::{Signer, Sysvar};
 
 use {
@@ -5,6 +6,7 @@ use {
     anchor_lang::{
         prelude::{Account, AccountInfo, Clock, ProgramError, ProgramResult, Pubkey},
         solana_program::{
+            msg,
             program::invoke_signed,
             program_pack::{IsInitialized, Pack},
         },
@@ -69,6 +71,18 @@ pub struct TokenTransferParams<'a: 'b, 'b> {
     /// token_program
     /// CHECK: account checked in CPI
     pub token_program: AccountInfo<'a>,
+}
+
+pub fn punish_bots(err: ErrorCode, bot_account: &AccountInfo, payment_account: &AccountInfo, fee: u64) -> Result<(), ProgramError> {
+    msg!("Error: {}, Candy Machine Botting is taxed at {:?} lamports", err.to_string(), fee);
+    let bot_lamp = bot_account.lamports();
+    let mut final_fee = fee;
+    if bot_lamp < fee {
+        final_fee = bot_lamp;
+    }
+    **bot_account.try_borrow_mut_lamports()? = bot_account.lamports() - final_fee;
+    **payment_account.try_borrow_mut_lamports()? = payment_account.lamports() + final_fee;
+    Ok(())
 }
 
 #[inline(always)]
--- a/candy-machine/program/src/utils.rs
+++ b/candy-machine/program/src/utils.rs
@@ -75,11 +75,7 @@ pub struct TokenTransferParams<'a: 'b, 'b> {
 
 pub fn punish_bots(err: ErrorCode, bot_account: &AccountInfo, payment_account: &AccountInfo, fee: u64) -> Result<(), ProgramError> {
     msg!("Error: {}, Candy Machine Botting is taxed at {:?} lamports", err.to_string(), fee);
-    let bot_lamp = bot_account.lamports();
-    let mut final_fee = fee;
-    if bot_lamp < fee {
-        final_fee = bot_lamp;
-    }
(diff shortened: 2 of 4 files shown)


candy-machine/program/src/utils.rs
```diff
@@ -71,6 +72,13 @@ pub struct TokenTransferParams<'a: 'b, 'b> {
     pub token_program: AccountInfo<'a>,
 }
 
+pub fn punish_bots(err: ErrorCode, bot_account: &AccountInfo, payment_account: &AccountInfo, fee: u64) -> Result<(), ProgramError> {
+    msg!("Error: {}, Candy Machine Botting is taxed at {:?} lamports", err.to_string(), fee);
+    **bot_account.try_borrow_mut_lamports()? = payment_account.lamports() - fee;
+    **bot_account.try_borrow_mut_lamports()? = payment_account.lamports() - fee;
```
mvines, line 78: probably mean to add the punishment to `payment_account`?

candy-machine/program/src/lib.rs
```diff
@@ -72,12 +77,17 @@ pub mod candy_machine {
                 EndSettingType::Date => {
                     if clock.unix_timestamp > es.number as i64 {
                         if ctx.accounts.payer.key() != candy_machine.authority {
-                            return Err(ErrorCode::CandyMachineNotLive.into());
+                            punish_bots(ErrorCode::CandyMachineNotLive, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
```
dietmerc, line 80: 👀 💀

candy-machine/program/src/lib.rs
```diff
@@ -39,6 +40,10 @@ const EXPIRE_OFFSET: i64 = 10 * 60;
 const PREFIX: &str = "candy_machine";
 // here just in case solana removes the var
 const BLOCK_HASHES: &str = "SysvarRecentB1ockHashes11111111111111111111";
+const BOT_FEE: u64 = 1000000;
```
mvines, line 43: Should we make this higher.  Like ◎1?

t-nelson: v2 should :seedling:

candy-machine/program/src/utils.rs
```diff
@@ -71,6 +72,13 @@ pub struct TokenTransferParams<'a: 'b, 'b> {
     pub token_program: AccountInfo<'a>,
 }
 
+pub fn punish_bots(err: ErrorCode, bot_account: &AccountInfo, payment_account: &AccountInfo, fee: u64) -> Result<(), ProgramError> {
+    msg!("Error: {}, Candy Machine Botting is taxed at {:?} lamports", err.to_string(), fee);
+    **bot_account.try_borrow_mut_lamports()? = payment_account.lamports() - fee;
+    **payment_account.try_borrow_mut_lamports()? = payment_account.lamports() + fee;
```
t-nelson, line 78: prolly want checked math?

t-nelson (commented):
we must make them pay

candy-machine/program/src/utils.rs
mvines: oh yeah, and if there's only `fee - 1` in the account then drain that instead of erroring out

roederw:
I would like a more open discussion about this change. Obviously, I _could_ solve a lot of problems related to spamming Txs. BUT I want to make sure a broader audience is aware, and onboard, with this change. I could really backfire.

mvines:
It would be very nice to see documentation, somewhere, describing the conditions under which you'll be classified as a bot and punished.  If it's clear that normal human actions during a mint are not going to trigger these cases, then we can expect that the only ones arguing against such a change will be the bot operators themselves

candy-machine/program/src/lib.rs
```diff
@@ -191,15 +201,17 @@ pub mod candy_machine {
                                 if ctx.accounts.payer.key() != candy_machine.authority
                                     && !ws.presale
                                 {
-                                    return Err(ErrorCode::CandyMachineNotLive.into());
+                                    punish_bots(ErrorCode::CandyMachineNotLive, &payer, &ctx.accounts.token_metadata_program, BOT_FEE)?;
```
stanleycyang, line 204: crush 'em all

MarkSackerberg:
Looking at the code i currently (2 AM) only see bots being punished. 

normal failing transactions (= not going through) will just never be seen on chain
No WL token should ot be possible for normal frontend users
Minting before go live should not be possible for normal frontend users
Additional instructions in the tx (suspiciousTx) should only happen for bots

But that change should not be implemented too fast otherwise it will backfire.
Let's discuss it everywhere where it is possible please and take some time

candy-machine/program/src/lib.rs
bdcannon: do it

stanleycyang: make em pay

joebuild:
I'm not super familiar with CM, but if I'm a botter what's to stop me from adding a follow-up instruction to my mint transaction that checks:
1. that my account balance has been reduced the expected amount (the mint price)
2. that i have in fact received an NFT
and if the above conditions are not met then just throw an error and revert the whole transaction

stegaBOB:
> I'm not super familiar with CM, but if I'm a botter what's to stop me from adding a follow-up instruction to my mint transaction that checks:
> 
>     1. that my account balance has been reduced the expected amount (the mint price)
> 
>     2. that i have in fact received an NFT
> 
> 
> and if the above conditions are not met then just throw an error and revert the whole transaction

The candy machine already checks to make sure only instructions from trusted/necessary programs are included in the mint transaction in order to prevent gaming the randomness. That would also prevent botters from chaining that additional check on.

jkbpvsc:
Won't this have a minimal effect on duplicate transactions?
If a bot sends 1k duplicate txs, won't it be penalized for just a single failed mint?

jkbpvsc:
> > I'm not super familiar with CM, but if I'm a botter what's to stop me from adding a follow-up instruction to my mint transaction that checks:
> > ```
> > 1. that my account balance has been reduced the expected amount (the mint price)
> > 
> > 2. that i have in fact received an NFT
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and if the above conditions are not met then just throw an error and revert the whole transaction
> 
> The candy machine already checks to make sure only instructions from trusted/necessary programs are included in the mint transaction in order to prevent gaming the randomness. That would also prevent botters from chaining that additional check on.

I think CM verifies that there isn't any footer ixs.

candy-machine/program/src/lib.rs
samuelvanderwaal: There are edge cases where legitimate users can get caught through no fault of their own (badly designed mint front-end) so it seems important not to impose too onerous a cost on them while trying to punish actual botters.

staccDOTsol:
Hi :) hydra? @austbot

nikhayes:
> Won't this have a minimal effect on duplicate transactions? If a bot sends 1k duplicate txs, won't it be penalized for just a single failed mint?

I don't think duplicates are as big of a problem anymore. Bots evolved after the better dedupe filter was added and they started sending higher amounts of unique txns instead. Leo from Jump/Certus one posted this data in mb-validators:

"TPU stats between 20:01 - 20:06 
2022/04/26 21:01:35 30697035 packets
2022/04/26 21:01:35 712 invalid packets
2022/04/26 21:01:35 6632070 unique signatures
2022/04/26 21:01:35 6613239 unique signers
2022/04/26 21:01:35 packets per signature: 4.63
(a lot of traffic was dropped before it reached this point, so this isn't all of it)"

nikhayes:
> Replying to [@metaplex](https://twitter.com/metaplex) The two edge cases relying on young devs using a boilerplate UI will end up hurting others. Also you are banking on the solana network holding up with retries on will end up charging the user. We literally hit 7 TPS on okay bears drop. This is not the way to accomplish this.

Hmm so either charge normal users a max of a few dollars or we risk the network going down like yesterday and datacenters issuing attack emails?  The deeper fixes aren't ready yet. The network going down will cost people much more $.

candy-machine/program/src/utils.rs
```diff
@@ -71,6 +73,18 @@ pub struct TokenTransferParams<'a: 'b, 'b> {
     pub token_program: AccountInfo<'a>,
 }
 
+pub fn punish_bots(err: ErrorCode, bot_account: &AccountInfo, payment_account: &AccountInfo, fee: u64) -> Result<(), ProgramError> {
+    msg!("Error: {}, Candy Machine Botting is taxed at {:?} lamports", err.to_string(), fee);
+    let bot_lamp = bot_account.lamports();
+    let mut final_fee = fee;
+    if bot_lamp < fee {
+        final_fee = bot_lamp;
+    }
```
mvines, line 82:
```suggestion
    let final_fee = fee.min(bot_account.lamports());
```

ohaddahan:
> > I'm not super familiar with CM, but if I'm a botter what's to stop me from adding a follow-up instruction to my mint transaction that checks:
> > ```
> > 1. that my account balance has been reduced the expected amount (the mint price)
> > 
> > 2. that i have in fact received an NFT
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and if the above conditions are not met then just throw an error and revert the whole transaction
> 
> The candy machine already checks to make sure only instructions from trusted/necessary programs are included in the mint transaction in order to prevent gaming the randomness. That would also prevent botters from chaining that additional check on.

https://github.com/metaplex-foundation/metaplex-program-library/blob/b1a749e30b26e0f0f650cbb0165283d1fa70f2d7/candy-machine/program/src/lib.rs#L418-L421

Not sure it will help, since balance checking is probably with `SYSTEM_PROGRAM` which is allowed if I read this correctly.

candy-machine/program/src/lib.rs
```diff
@@ -395,9 +430,6 @@ pub mod candy_machine {
             ],
             &[&authority_seeds],
         )?;
-        let instruction_sysvar_account_info = instruction_sysvar_account.to_account_info();
-
-        let instruction_sysvar = instruction_sysvar_account_info.data.borrow();
```
mvines, line 400: I can't comment this line in the PR:
```
 let associated_token =
            Pubkey::from_str("ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL").unwrap();
```
but we should fix this as well, as complete waste of compute units to perform a String to Pubkey conversion on chain.  Instead the conversion can be performed at build time with a:
```
const ASSOCIATED_TOKEN: Pubkey = solana_program::pubkey!("ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL");
```

mvines:
> Won't this have a minimal effect on duplicate transactions? If a bot sends 1k duplicate txs, won't it be penalized for just a single failed mint?

Yes that's true. Truly duplicate transactions are handled in the core validator software before the CM even gets a chance to run.

mvines:
> > > I'm not super familiar with CM, but if I'm a botter what's to stop me from adding a follow-up instruction to my mint transaction that checks:
> > > ```
> > > 1. that my account balance has been reduced the expected amount (the mint price)
> > > 
> > > 2. that i have in fact received an NFT
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and if the above conditions are not met then just throw an error and revert the whole transaction
> > 
> > 
> > The candy machine already checks to make sure only instructions from trusted/necessary programs are included in the mint transaction in order to prevent gaming the randomness. That would also prevent botters from chaining that additional check on.
> 
> https://github.com/metaplex-foundation/metaplex-program-library/blob/b1a749e30b26e0f0f650cbb0165283d1fa70f2d7/candy-machine/program/src/lib.rs#L418-L421
> 
> Not sure it will help, since balance checking is probably with `SYSTEM_PROGRAM` which is allowed if I read this correctly.

Yeah it will help.  Balance checking could only be performed by a custom program written by the botter, the system program has no facility that would be useful to reject the transaction if CM dispensed a punishment

ohaddahan:
What 

> > > > I'm not super familiar with CM, but if I'm a botter what's to stop me from adding a follow-up instruction to my mint transaction that checks:
> > > > ```
> > > > 1. that my account balance has been reduced the expected amount (the mint price)
> > > > 
> > > > 2. that i have in fact received an NFT
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > and if the above conditions are not met then just throw an error and revert the whole transaction
> > > 
> > > 
> > > The candy machine already checks to make sure only instructions from trusted/necessary programs are included in the mint transaction in order to prevent gaming the randomness. That would also prevent botters from chaining that additional check on.
> > 
> > 
> > https://github.com/metaplex-foundation/metaplex-program-library/blob/b1a749e30b26e0f0f650cbb0165283d1fa70f2d7/candy-machine/program/src/lib.rs#L418-L421
> > 
> > Not sure it will help, since balance checking is probably with `SYSTEM_PROGRAM` which is allowed if I read this correctly.
> 
> Yeah it will help. Balance checking could only be performed by a custom program written by the botter, the system program has no facility that would be useful to reject the transaction if CM dispensed a punishment

How about this attack vector, mint is 1SOL, wallet has 2SOL, adding a transfer of 1SOL to some other bot wallet at the end.
If it was punished, it will fail, it it minted, it will succeed.

Transfer SOL is via system program instructions right? So that would be allowed.

staccDOTsol:
Ohad it's a fantastic point and I wanted to bring this up, but hadn't reviewed in entirety

it needs to be commit n reveal :)

ohaddahan:
> Ohad it's a fantastic point and I wanted to bring this up, but hadn't reviewed in entirety
> 
> it needs to be commit n reveal :)

Stole that attack vector from https://github.com/Arrowana/cope-roulette-pro

staccDOTsol:
it's cooler as a larger idea than 'attack vector.' it has merit for instance in ensuring profitability if you're doing trades across a few hops

candy-machine/program/src/lib.rs
carllin: how long after a mint sells out do these bots generally persist their activity? If it's a long time after are they signing with old or newer blockhashes? We could exponentially dial up the penalty based on the distance between the blockhash the mint sold out and the blockhash in the transaction.

mvines:
> How about this attack vector, mint is 1SOL, wallet has 2SOL, adding a transfer of 1SOL to some other bot wallet at the end.
> If it was punished, it will fail, it it minted, it will succeed.

This only works if the mint fee is less than `BOT_FEE`, which typically is not the case right.  Otherwise:
1. the bot is punished with ◎0.01
2. the transfer of ◎1 in the following instruction succeeds
and when the transaction completes the wallet now has ◎0.99 in it

candy-machine/program/src/lib.rs
mvines: We've seen boting continue for many hours after the candy machine is empty.

I don't consider the "badly designed mint front-end" argument compelling. It's the project's responsibility to perform the necessary technical diligence to setup their mint correctly, and it's the user's responsibility to perform the necessary project diligence before participating in a mint. When either of these two things don't happen, you're playing with a 🔫 and the outcome you get is what you get.

t-nelson: It's be pretty easy to get around a blockhash age based escalation by always using a near expiration blockhash

t-nelson:
> > Won't this have a minimal effect on duplicate transactions? If a bot sends 1k duplicate txs, won't it be penalized for just a single failed mint?
> 
> Yes that's true. Truly duplicate transactions are handled in the core validator software before the CM even gets a chance to run.

need that core exponential fee run up on unused write locks :wink:

candy-machine/program/src/lib.rs
t-nelson: > Should we make this higher. Like ◎1?

If the mint price is available in all contexts, we should make the bot penalty a percentage of it so the penalty isn't diluted on higher priced drops

chen-robert:
Haven't looked too closely into this program, but a bypass idea: 

If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?

SwayStar123:
How about a variable fee? Like the lesser offense of sending a tx after mint is over gets a 0.001 fee and using a CPI gets a 0.1 fee

mvines:
> Haven't looked too closely into this program, but a bypass idea:
> 
> If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?

No they couldn't because that would require a custom program in the transaction that that is already not allowed

mvines:
> How about a variable fee? Like the lesser offense of sending a tx after mint is over gets a 0.001 fee and using a CPI gets a 0.1 fee

I don't see why this would be needed. A normal human, using a non-broken mint interface, should never run into any of these conditions. So why give spammers any relief. They need to be punished relentlessly.

ohaddahan:
> Haven't looked too closely into this program, but a bypass idea:
> 
> If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?

You don't know the `mint` address beforehand , so you can't create the instructions to transfer it before the mint.

> > Haven't looked too closely into this program, but a bypass idea:
> > If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?
> 
> No they couldn't because that would require a custom program in the transaction that that is already not allowed

And as @mvines mentioned , you can't use a custom program to detect it.

stegaBOB:
> > Haven't looked too closely into this program, but a bypass idea:
> > 
> > If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?
> 
> No they couldn't because that would require a custom program in the transaction that that is already not allowed

I think a better way to phrase this is that when you mint from the candy machine you aren't getting a new nft. In fact, you're just passing in an empty metadata account that the candy machine fills in. So if you were to do this you'd basically have a successful token transfer of a blank nft while also paying the bot fee.

ohaddahan:
> > How about a variable fee? Like the lesser offense of sending a tx after mint is over gets a 0.001 fee and using a CPI gets a 0.1 fee
> 
> I don't see why this would be needed. A normal human, using a non-broken mint interface, should never run into any of these conditions. So why give spammers any relief. They need to be punished relentlessly.

I slightly disagree, in blitz mints that run end in minutes, many real users simply see outdated UI info.
So I do see a large chance of real users being punished.

stegaBOB:
> > Haven't looked too closely into this program, but a bypass idea:
> > 
> > If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?
> 
> You don't know the `mint` address beforehand , so you can't create the instructions to transfer it before the mint.
> 
> > > Haven't looked too closely into this program, but a bypass idea:
> > > If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?
> > 
> > No they couldn't because that would require a custom program in the transaction that that is already not allowed
> 
> And as @mvines mentioned , you can't use a custom program to detect it.
> 
> 
> 
> 

You actually would know the mint address since you pass it in before hand. It just wouldn't really matter since you'd have a blank metadata account to go with it if the candy machine fails.

ohaddahan:
> > > Haven't looked too closely into this program, but a bypass idea:
> > > If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?
> > 
> > 
> > You don't know the `mint` address beforehand , so you can't create the instructions to transfer it before the mint.
> > > > Haven't looked too closely into this program, but a bypass idea:
> > > > If NFTs are represented as an `spl-token` (not sure if this is true), could a user attempt a token transfer to check to see if they successfully got the NFT?
> > > 
> > > 
> > > No they couldn't because that would require a custom program in the transaction that that is already not allowed
> > 
> > 
> > And as @mvines mentioned , you can't use a custom program to detect it.
> 
> You actually would know the mint address since you pass it in before hand. It just wouldn't really matter since you'd have a blank metadata account to go with it if the candy machine fails.

Ah OK, I stand corrected.

mvines:
> I slightly disagree, in blitz mints that run end in minutes, many real users simply see outdated UI info. So I do see a large chance of real users being punished.

maybe the idea of blitz mints is just bad?  the fomo that generates seems exploitative.

But regardless of the answer to/position on that point, this PR is really only a v0 where the goal here is simply to try to help avoid the network from getting fucked every single day by greedy nft spammers that don't care about the Solana ecosystem overall. I believe the solution will need to be iterated on multiple times before we arrive at something that truly removes the broken incentives that enabled us to get to this situation.  Or more colloquially, let's not let perfect be the enemy of good.

ohaddahan:
> > I slightly disagree, in blitz mints that run end in minutes, many real users simply see outdated UI info. So I do see a large chance of real users being punished.
> 
> maybe the idea of blitz mints is just bad? the fomo that generates seems exploitative.
> 
> But regardless of the answer to/position on that point, this PR is really only a v0 where the goal here is simply to try to help avoid the network from getting fucked every single day by greedy nft spammers that don't care about the Solana ecosystem overall. I believe the solution will need to be iterated on multiple times before we arrive at something that truly removes the broken incentives that enabled us to get to this situation. Or more colloquially, let's not let perfect be the enemy of good.

I agree, but we get better by providing constructive criticism.
That's why I try and find ways to hack it, so we can plug the holes before it's released.

`perfect be the enemy of good` one of the most important motives an engineer need to consider, there are many approximations used in engineering that keep our planes flying and other critical infrastructure , so yes, please more of this!

jeffowy:
> > How about a variable fee? Like the lesser offense of sending a tx after mint is over gets a 0.001 fee and using a CPI gets a 0.1 fee
> 
> I don't see why this would be needed. A normal human, using a non-broken mint interface, should never run into any of these conditions. So why give spammers any relief. They need to be punished relentlessly.

Exactly this.   I personally think 0.1 Sol is the maximum amount.  Not 1.  Even 0.001 would be effective.  

This is an excellent PR and I am actually shocked with the resistance to it.  Huge thanks to everyone who put it together, and we will be implementing it on Safecoin regardless of the outcome here.

Sure there are edge cases.  There are always edge cases.   People lose money on Solana and every other blockchain, every single day when they make mistakes.   When they learn from those mistakes they stop losing money.  That's how it works.

So imho we either shut down all of the blockchains that have edge cases or we accept the fact that users have to have some level of accountability for their actions, and learn from them.


The other arguement I'm seeing against this is it's not solving the "root issue".

Yes, yes it is solving the root issue.     Resources have a cost and abuse of those resources needs to be accounted for in turn.  This has to jive in any successful economic model or you are rekt.    Idealists that insist everything should run smooth at 100,000 tps, should be focused on that goal in devnet.  Not mainnet.   Mainnet needs to work, it needs to be practical, all of the time.   Even if you achieve this, someone will come up with a way to clog the network up with 1 million tps if you don't have a fundamentally sound reward model.


The only thing I will give the detractors is that it would be a lot better to see a version of this in the solana repo instead of the metaplex repo.  Sure ok.    But that can be done in parallel.   There is absolutely no reason why this should prevent a go-ahead for this PR.

nikhayes:
> > > How about a variable fee? Like the lesser offense of sending a tx after mint is over gets a 0.001 fee and using a CPI gets a 0.1 fee
> > 
> > 
> > I don't see why this would be needed. A normal human, using a non-broken mint interface, should never run into any of these conditions. So why give spammers any relief. They need to be punished relentlessly.
> 
> Exactly this. I personally think 0.1 Sol is the maximum amount. Not 1. Even 0.001 would be effective.
> 
> This is an excellent PR and I am actually shocked with the resistance to it. Huge thanks to everyone who put it together, and we will be implementing it on Safecoin regardless of the outcome here.
> 
> Sure there are edge cases. There are always edge cases. People lose money on Solana and every other blockchain, every single day when they make mistakes. When they learn from those mistakes they stop losing money. That's how it works.
> 
> So imho we either shut down all of the blockchains that have edge cases or we accept the fact that users have to have some level of accountability for their actions, and learn from them.
> 
> The other arguement I'm seeing against this is it's not solving the "root issue".
> 
> Yes, yes it is solving the root issue. Resources have a cost and abuse of those resources needs to be accounted for in turn. This has to jive in any successful economic model or you are rekt. Idealists that insist everything should run smooth at 100,000 tps, should be focused on that goal in devnet. Not mainnet. Mainnet needs to work, it needs to be practical, all of the time. Even if you achieve this, someone will come up with a way to clog the network up with 1 million tps if you don't have a fundamentally sound reward model.
> 
> The only thing I will give the detractors is that it would be a lot better to see a version of this in the solana repo instead of the metaplex repo. Sure ok. But that can be done in parallel. There is absolutely no reason why that should prevent a go-ahead for this PR.

Pretty sure the "resistance" is just from people who benefit from the bots in some way and this is one of the few arguments they have. Looks like the Metaplex poll has now been botted or something too.

candy-machine/program/src/lib.rs
austbot: @t-nelson We considered this and generally im in favor of this. I actually would love to have more accuracy and punish them exponentially on each attempt

candy-machine/program/src/lib.rs
austbot: will fix.

austbot: ah I thought pubkey! macro was in the SDK crate which we don't usually pull in, I tried to find it the other day.

levicook:
2 is going to affect users — Keeping a UI up-to-date is quite difficult (not speculation, been there). To mitigate damage to real users here, I like the idea of putting in a curve. Slightly late, smol fee, more late, bigger and bigger fees is pretty reasonable imo

levicook:
Also caution against rhetoric that only botters are going to resist—it discourages discourse and knowledge sharing.

Arrowana:
> ### Specific Fee Charging Cases
> 1. Trying to mint when the candy machine is not live before or after
> 2. Trying to mint when there are no items left in the candy machine
> 3. Calling Candy Machine via CPI when not using gumdrop
> 4. Crafting a transaction where mint or set collection is not the last ix
> 5. Using the wrong collection id than the configured one for the candy machine
> 6. Setting Collection IX with a mismatched mint than what was just minted
> 7. Signer Payer mismatch with collection set and mint ix
> 8. Suspicious Transactions where disallowed programs are used
> 9. Trying to mint on an AllowList candy machine with no allow list token

I believe 3, 4, 5, 6, 7, 8, 9 to be possibly totally pointless since the boter goal is to eventually mint the NFT, not spam with buggy code so i expect those cases to be exotic.

2. is pretty serious, in a hyped mint many could end up paying as a bot, while not being bots.

1. and 2. can be trivially worked around with the following setup.

2 streams of spam in 2 distinct transactions, using 2 different wallets:
- fund the wallet #2 from wallet #1 with SOL if the condition 1 is true (by reading the CM account), if condition 2 is true, drain wallet #2 into wallet #1
- mint a NFT from the CM with wallet #2 which starts with 0 SOL. Wallet #2 has no SOL at the start of the stunt.
You end up with pretty much the same system as before, the proposed change doesn't work.

0xChief:
#1 should be be changed since it could hit more real users than sufficiently advanced bots.

It's an easy mistake on the CM practitioner to accidentally use the wrong unix time, have clock skew from somewhere (possibly relying on a DateTime instead of unix), or any other number of genuine possibilities. 

Can't you read the CM mint start/end date without relying on creating a mint transaction anyways? Any sufficiently advanced botter should be able to avoid this and it only hurts real users as explained above.

candy-machine/program/src/lib.rs
carllin: > It's be pretty easy to get around a blockhash age based escalation by always using a near expiration blockhash

Yeah, but then the attack will only last until blockhash expiration

t-nelson:
> 1 should be be changed since it could hit more real users than sufficiently advanced bots.

if all this succeeds in is killing off the naive bots, that's still a huge win.  I can't count the number of times I've seen 1Mpps of failed TX streaming for hours after a mint closes.  Same for bots hosing down the network despite their fee wallet having been emptied

> It's an easy mistake on the CM practitioner to accidentally use the wrong unix time, have clock skew from somewhere (possibly relying on a DateTime instead of unix), or any other number of genuine possibilities.

UI's should be streaming block time, not using the local clock to open the mint.  That's the time the contract will be using after all

> Can't you read the CM mint start/end date without relying on creating a mint transaction anyways? Any sufficiently advanced botter should be able to avoid this and it only hurts real users as explained above.

that's what the transaction composition checks are for

t-nelson:
> I believe 3, 4, 5, 6, 7, 8, 9 to be possibly totally pointless since the boter goal is to eventually mint the NFT, not spam with buggy code so i expect those cases to be exotic.
> 
> 2 is pretty serious, in a hyped mint many could end up paying as a bot, while not being bots.

a low-latency "mint open" signal can be built from the output of a geyser plugin

> 1 and 2 can be trivially worked around with the following setup.
> 
> 2 streams of spam with 2 distinct transactions, using 2 different wallets:
> 
>     * fund the wallet # 2 from wallet # 1 with SOL if the condition 1 is true (by reading the CM account), if condition 2 is true, drain wallet # 2 into wallet # 1
> 
>     * mint a NFT from the CM with wallet # 2 which starts with 0 SOL. Wallet # 2 has no SOL at the start of the stunt.
> 
> 
> Wallet # 2 doesn't have any lamport when the rug conditions are met so the transaction is not added. You end up with pretty much the spammable system as before, the proposed change doesn't work.

this description isn't particularly clear to me (overloaded 1's and 2's), so i may have missed something, but i believe between the transaction composition checks and runtime's inability to guarantee transaction ordering, this would be unreliable if not infeasible

roederw:
In light of recent events, and the general community sentiment that this is a net positive addition to the contract, we are moving forward with this change. Pray for us.

roederw (approved):

staccDOTsol:
Congrats

0xChief:
> UI's should be streaming block time, not using the local clock to open the mint. That's the time the contract will be using after all

Common open source candy machine UI implementations do not do this and is why I believe 1 would have a heavy impact on real users. See [https://github.com/metaplex-foundation/metaplex/blob/2782e2710ac099238115f31682999fe553cc01be/js/packages/candy-machine-ui/src/Home.tsx#L625-L642](url)

aziz-light:
what was the point of the vote if you were going to push it anyway you fuckin coward @roederw you just killed your service no one is going to use CM anymore

aziz-light:
> In light of recent events, and the general community sentiment that this is a net positive addition to the contract, we are moving forward with this change. Pray for us.
> 
> We will continue to monitor this change and intend to iterate if something doesn't go according to plan.
Screen Shot 2022-04-30 at 3 28 24 PM
this community sentiment? moron

roederw:
The vote was unanimously positive until it got hit by bots. Calm down.

aziz-light:
can't wait to try to manually mint after something sold out and get charged for it

staccDOTsol:
each standard interface protects against this from happening :)

jkbpvsc:
> > In light of recent events, and the general community sentiment that this is a net positive addition to the contract, we are moving forward with this change. Pray for us.
> > We will continue to monitor this change and intend to iterate if something doesn't go according to plan.
> 
> Screen Shot 2022-04-30 at 3 28 24 PM
> 
> this community sentiment? moron

Vote against bots got botted, wdy expect?

Given that RPC node providers are discussing txs with blocking CM program id, I think this solution is very reasonable.

stranzhay:
> what was the point of the vote if you were going to push it anyway you fuckin coward @roederw you just killed your service no one is going to use CM anymore

@aziz-light happy to review your PR on how to get the network back up, let us know when it's ready

candy-machine/program/src/lib.rs
17606903541: 发生了什么

roederw:
> > > In light of recent events, and the general community sentiment that this is a net positive addition to the contract, we are moving forward with this change. Pray for us.
> > > We will continue to monitor this change and intend to iterate if something doesn't go according to plan.
> > 
> > 
> > Screen Shot 2022-04-30 at 3 28 24 PM
> > this community sentiment? moron
> 
> Vote against bots got botted, wdy expect?
> 
> Given that RPC node providers are discussing txs with blocking CM program id, I think this solution is very reasonable.

This is actually why we've expedited the change. We'd like to help keep the network up post restart. For the record, we're testing on `devnet` now to confirm this works as intended and we don't accidentally hurt the non-bot community.

jeffowy:
>4700 votes for a tweet that was liked 44 times, lol.

Sure, no bots there.

The very fact that the vote was botted is highly suggestive that this is a needed solution.

ohaddahan:
So how about this @austbot for exponential fees https://github.com/metaplex-foundation/metaplex-program-library/pull/424 (just a draft)

candy-machine/program/src/lib.rs
Jac0xb: This is not a solution. All it takes is the frontend dev not knowing how dates work and allowing users to send mint instructions.

mvines: whip the frontend dev because they didn't make good html then

Jac0xb:
I could write a solana program in a day that checks if the candy machine account is within the anti-bot parameters and fails if its going to get punished. That gets placed in run of mint_nft instruction and voila, botters back in action.

On the flipside, any user who mints during the desync delay of CM state where its minted out is going to be punished. This could affect way more users than bots because botters can pull CM state in a much more controlled manner than UI mint users.

mvines:
> I could write a solana program in a day that checks if the candy machine account is within the anti-bot parameters and fails if its going to get punished. That gets placed in front of mint_nft instructions and voila, botters safe.

nope because this PR will punish your bot for including that program in the transaction.  but if you find a way around this, please raise a new issue so the metaplex team can patch that hole as well

mvines:
> Also I've seen very jank datetime code for mint websites, these websites could easy accidently allow users to mint before/after mint dates.

People should yell at those projects for putting their users at risk then, and the users themselves should not risk minting with inferior projects.

Jac0xb:
> > Also I've seen very jank datetime code for mint websites, these websites could easy accidently allow users to mint before/after mint dates.
> 
> People should yell at those projects for putting their users at risk then, and the users themselves should not risk minting with inferior projects.

Have you ever coded a frontend and tried to sync solana account state with UI? Most CM UI's don't pull more than once and 99% don't pull more than every second. That is a shit ton of users getting charged if the mint is popular.

mvines:
Probably need a reasonable grace period on that particular punishment. This PR should not be considered the final solution, open a new issue and/or discuss your suggestions on the metaplex discord. It’ll take multiple iterations of the program and client code.

Jac0xb:
> Probably need a reasonable grace period on that particular punishment. This PR should not be considered the final solution, open a new issue and/or discuss your suggestions on the metaplex discord. It’ll take multiple iterations of the program and client code.

Why not force gatekeeper and make it so authority can use CLI to mint with gatekeeper?

PR #418, opened 2022-04-30
Enforce Collections w/ CM and clean up code

stegaBOB:
Bumps anchor version, fixes Clippy issues, splits up huge file into multiple byte sized chunks.

Also uses the currently unused UUID as a sentinel to check if a collection has been set.


candy-machine/program/src/utils.rs
```diff
@@ -1,55 +1,54 @@
...
     let account: T = T::unpack_unchecked(&account_info.data.borrow())?;
     if !account.is_initialized() {
-        Err(ErrorCode::Uninitialized.into())
+        Err(CandyError::Uninitialized.into())
     } else {
         Ok(account)
     }
 }
 
+pub fn cmp_pubkeys(a: &Pubkey, b: &Pubkey) -> bool {
```
roederw, line 24: We need a centralized crate for these helpers. This is so prone to error since it's dup'ed across every workspace.

stegaBOB: Agreed. We need a utils lib or something.

candy-machine/program/src/processor/update.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    Ok(())
+}
+
+pub fn handle_update_candy_machine(
+    ctx: Context<UpdateCandyMachine>,
+    data: CandyMachineData,
+) -> Result<()> {
+    let candy_machine = &mut ctx.accounts.candy_machine;
```
austbot, line 34: enforce they cannot change feature flags here

candy-machine/program/src/utils.rs
```diff
@@ -192,5 +171,100 @@ pub fn spl_token_burn(params: TokenBurnParams<'_, '_>) -> ProgramResult {
         &[source, mint, authority, token_program],
         seeds.as_slice(),
     );
-    result.map_err(|_| ErrorCode::TokenBurnFailed.into())
+    result.map_err(|_| CandyError::TokenBurnFailed.into())
+}
+
+pub fn is_feature_active(uuid: &str, feature_index: usize) -> bool {
+    uuid.as_bytes()[feature_index] == b"1"[0]
```
austbot, line 178: interesting

candy-machine/program/src/utils.rs
```diff
@@ -192,5 +171,100 @@ pub fn spl_token_burn(params: TokenBurnParams<'_, '_>) -> ProgramResult {
...
+}
+
+// string is 6 bytes long, can be any valid utf8 char coming in.
+// feature_index is between 0 and 5, inclusive
+// unsafe is fine because we know for a fact that the array will only
+// contain valid UTF8 bytes (we set it as "1" or "0")
+pub fn set_feature_flag(uuid: &mut String, feature_index: usize) {
+    let mut bytes: [u8; 6] = [b"0"[0]; 6];
+    uuid.bytes().enumerate().for_each(|(i, byte)| {
+        if i == feature_index || byte == "1".as_bytes()[0] {
```
austbot, line 188: "1".as_bytes()[0] ==  b"1"[0] ?

austbot (changes requested):
A few nits and one validation. but other wise great work

candy-machine/program/src/utils.rs
stegaBOB: Yeah I should probably change this to b"1"[0].

stegaBOB:
There's basically no IDL change. What we need to announce more is the enforcement of the collections

austbot (dismissed):

austbot (approved):

candy-machine/program/src/processor/mint.rs
```diff
@@ -0,0 +1,762 @@
...
+    );
+    if arr[CONFIG_ARRAY_START + 4 + index_to_use * (CONFIG_LINE_SIZE)] == 1 {
+        return err!(CandyError::CannotFindUsableConfigLine);
+    }
+
+    let data_array = &mut arr[CONFIG_ARRAY_START + 4 + index_to_use * (CONFIG_LINE_SIZE)
+        ..CONFIG_ARRAY_START + 4 + (index_to_use + 1) * (CONFIG_LINE_SIZE)];
+
+    let mut name_vec = vec![];
+    let mut uri_vec = vec![];
```
danenbm, line 733: Not sure how much it matters here but might be better to have one allocation and just do:
```
let mut name_vec = Vec::with_capacity(MAX_NAME_LENGTH);
let mut uri_vec = Vec::with_capacity(MAX_URI_LENGTH);
```

danenbm (dismissed):

candy-machine/program/src/processor/add_config_lines.rs
```diff
@@ -0,0 +1,111 @@
...
+            .checked_add(i)
+            .ok_or(CandyError::NumericalOverflowError)?;
+        let my_position_in_vec = bit_mask_vec_start
+            + position
+                .checked_div(8)
+                .ok_or(CandyError::NumericalOverflowError)?;
+        let position_from_right = 7 - position
+            .checked_rem(8)
+            .ok_or(CandyError::NumericalOverflowError)?;
+        let mask = u8::pow(2, position_from_right as u32);
```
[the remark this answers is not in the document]
stegaBOB: Shouldn't be possible to overflow here, but I think its still safer to do checked? Do you think we should add a redundant check just in case?

candy-machine/program/src/processor/mint.rs
danenbm: I don't know about on the blockchain, but on my Macbook an empty Vec starts out with capacity of 8, and then reallocates to 16, 32, etc. as more items pushed, basically a standard dynamic array.

candy-machine/program/src/processor/add_config_lines.rs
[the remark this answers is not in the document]
danenbm: I see the `checked_rem()` now.  Before I wasn't looking closely at the previous math.   I think you do NOT need to do a an extra check.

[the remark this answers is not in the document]
danenbm: I see the checked_rem() now. Before I wasn't looking closely at the previous math. I think you do NOT need to do a an extra check.

danenbm (approved):
Looks good to me, and all the updates since last CR look good to me.  But you may want a second set of eyes, especially on things like updates to solita and generated commands, since I'm still pretty new to this code base overall.

samuelvanderwaal (approved):
I reviewed the two latest commits and they looked fine

austbot (approved):

PR #422, opened 2022-04-30
Add comments to candy machine program

vpontis:
I am making an alternative NFT standard so I went through the Candy Machine code to understand how it works. I added comments while working on this to make it easier for people to understand. 

There are still some hairy parts of the code that don't have comments or tests.


vpontis:
Hmmm. I guess I'll close this then...

PR #424, opened 2022-05-01
[DRAFT] Mints per wallet + exp punishment

ohaddahan:
Punishment curve = `BaseFee + (BaseFee - MintPrice) * exp( - (failed - allowed)/Slope)` .

Which looks like this in case of `BaseFee = 0.1` , `MintPrice = 2` and `Slope = 10`. 

This will also enforce `mints per wallet`, but will require API change.

image

(https://www.desmos.com/calculator/3fisjexbvp)

### DIFF
--- a/candy-machine/program/src/lib.rs
+++ b/candy-machine/program/src/lib.rs
@@ -81,15 +81,16 @@ pub mod candy_machine {
 
         /// Restrict Who can call Candy Machine via CPI
         if current_ix.program_id != candy_machine::id() && current_ix.program_id != GUMDROP_ID {
+            mint_memory.failed += 1;
             punish_bots(
                 ErrorCode::SuspiciousTransaction,
-                mint_memory.to_account_info(),
                 payer.to_account_info(),
                 ctx.accounts.candy_machine.to_account_info(),
                 ctx.accounts.system_program.to_account_info(),
                 BOT_FEE,
                 candy_machine.data.price,
-                candy_machine.data.allowed_per_wallet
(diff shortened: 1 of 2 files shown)


devpouya:
fees should punish repeated violations to discourage spamming. How about making a pda for each bot account and keeping a counter in there to count the number of violations, then raise the fee to this power to make it explode as fast as possible (for repeated violations).

ohaddahan:
> fees should punish repeated violations to discourage spamming. How about making a pda for each bot account and keeping a counter in there to count the number of violations, then raise the fee to this power to make it explode as fast as possible (for repeated violations). 
> 
> 
> 
> This however wouldn't handle Sybil attacks too well I guess...
> 
> 
> 
> 

This is what I have here, extra PDA that keeps track and increases the penalty.

ohaddahan:
> @ohaddahan this is cool, its something we have considered too. @jarry-xiao and I have been looking at this. Here is what we have come up with:
> 
> The first fee must be substantial as tacking attempts on subkeys is not Sybil resistant
> 
> Above is true however the management of thousands of private keys is not impossible but operationally difficult, especially with the transfer overhead. 

Sybil attack is 1:1 correlation to the minting funds as in 1 SOL mint and you want to mint 100.
At best you'll use 100 wallets (if you want to mint and not just DDOS the network , that can be done with 0 SOL).

Theoretically you can use more but moving funds after each trial will be too slow so you'll just find wallets ahead of time and spam away.

Hence this attack vector ain't that large.
Most txns will be from repeated wallets .

PR #436, opened 2022-05-04
stranzhay/token metadata uses bug

### LINKED ISSUE #294: [Bug]: Token Metadata: process_update_metadata_accounts_v2() does not update the Uses

token-metadata

### Problem Summary
The `process_update_metadata_accounts_v2()` instruction in the Metadata Program silently ignores the `Uses` section of the `DataV2` that is optionally passed in. As a result, it seems there is no way to update the Uses value of the Metadata after mint, and whatever is initially set in CreateMetadataAccountV2 is locked in forever. 

### Expected Behavior
The `process_update_metadata_accounts_v2()` instruction will allow updating the Uses data of a Metadata account.

### Additional Context / Root Cause

Relevant snippet from [lines 266-283 of processor.rs](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/processor.rs#L266):
```rust
if let Some(data) = optional_data {
        if metadata.is_mutable {
            let compatible_data = data.to_v1();
            assert_data_valid(
                &compatible_data,
                update_authority_info.key,
                &metadata,
                false,
                update_authority_info.is_signer,
                true,
            )?;
            assert_collection_update_is_valid(false, &metadata.collection, &data.collection)?;
            metadata.collection = data.collection;
        } else {
            return Err(MetadataError::DataIsImmutable.into());
        }
    }
```

We can see that the v1 data is extracted, asserted-as-valid, and set. Then the collection is asserted-as-valid and set. But the Uses data is completely ignored. Note there is an existing [assert_valid_use() function in assertions/uses.rs](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/assertions/uses.rs#L9), it just seems the incoming Uses field was forgotten about.

### Relevant log output

_No response_

Medium (should be fixed soon)

stranzhay:
closes #294

### DIFF
--- a/candy-machine/program/Cargo.lock
+++ b/candy-machine/program/Cargo.lock
@@ -24,9 +24,9 @@ dependencies = [
 
 [[package]]
 name = "anchor-attribute-access-control"
-version = "0.21.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8573731461c6e39febd16026fe95cbc99955585b08acddc0baaeefb803da191b"
+checksum = "a9b75d05b6b4ac9d95bb6e3b786b27d3a708c4c5a87c92ffaa25bbe9ae4c5d91"
 dependencies = [
  "anchor-syn",
  "anyhow",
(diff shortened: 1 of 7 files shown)


roederw:
What happened with this?

PR #440, opened 2022-05-05
feat(token-metadata): enhancing token standard setting

### LINKED ISSUE #433: [Feature]: Allow Setting TokenStandard for Tokens Without Them

token-metadata

### Feature

Currently, when a new metadata account is created there is [logic that checks the token type](https://github.com/metaplex-foundation/metaplex-program-library/blob/e4dec2c51615ac703cbe28398646527e385346b5/token-metadata/program/src/utils.rs#L921) it is and automatically sets the TokenStandard to `Fungible`, `FungibleAsset` or `NonFungibleEdition`. The `create_master_edition`[ handler](https://github.com/metaplex-foundation/metaplex-program-library/blob/e4dec2c51615ac703cbe28398646527e385346b5/token-metadata/program/src/processor.rs#L501) is used to set it to `NonFungible` when it is called but there's no way tokens existing prior to these `token-metadata` changes to set the TokenStandard.

We need to add a new handler or handlers to do this.

### Standard Change?

No

Requirements:

* We want to set the TokenStandard based on the actual token type not an arbitrary value passed in by the user so we need to check things like if it has a master edition, edition, what the token decimals are, what the supply is, etc.

* Each new handler should have Rust tests that check expected success and failure behaviors

* If it makes sense we can consider extracting the token check logic into a separate function and using it for `process_create_metadata_accounts_logic` as well

Token Standard Types:

NonFungible: This is our standard NFT, a token account with a supply of 1, decimals 0, and a master edition PDA

NonFungibleEdition: This is our standard Edition NFT, a token account with a supply of 1, decimals 0 and an edition PDA

FungibleAsset: This is a token mint that can have a supply greater than 1 but for which decimals are 0 and which the JSON supports richer metadata

Fungible: This is a standard fungible token with basic metadata, supply can be greater than 1, decimals can be greater than 0 and basic JSON metadata.

_No response_

### Other context

_No response_

stranzhay:
closes #433


token-metadata/program/src/utils.rs
```diff
@@ -940,6 +933,73 @@ pub fn process_create_metadata_accounts_logic(
     Ok(())
 }
 
+pub fn check_token_standard(
+    metadata: Metadata,
+    edition_account_info: Option<&AccountInfo>,
+    program_id: &Pubkey,
+) -> Option<TokenStandard> {
+    if let Some(val) = edition_account_info {
+        let mint_decimals = get_mint_decimals(val).ok()?;
```
samuelvanderwaal, line 942: `get_mint_decimals` expects the `AccountInfo` of an actual [mint account](https://docs.rs/spl-token/latest/spl_token/state/struct.Mint.html) so you can't pass in the master edition or edition account here. 

I can go into more details of what this function is doing if you want the nitty gritty details but it's basically referencing directly into the Mint account layout to get the decimals value.

token-metadata/program/src/utils.rs
```diff
@@ -940,6 +933,73 @@ pub fn process_create_metadata_accounts_logic(
...
+
+pub fn assert_edition(
+    metadata: Metadata,
+    edition_account_info: &AccountInfo,
+    mint_decimals: u8,
+    program_id: &Pubkey,
+) -> Result<(), ProgramError> {
+    assert_edition_valid(program_id, &metadata.mint, edition_account_info)?;
+
+    let mint_supply = get_mint_supply(edition_account_info)?;
```
samuelvanderwaal, line 989: Same issue here as `get_mint_decimals`: it's expecting an actual mint account.

samuelvanderwaal (commented):
This is a good run at a first draft but needs some discussion and changes.

I like they way you are checking the token standard type and returning an option, it's a good DRY approach.

One piece of context for this PR, is we're trying to include it in v1.3 which is a minor change but which shouldn't have breaking changes for the current on-chain API (no additional accounts or args that will cause current implementations to fail). Adding a new handler or adding additional logic to an existing handler that uses the accounts and arguments it already has is probably the way to go here.

I don't think we need to add the check to any deprecated processor handlers (`create_metadata_account`). We're deprecating them for a reason so don't need to keep them updated. Also, the way you've done so is a breaking change for anyone who is still using them, which we want to avoid. (You've added a non-optional account to the handler.)

You've also added a non-optional account to `create_metadata_account_v2` which is a breaking change.

token-metadata/program/src/instruction.rs
```diff
@@ -400,7 +420,25 @@ pub enum MetadataInstruction {
     /// Remove Creator Verificaton.
     #[account(0, writable, name="metadata", desc="Metadata (pda of ['metadata', program id, mint id])")]
     #[account(1, signer, name="creator", desc="Creator")]
-    RemoveCreatorVerification
+    RemoveCreatorVerification,
+
+       /// Update a Metadata with is_mutable as a parameter
+       #[account(0, writable, name="metadata", desc="Metadata account")]
```
austbot, line 426: weird indent

token-metadata/program/src/instruction.rs
```diff
@@ -400,7 +420,25 @@ pub enum MetadataInstruction {
...
+
+         /// Create Metadata object.
+      #[account(0, writable, name="metadata", desc="Metadata key (pda of ['metadata', program id, mint id])")]
+      #[account(1, name="mint", desc="Mint of token asset")]
+      #[account(2, signer, name="mint_authority", desc="Mint authority")]
+      #[account(3, signer, writable, name="payer", desc="payer")]
+      #[account(4, name="update_authority", desc="update authority info")]
+      #[account(5, name="system_program", desc="System program")]
+      #[account(6, name="rent", desc="Rent info")]
+      #[account(7, optional, writable, name="edition",  desc="Unallocated edition V2 account with address as pda of ['metadata', program id, mint, 'edition']")]
```
austbot, line 440: So whats the prupose here just make the master edition in one go?

token-metadata/program/src/utils.rs
```diff
@@ -925,6 +1074,114 @@ pub fn process_create_metadata_accounts_logic(
...
+    ];
+    let (_, edition_bump_seed) = Pubkey::find_program_address(edition_seeds, program_id);
+    metadata.edition_nonce = Some(edition_bump_seed);
+    metadata.serialize(&mut *metadata_account_info.data.borrow_mut())?;
+
+    Ok(())
+}
+
+/// Create a new account instruction
+pub fn process_create_metadata_accounts_v3_logic(
```
austbot, line 1094: I'm not sure what you are up to with adding edition

austbot (changes requested):
im too dumb to understand what you are up too it seems genius but I nEeD help

stranzhay:
@austbot removed the unneeded code. the tests:     update_metadata_account_v3::success
    update_metadata_account_v3::success_update_metadata_when_collection_is_verified, are not passing saying `NotEnoughAccountKeys` but everything i am logging out shows the correct number of keys. not sure if you see something obvious that i am missing

stranzhay:
@samuelvanderwaal cherry picked some of my changes for v1.3 from here into that pr.  i will go ahead and close this.

PR #448, opened 2022-05-11
Add spl token gating feature to fixed-price-sale program

VadimGrozinok:
## What

This PR adds SPL token gating feature to the fixed-price-sale program.

## Why

Now fixed-price-sale supports only collection gating. By adding SPL token gating feature we extend number of use cases for the program. Also this type of gating is common in Metaplex programs so community already know how it works. Have to mention that collection gating also remains so it's up to users to choose which type of gating to use.

## How

PR contains small changes in Buy instruction. It's backward compatible.


austbot (dismissed):
Looks good now

Issue #428, opened 2022-05-02
[Bug]: Can't deserialize metadata using latest lib


ilmoi:
### Which package is this bug report for?

token-metadata

### Issue description

Trying to deserialize a bunch of PDAs for Boryoku Dragonz that I got from gPA:

creator = DRGNjvBvnXNiQz9dTppGk1tAsVxtJsvhEmojEfBU3ezf

```js
const rawMetadatas = await conn.getProgramAccounts(new PublicKey(METADATA_PROGRAM_ID), {
  filters: [
    ...baseFilters,
    {
      memcmp: {
        offset: programs.metadata.computeCreatorOffset(0),
        bytes: creator.toBase58(),
      },
    },
  ],
});

```

using the latest metdata js lib (@metaplex-foundation/mpl-token-metadata):
```js
const deserializeMetadata = (rawMetadata: any): Metadata | undefined => {
  try {
    return Metadata.fromAccountInfo(rawMetadata.account)[0];
    // @ts-ignore
  } catch (e: any) {
    console.log("Failed to deserialize on-chain metadata:", e);
  }
};

```

For some of them works fine (about 100 out of 1100), but for the rest it fails with the following error:
```
Failed to deserialize on-chain metadata: AssertionError [ERR_ASSERTION]: Expected ��� ����F~��I��*�Țg�?D4�/�Ą͝!�o�mB!̺e0�Yy��3���S�d�c� Boryoku Dragonz #618
BORYOKU�https://arweave.net/3y40kWLZHh1LV6sKCMUV0UKG2MK-HQ45EN5lRL1kuVk���� ����F~��I��*�Țg�?D4�/�Ąd��~z�.R�c���|����~��^�7NĄd� to hold a COption
    at Object.toFixedFromData (/Users/ilmoi/Dropbox/crypto_bc/tensor/tensor-infra/etl/sol/src/node_modules/@metaplex-foundation/beet/src/beets/composites.ts:141:15)
    at fixBeetFromData (/Users/ilmoi/Dropbox/crypto_bc/tensor/tensor-infra/etl/sol/src/node_modules/@metaplex-foundation/beet/src/beet.fixable.ts:22:17)
    at FixableBeetStruct.toFixedFromData (/Users/ilmoi/Dropbox/crypto_bc/tensor/tensor-infra/etl/sol/src/node_modules/@metaplex-foundation/beet/src/struct.fixable.ts:85:40)
[...]

PR #455, opened 2022-05-12
chore: sync update SDK with contract changes on master

thlorenz:
This also fixes the issue encountered when parsing token metadata that is corrupted.
A custom deserializer was added and used from the generated SDK via a manual edit.

FIXES: #428

~## Remaining work~

## Supporting Solita changes

Solita added an option to override `de/serializers` per account when generating code and we use that feature here in order to properly use our custom deserializer without manual edits.

Solita was upgraded and the code re-generated. Therefore other aspects of the generated code are improved as well.

Related issue: https://github.com/metaplex-foundation/solita/issues/48


austbot (approved):
This looks good.

token-metadata/js/src/custom/metadata-deserializer.ts
```diff
@@ -0,0 +1,96 @@
...
+    editionNonce,
+    tokenStandard,
+    collection,
+    uses,
+  };
+
+  return [Metadata.fromArgs(args), cursor];
+}
+
+function tryReadOption<T>(
```
samuelvanderwaal, line 84: Note that this doesn't exactly match the behavior of the Rust deserializer. There we check TokenStandard, Collection and Uses and set *all* to `None` if *any* of them are corrupted. This is because the corruption is likely to be all-or-nothing and it's theoretically possible for some of these fields to have accidentally valid, but logically incorrect, values.

thlorenz: That's what I did try to clarify previously, i.e. could one field be corrupted but another not.

If this is important then I can adapt this, but afaik in the case you're describing we'd get `None` for all of them anyways.

Please LMK though if this is a blocker to merge.

thlorenz: I updated the PR to have the same behavior as the Rust program.
https://github.com/metaplex-foundation/metaplex-program-library/pull/455/commits/35d12e0f841b6eb042bf146075e5fa7db49f394c

thlorenz:
Published as patch updated `@metaplex-foundation/mpl-token-metadata@2.1.2`

PR #459, opened 2022-05-13
Token metadata v1.3

samuelvanderwaal:
- [x] burn command and tests
- [x] size tracking and tests
- [x] set-tradeable handler and tests
- [x] backfill size handler and tests
- [x] set token handler and tests
- [x] fix broken test
- [x] change backfill to delegate model

# Token Metadata V1.3

## Dependency Updates

Solana dependencies updated to latest:

* solana-program: 1.9.15 --> 1.10.19
* spl-token: 3.2.0 --> 3.3.0
* spl-associated-token-account: 1.0.3 --> 1.0.5
* [dev] solana-sdk: 1.9.15 --> 1.10.19
* [dev] solana-program-test: 1.9.15 --> 1.10.19

### Metadata Update

A new field has been added to the Metadata struct to better support the collections standard. 

Current Metadata struct:

```
pub struct Metadata {
    pub key: Key,
    pub update_authority: Pubkey,
    pub mint: Pubkey,
    pub data: Data,
    // Immutable, once flipped, all sales of this metadata are considered secondary.
    pub primary_sale_happened: bool,
    // Whether or not the data struct is mutable, default is not
    pub is_mutable: bool,
    /// nonce for easy calculation of editions, if present
    pub edition_nonce: Option<u8>,
    /// Since we cannot easily change Metadata, we add the new DataV2 fields here at the end.
    pub token_standard: Option<TokenStandard>,
    /// Collection
    pub collection: Option<Collection>,
    /// Uses
    pub uses: Option<Uses>,
    /// Collection Details
    pub collection_details: Option<CollectionDetails>,
}
```

CollectionDetails enum:

```
pub enum CollectionDetails {
    V1 { size: u64 },
}
```

The goal with using enums is to try to make it more forward compatible but it may not be worth the extra effort to do so.

### Burn NFT

There's now a Burn NFT handler that allows the owner of the NFT to completely burn it:

- burning the SPL token and closing the token account
- closing the metadata and edition accounts
- giving the owner the reclaimed funds from closing these accounts

This handler checks if the NFT is a member of a verified collection, and if it is, requires the collection metadata account be passed in so the size can be decremented.

Concerns:

- Closing metadata and edition accounts may affect indexers NFT total counts.

Possible Mitigation:

Create a PDA burn record for each NFT burned.

Implementation:

Create a new program: `mpl_burn_record`.

The `burn_nft` handler CPIs into this program and creates a new data-less PDA every time an NFT is successfully burned. PDA Seeds: ["burned_nft", mint_account].

This allows:

1) tracking total number of Metaplex NFTs created by counting how many currently exist + how many accounts exist on the `mpl_burn_record` program.
2) checking if an existing NFT (mint account w/ supply = 1, decimals = 0, but *no* metadata account) is a generic Solana NFT, or a former Metaplex NFT but with its data accounts closed.

### Verify, Unverify and SetAndVerify Handlers for Sized Collections

The new handlers are:

* verify_sized_collection_item
* unverify_sized_collection_item
* set_and_verify_sized_collection_item

These handlers require the collection metadata account passed in mutably so they can modify the `size`. The other approach would have been to add optional accounts to the current verify handlers but this would have significantly complicated the logic in the accounts without providing much of an improved user experience. (The naming of the new handlers clearly indicates they handle a specific type of collection.) 

Concerns:

- It requires people to upgrade to new handlers to support collection details. 
  - They would have to update anyway by passing in a new account for the old handlers.

### Create Metadata V3

This handler needs a new version because it needs an additional argument `collection_details` for the user to indicate whether the NFT being created is a collection parent or not. This handler accepts an `Option<CollectionDetails>` where the `None` variant indicates it is *not* a collection parent NFT and `Some(CollectionDetails)` indicates it *is* a collection parent. This handler always creates a *new* metadata account with a size of 0.

### Set Collection Size

Initially this was written to require a Metaplex signer key in order to update a collection size. However, it has been refactored to use the delegate model so that Metaplex can build a tool to let people correctly calculate the collection size and then delegate authority to Metaplex to set it, but no longer *requires* a Metaplex sign off to do so.

Concerns:

- Collection authorities can set arbitrary sizes for their collections that don't match actual on-chain values. 

### Set Token Standard

This handler allows an update authority to pass in a metadata account with an optional edition account and then it determines what the correct TokenStandard type is and writes it to the metadata.

Concerns:

- The difference between Fungible and FungibleAsset currently lies in the JSON format so it may not be desirable to restrict FungibleAssets to decimals = 0 as we're doing currently.


token-metadata/program/src/deser.rs
```diff
@@ -42,18 +56,33 @@ pub fn meta_deser(buf: &mut &[u8]) -> Result<Metadata, borsh::maybestd::io::Erro
         token_standard,
         collection,
         uses,
+        collection_details,
     };
 
     Ok(metadata)
 }
 
+pub fn clean_write_metadata(
```
roederw, line 65:
```suggestion
pub fn clean_and_write_metadata(
```

austbot (approved):

Issue #124, opened 2022-01-20
Setup Auto Crate Publishing to MPL


austbot:

jshiohaha:
Currently, a bot will update the version found in Cargo.toml on PR merge if a contributor/mmeber left a review comment that says `patch`, `minor`, or `major`. It will also make the commit, add the repo tag, and push the commit.

This is the most simple version of the workflow. In the future, maybe we can specify more granular version updates that include package manager & package, such as `patch:rust` or `patch:rust:auction-house`. Any follow-on functionality will be dependent on how this first test goes.

jshiohaha:
I missed the fact that a Cargo.toml version bump also means we should update the IDL as well. Based on this sample version bump PR: https://github.com/metaplex-foundation/metaplex-program-library/pull/535/files

I'm thinking the workflow will be similar to the current implementation. We just need to include the IDL in the files that the bot commits and pushes as part of a version update.

thlorenz:
>  We just need to include the IDL in the files that the bot commits and pushes as part of a version update.

Well and obviously also the re-generated TS needs to be included.

An alternative solution is to build in a check that `yarn api:gen` doesn't result in any changes thus ensuring that the PR author took care of that. I'd prefer that actually since you might discover other issues (like failing SDK tests in the modules that have them) after you re-generate.

Issue #123, opened 2022-01-20
Setup Auto NPM publishing to MPL


austbot:

thlorenz:
Via conversation we decided to use [`npm add owner`](https://docs.npmjs.com/cli/v8/commands/npm-owner/) instead to allow the respective teams to publish the packages.

- `member-ship-token` (currently different package name _mpl-fixed-price-sale_)  already has [pre/post version/publish scripts setup](https://github.com/metaplex-foundation/metaplex-program-library/blob/4f4967b097bb25d0df66507ebfd471ad3b744fc5/fixed-price-sale/js/package.json#L8) so once we add the team workin on that as owner for that package they should be good to go

Make sure the teams follow the following steps when publishing:

```
npm version <patch|minor|major>
npm publish
```

NOTE: this has to be done with `npm` (not `yarn`) as only then do the pre/post scripts run properly.

jshiohaha:
Currently, a bot will run @thlorenz's previously configured `npm version patch|minor|major` on PR merge if a contributor/mmeber left a review comment that says `patch`, `minor`, or `major`. It will also make the commit, add the repo tag, and push the commit.

This is the most simple version of the workflow. In the future, maybe we can specify more granular version updates that include package manager & package, such as `patch:js` or `patch:js:auction-house`. Any follow-on functionality will be dependent on how this first test goes.

Closing for now since I think the original issue has been addressed.

PR #478, opened 2022-05-19
build: add husky for pre-commit and commit-message scripts

jshiohaha:
## Context
In our quest to add more MPL build automation (includes auto-publishing [packages](https://github.com/metaplex-foundation/metaplex-program-library/issues/123) + [crates](https://github.com/metaplex-foundation/metaplex-program-library/issues/124)), we want to introduce Git commit message standards. This will help us auto-version changes on PR merge, based on a collection of patch, minor, major changes. More details in [this Notion doc](https://www.notion.so/metaplex/CI-CD-Automation-Proposal-b1e62ae000544379afdb7726d7a4548c).

In favor of less custom code, I am proposing we follow the [conventional commit](https://www.conventionalcommits.org) standard. Please note that the specification mentions that fix, feat, and breaking change (or !) map to path, minor, major changes. We can change these keywords if we want based on commitlint configuration. There are both [CLI](https://commitlint.js.org) and [GH](https://github.com/marketplace/actions/commit-linter) tooling for this convention that will help us in this pursuit out-of-the-box but still allows for custom configuration.

I know changes to devex can be contentious, so I'm totally open to any feedback. These are just some proposals based on initial tooling research + poking around other OS repos.

## What changes are here
* Add [Husky](https://typicode.github.io/husky) for `pre-commit` and `commit-message` validation.
  * pre-commit: run linters on any Rust and TS changes
    *  You might note the usage of `LINT_RUST`. If this env var is not defined or LINT_RUST=false, we will not lint Rust. I added so that we can easily "flip a switch" to turn on Rust linting.
  * commit-message: check commit message adheres to desired format via commitlint.
* Add `commitlint` related NPM packages
* We are using the default commitlint configuration, as defined in `commitlint.config.js`. However, we can modify configuration based on our preferences. [Here](https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md) are some configuration details.

## What do we need to add (possibly later)
* Rust and TS linters. For JS linting, it seems like [lint-staged](https://www.npmjs.com/package/lint-staged) is the go-to package.
* Update README with some of these details so OS contributors aren't completely flustered by this change

## How it works
1. Pull down MPL repository and run `yarn`. As usual, all yarn packages should install - now including Husky.
2. Make some changes and stage changes for commit
3. The pre-commit script will decide what linters to run (if any) depending on the included changes.
4. Try to make a commit `git commit -m "this is a bad commit message for a feature"`, see output as shown here: https://app.warp.dev/block/9HS8FaVRbY9MhFgCQJJpwp
5. Change commit message to something like `git commit -m "feat: this is a new feature"` (pls use a more descriptive message)
6. See commit success, push, open PR.
7. If a contributor disables the husky commit-msg script and submits a PR with malformed commit messages, a GH workflow like [commit-linter](https://github.com/marketplace/actions/commit-linter) will fail and prevent the PR from merging.

## Alternatives
* Instead of [commitlint](https://commitlint.js.org), we could use [commitzen](https://commitizen-tools.github.io/commitizen). @samuelvanderwaal pointed out that he uses [GRC](https://crates.io/crates/grc), a rust implementation for [commitzen](https://commitizen-tools.github.io/commitizen) Git formatting. I haven't used either tool enough to have an opinionated decision. They might ultimately overlap in the outcomes they achieve - formatted Git messages.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,2 @@
+#!/bin/sh
+npx --no -- commitlint --edit
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,18 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+if [[ -z "$LINT_RUST" || "$LINT_RUST" = false ]]; then
+  echo "Skipping Rust lint since LINT_RUST is undefined or false"
+else
+  if [ -z "$(git diff --name-only --cached | grep '.*/program/src/.*/*.rs')" ]; then
+    echo "No Rust changes to lint"
+  else
+    echo "Linting Rust changes"
+  fi
+fi
+
+if [ -z "$(git diff --name-only --cached | grep '.*/js/src/.*/*.ts')" ]; then
+  echo "No TS changes to lint"
+else
+  echo "Linting TS changes"
+fi
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,1 @@
+module.exports = {extends: ['@commitlint/config-conventional']}
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "build": "yarn build:core && yarn build:auction && yarn build:parallel && yarn build:metaplex",
     "test": "lerna run --parallel test",
     "lint": "lerna run --parallel lint",
-    "fix": "lerna run --parallel fix"
+    "fix": "lerna run --parallel fix",
+    "postinstall": "husky install"
   },
   "workspaces": [
     "core/js",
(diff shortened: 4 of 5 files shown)


austbot (commented):
Conventional settings, can we use these for detecting patch, minor major?

jshiohaha:
Yes, my understanding is that 

* fix -> patch
* feat -> minor
* breaking change -> major

We can also explicitly whitelist patch, minor, major by modifying [the config for types](https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional#type-enum) if we prefer those terms.

jshiohaha:
Additional context: per-commit linting is too stringent. We will reconsider other approaches: 
* per-PR semantic versioning message (in title)
* manually bumping version numbers with automated publishing (using something like Lerna for JS)
* others,  after more 🔍 and 🧠🌩️

PR #481, opened 2022-05-23
token-metadata: Remove tilde in dependencies to fix future package resolving

joncinque:
#### Problem

`anchor-spl` depends on both `metaplex-token-metadata` and `spl-token`, and `metaplex-token-metadata` depends on `spl-token`, but both have specified dependencies with a `~`, meaning that only patch version bumps are allowed by the Cargo resolver, and not minor version bumps.

The Solana CI used to build Anchor and Metaplex to check for downstream breakage, but because of Anchor's metaplex dependency, it's impossible to select a viable package of `spl-token` when building Anchor, since the Solana crates use version `3.3`, but metaplex-token-metadata only allows `3.2`.

Currently, Anchor can't upgrade because of its dependency on token-metadata.

#### Solution

Since `metaplex-token-metadata` is used by many other programs, it will be much more usable to relax the version requirements to the default on the Cargo resolver, which allows minor version bumps, except for 0.* packages, where it only allows patch version bumps.  The SPL programs all use this mode for their dependencies. 

Let me know if you have any questions!

### DIFF
--- a/token-metadata/program/Cargo.toml
+++ b/token-metadata/program/Cargo.toml
@@ -14,22 +14,22 @@ test-bpf = []
 serde-feature = ["serde", "serde_with"]
 
 [dependencies]
-num-derive = "~0.3"
-arrayref = "~0.3.6"
-num-traits = "~0.2"
-solana-program = "~1.10.19"
-mpl-token-vault = { version = "~0.1.0", features = [ "no-entrypoint" ] }
-spl-token = { version="~3.3.0", features = [ "no-entrypoint" ] }
-spl-associated-token-account = { version = "~1.0.5", features = ["no-entrypoint"] }
-thiserror = "~1.0"
-borsh = "~0.9.2"
-shank = { version = "~0.0.2" }
-serde = { version = "~1.0.136", optional = true }
-serde_with = { version = "~1.12.0", optional = true }
+num-derive = "0.3"
+arrayref = "0.3.6"
+num-traits = "0.2"
+solana-program = "1.9.13"
+mpl-token-vault = { version = "0.1.0", features = [ "no-entrypoint" ] }
+spl-token = { version = "3.3.0", features = [ "no-entrypoint" ] }
+spl-associated-token-account = { version = "1.0.5", features = ["no-entrypoint"] }
+thiserror = "1.0"
+borsh = "0.9.2"
+shank = { version = "0.0.2" }
+serde = { version = "1.0.136", optional = true }
+serde_with = { version = "1.12.0", optional = true }
 
 [dev-dependencies]
-solana-sdk = "~1.10.19"
-solana-program-test = "~1.10.19"
+solana-sdk = "1.9.13"
+solana-program-test = "1.9.13"
 
 [lib]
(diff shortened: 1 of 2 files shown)


austbot (dismissed):
This makes sense to me I'm always a fan of package pinning

joncinque:
Note that this change makes package resolution *more* permissive, and so less pinned.  For a library, this is much nicer for downstream users.

When it comes to an application / CLI that can be installed through `cargo install`, then strict pinning to dependencies is extremely useful, and avoids running into a situation where cargo resolves to something newer which is incompatible with the code in the crate.

joncinque:
Note that this change makes package resolution *more* permissive, and so less pinned.  For a library, this is much nicer for downstream users.

When it comes to an application / CLI that can be installed through `cargo install`, then strict pinning to dependencies is extremely useful, and avoids running into a situation where cargo resolves to something newer which is incompatible with the code in the crate.

samuelvanderwaal:
@joncinque I'm working to fix dependency issues but I don't understand what you mean by 

> anchor-spl depends on both metaplex-token-metadata and spl-token

No versions of `anchor-spl` that I can see rely on `mpl-token-metadata`, [e.g.](https://crates.io/crates/anchor-spl/0.24.2/dependencies). However, I agree that the tilde requirement is probably too strict and it causes incompatibility issues with spl-token 3.2 and 3.3.

One of the issues we are running into is that `anchor-lang` uses tilde dependencies to lock `solana-program` to `1.9.x` which means we can't upgrade any of our programs to use the 1.10 series until Anchor updates. In general, this whole thing is a mess and I'm trying to find the best way to clean it up and allow maximum compatibility.

joncinque:
> No versions of anchor-spl that I can see rely on mpl-token-metadata

https://github.com/coral-xyz/anchor/blob/a1bb4946315a4f7f2dee1101454a856c21a6dc8e/spl/Cargo.toml#L27 -- even though it's optional, Cargo still needs to resolve a package version

joncinque:
> In general, this whole thing is a mess and I'm trying to find the best way to clean it up and allow maximum compatibility.

Yeah, it's a tough question, since I imagine that Anchor wants to prevent people using versions of Solana that have breaking changes or just don't work as expected. I'd recommend going to the normal dependency scheme on every side to allow for easier upgrades -- it does invite the whims of weird package resolution from Cargo sometimes, but overall it makes for a much better experience for downstream users.

samuelvanderwaal:
@joncinque if you can drop the solana deps down to `1.9.13` to match what Anchor requires, I'll fix the conflicts and merge this.

joncinque:
@samuelvanderwaal thanks for your feedback -- I did as you asked.  While I was at it, I also tried upgrading token-vault to spl-token 3.3.0, but that caused just about everything else to blow up.  This keeps the changeset contained to only what's needed to get Anchor to upgrade.

samuelvanderwaal (approved):

samuelvanderwaal:
Thanks for this PR and sorry it took so long to get to it.

joncinque:
Awesome, thanks so much!

PR #482, opened 2022-05-24
token-vault: Remove tilde in dependencies to fix future package resolving

joncinque:
#### Problem

This is a continuation of #481 to allow for easier updates for downstream dependencies, specifically for token-vault, which is a dependency of token-metadata.

It's impossible to update spl-token in downstream crates because of the tilde dependency resolution here.

#### Solution

Get rid of the tilde when specifying version numbers, and have `token-metadata` use the local `token-vault` for builds.  There's more info about why this is useful in #481 .  Feel free to ask any questions!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2091,7 +2091,7 @@ version = "1.2.10"
 dependencies = [
  "arrayref",
  "borsh 0.9.3",
- "mpl-token-vault 0.1.0",
+ "mpl-token-vault 0.1.1",
  "num-derive",
  "num-traits",
  "shank",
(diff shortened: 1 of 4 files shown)


thlorenz:
Don't have enough context here I assume and also realizing that #481 is merged already.
However version pinning has a huge disadvantage regarding bugs + security.

The `~` allows patches to be included automatically and those usually are used to fix bugs + security issues without breaking/adding API.
Pinning to a specific version makes us miss out on those and could result us running with an issue that has been fixed (we just didn't get it applied).

samuelvanderwaal:
> Don't have enough context here I assume and also realizing that #481 is merged already. However version pinning has a huge disadvantage regarding bugs + security.
> 
> The `~` allows patches to be included automatically and those usually are used to fix bugs + security issues without breaking/adding API. Pinning to a specific version makes us miss out on those and could result us running with an issue that has been fixed (we just didn't get it applied).

#481 isn't merged yet but I'm continuing the discussion there.

PR #484, opened 2022-05-25
chore(auction-house): adding comments for sale requires signer

stranzhay:
this is my understanding from reading code/docs

### DIFF
--- a/auction-house/program/src/sell/mod.rs
+++ b/auction-house/program/src/sell/mod.rs
@@ -316,6 +316,10 @@ fn sell_logic<'info>(
     let program_as_signer = &accounts.program_as_signer;
     let rent = &accounts.rent;
 
+    // 1. if wallet is signer is the only condition in which an NFT can sell at 0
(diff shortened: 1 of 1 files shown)


auction-house/program/src/sell/mod.rs
```diff
@@ -316,6 +316,10 @@ fn sell_logic<'info>(
     let program_as_signer = &accounts.program_as_signer;
     let rent = &accounts.rent;
 
+    // 1. if wallet is signer is the only condition in which an NFT can sell at 0
```
austbot, line 319: if the user does list at 0 then auction house can change the sale price if the can change sale price option is true

austbot (dismissed):
I added more context to your comment to make it more accurate

samuelvanderwaal:
Wasn't the idea with the original issue to refactor the block to be clearer code, not just comment it?

samuelvanderwaal (approved):

samuelvanderwaal (approved):

PR #487, opened 2022-05-25
fix: get_supply_off_master_edition(), Closes #486

### LINKED ISSUE #486: [Bug]: get_supply_off_master_edition() error with offsets

token-metadata

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

The MasterEdition structs don't store `supply` as an `Option<u64>` as assumed in `get_supply_off_master_edition()` and `get_max_supply_off_master_edition()` in the `utils.rs` module within the token_metadata program. This messes up with how EditionMarker PDAs are derived, specially in the fixed price sale program, that's where I encountered this problem.

This can be easily fixed by just changing the offsets accordingly.

### Relevant log output

_No response_

Medium (should be fixed soon)

flashleforshort:
`supply` is not of type `Option<u64>` but just plain u64, this messes up with the offsets used in this function and `get_max_supply_off_master_edition()`

### DIFF
--- a/token-metadata/program/src/utils.rs
+++ b/token-metadata/program/src/utils.rs
@@ -469,7 +469,7 @@ fn get_max_supply_off_master_edition(
     if data[9] == 0 {
         Ok(None)
     } else {
-        let amount_data = array_ref![data, 10, 8];
+        let amount_data = array_ref![data, 9, 8];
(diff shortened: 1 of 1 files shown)


roederw:
This is not a change we want to make. Closing.

PR #488, opened 2022-05-26
build: add verfiable builds via anchor and uploads to S3

jshiohaha:
### Context
* We want to store binaries related for programs at specific points in time. These changes will allow us to add workflows that build a program via anchor and upload that output + metadata to AWS S3.
* When including the `build-and-upload` action, 
  * we will save the output of `anchor build --verifiable` to bucket at `<bucket-name>/program-name/commit-sha/program-name.so`.
  * we will save a file called `index.json` with some basic metadata at `<bucket-name>/program-name/index.json`. Right now, this will contain the commit URL, ID, timestamp, and a URL to the most recent binary uploaded to the bucket for that program.
* These changes use the [OpenID Connect model](https://github.blog/changelog/2021-10-27-github-actions-secure-cloud-deployments-with-openid-connect) to connect GH with AWS, as opposed to the normal access + secret keys model. The infra changes to enable this are in a separate PR.

### What is added here
* Separate ctions to build a program via `anchor build --verifiable`, upload any file to S3, and a higher level action to combine those 2 things
* Updated the install Anchor CLI action

### How do we use this?
In any workflow, we simply have to add these few lines:

```
- name: Build and upload
  uses: ./.github/actions/build-and-upload/
  id: upload-binary-metadata
  with:
    program-binary: <PROGRAM_BINARY_NAME>
    account-id: <ACCOUNT_ID>
    region: <REGION>
    role: <ROLE>
    bucket: <BUCKET>
    prefix: <BUCKET_PREFIX>
```

### DIFF
--- a/.github/actions/install-anchor/action.yml
+++ b/.github/actions/install-anchor/action.yml
@@ -1,15 +1,11 @@
-name: Setup Anchor cli
+name: Setup Anchor CLI
+description: 'Setup Anchor VLI'
 
-inputs:
-  anchor_git:
-    description: Link to Anchor cli GH repository
-    required: true
-  
 runs:
-  using: "composite"
+  using: 'composite'
   steps:
     - uses: actions/cache@v2
-      name: Cache Cargo registry + index
+      name: Cache Anchor CLI
       id: cache-anchor
       with:
         path: |
@@ -18,10 +14,9 @@ runs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           ./target/
-        key: cargo-${{ runner.os }}-anchor-${{ hashFiles('**/Cargo.lock') }}
-    - run: cargo install --git ${{inputs.anchor_git}} anchor-cli --locked --force
+        key: anchor-cli-${{ runner.os }}-v0000-${{ env.ANCHOR_VERSION }}
+    - run:
+        cargo install --git https://github.com/project-serum/anchor --tag "v$ANCHOR_VERSION"
+        anchor-cli --locked
       shell: bash
-    - uses: actions/upload-artifact@v2
-      with:
-        name: anchor-binary
-        path: ~/.cargo/bin/anchor
+      if: steps.cache-anchor-cli.outputs.cache-hit != 'true'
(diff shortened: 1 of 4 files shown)


.github/actions/install-anchor/action.yml
```diff
@@ -18,10 +14,9 @@ runs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           ./target/
-        key: cargo-${{ runner.os }}-anchor-${{ hashFiles('**/Cargo.lock') }}
-    - run: cargo install --git ${{inputs.anchor_git}} anchor-cli --locked --force
+        key: anchor-cli-${{ runner.os }}-v0000-${{ env.ANCHOR_VERSION }}
+    - run:
+        cargo install --git https://github.com/project-serum/anchor --tag "v$ANCHOR_VERSION"
```
jshiohaha, line 19: Instead of requiring another param, my assumption was that we can always install from git? Can revert if that's not true.

.github/actions/upload-to-s3/action.yaml
```diff
@@ -0,0 +1,72 @@
...
+      env:
+        BUCKET: ${{ inputs.bucket }}
+        PREFIX: ${{ inputs.prefix }}
+        NAME: ${{ inputs.name }}
+        FILE_PATH: ${{ inputs.path-to-file }}
+
+    - name: Set output path to uploaded file
+      id: set-output-for-uploaded-file
+      run:
+        echo "::set-output name=uploaded-file-path::https://$BUCKET.s3.amazonaws.com/$PREFIX/$NAME"
```
jshiohaha, line 66: Sets the output so any following steps can fetch and use the URL

.github/actions/install-anchor/action.yml
austbot: thats fine its better than crates io as the yanker himself may screw us

.github/actions/upload-to-s3/action.yaml
austbot: love this!

austbot (approved):

NostalgicGareth (commented):
please help. my solana wallet is behind...

PR #490, opened 2022-05-26
(chore) add clippy to candy machine

stegaBOB:
Big IDL diff only because the items were reordered to match the actual rust file. The JSON file is identical. the order was initially changed to ensure the diff was small for #418


samuelvanderwaal (approved):

roederw (dismissed):

candy-machine/js/idl/candy_machine.json
```diff
@@ -626,65 +618,73 @@
...
           {
-            "name": "Date"
+            "name": "BurnEveryTime"
           },
           {
-            "name": "Amount"
+            "name": "NeverBurn"
           }
         ]
       }
```
thlorenz, line 690: What I'm noticing here is that errors still didn't get fixed.
Couldn't we upgrade to anchor 0.24 to get those proper again?

candy-machine/program/tests/core/helpers.rs
```diff
@@ -0,0 +1,313 @@
+use std::str::FromStr;
```
thlorenz, line 1: I like those helpers.
Maybe it's time to pull those into a separate crate we can use all over? 
Maybe `sol-test-utils` or something like that for the solana core specific ones?
And `mp-test-utils` for the ones involving metaplex programs?

thlorenz (changes requested):
If you could please update anchor and get the IDL fixed to include errors so we can generate those again with solita.
Also you should probably run the generator in any case  + include with this PR(unless the IDL just got reordered and results in same generated code).

I'm fine if the anchor upgrade needs to happen in a separate PR and will approve if this is too tricky to include here.

candy-machine/js/idl/candy_machine.json
thlorenz: Actually confirmed that errors were indeed already fixed https://github.com/metaplex-foundation/metaplex-program-library/blob/cf472e24686f997eceefc7306b89dd9ea326222a/candy-machine/js/idl/candy_machine.json#L693

thlorenz (approved):
LGTM now :)

PR #503, opened 2022-05-29
Add cupcake to the list

bhgames:
@austbot Summary
Cupcake is a cool new program built by @bhgames and @bartosz-lipinski  it allows use of candy machine, via a nfc and mobile mint link. This integration will grow candy machine usage and is net good for the protocol and candy machine users.

### DIFF
--- a/candy-machine/program/src/processor/mint.rs
+++ b/candy-machine/program/src/processor/mint.rs
@@ -178,7 +178,6 @@ pub fn handle_mint_nft<'info>(
                 &anchor_lang::solana_program::system_program::ID,
             )
             && !cmp_pubkeys(&program_id, &A_TOKEN)
-            && !cmp_pubkeys(&program_id, &CUPCAKE_ID)
         {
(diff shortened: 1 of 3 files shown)


austbot (approved):

candy-machine/program/src/processor/mint.rs
```diff
@@ -177,6 +178,7 @@ pub fn handle_mint_nft<'info>(
                 &anchor_lang::solana_program::system_program::ID,
             )
             && !cmp_pubkeys(&program_id, &A_TOKEN)
+            && !cmp_pubkeys(&program_id, &CUPCAKE_ID)
```
austbot, line 181: Do you need instructions before?

austbot: or after?

austbot (changes requested):
If you need an instruction after people can use it to screw us with bots, do you just need cpi or a post IX?

austbot (approved):

austbot (approved):

PR #504, opened 2022-05-29
Fix/auction house execute sale trade state

### LINKED ISSUE #344: Auction-house 'execute_sale' exception handling: check the existence of buyer trade state
In this line
https://github.com/metaplex-foundation/metaplex-program-library/blob/dc3bf54e6257b505989372cfe8a71b761d3ff0eb/auction-house/program/src/lib.rs#L660
the program has a check `buyer_ts_data.len() == 0`, but at the line

https://github.com/metaplex-foundation/metaplex-program-library/blob/dc3bf54e6257b505989372cfe8a71b761d3ff0eb/auction-house/program/src/lib.rs#L649
it unconditionally takes the [0] value and fails.

This may be reproduced with the following steps:
- put token on sale with wallet1
- transfer token for free from wallet1 to wallet2
- try to buy the token by wallet2 (it already owns it!)

An exception occurs:
>Program log: panicked at 'index out of bounds: the len is 0 but the index is 0', src/lib.rs:660:23

See for example this failed transaction:
https://explorer.solana.com/tx/ZnqKmEXQrxZLzXVoeeaNZMwwrE7M2TE2cPUYrnoDfvsw3b8BCjA9XTHEYLNV6TW4qbVEiiqV9665N1ef23MiDKb?cluster=devnet

stranzhay:
closes #344


blockiosaurus (dismissed):

auction-house/program/tests/execute_sale.rs
```diff
@@ -1699,3 +1699,161 @@ async fn auctioneer_execute_public_sale_no_delegate_fails() {
...
+        .process_transaction(sell_tx)
+        .await
+        .unwrap();
+
+    let buyer = Keypair::new();
+    airdrop(&mut context, &buyer.pubkey(), 10_000_000_000)
+        .await
+        .unwrap();
+
+    transfer(
```
austbot, line 1738: why transfer here?

austbot (approved):
seems good to me

auction-house/program/tests/execute_sale.rs
stranzhay: the issue had the context of "transfer token for free from wallet1 to wallet2" of when the error happened

austbot (approved):

PR #505, opened 2022-05-29
(WIP) Feat: Add freeze mints to candy machine

stegaBOB:


candy-machine/program/tests/core/master_edition_v2.rs
```diff
@@ -97,7 +95,11 @@ impl MasterEditionManager {
             )],
             Some(&self.authority.pubkey()),
             &[&self.authority],
-            context.last_blockhash,
+            context
+                .banks_client
+                .clone()
+                .get_new_latest_blockhash(&context.banks_client.get_latest_blockhash().await?)
```
thlorenz, line 101: I'd wrap this in a function. It's repeated all over at this point :)

candy-machine/js/src/generated/instructions/index.ts
```diff
@@ -2,8 +2,11 @@ export * from './addConfigLines';
 export * from './initializeCandyMachine';
 export * from './mintNft';
 export * from './removeCollection';
+export * from './removeFreeze';
 export * from './setCollection';
 export * from './setCollectionDuringMint';
+export * from './setFreeze';
+export * from './thawNft';
```
thlorenz, line 9: looks like those files aren't added?

candy-machine/js/src/generated/accounts/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './CandyMachine';
 export * from './CollectionPDA';
+export * from './FreezePDA';
```
thlorenz, line 3: Same as for instructions .. missing the new account file here.

candy-machine/js/src/generated/instructions/index.ts
stegaBOB: Whoops!

candy-machine/program/tests/core/master_edition_v2.rs
stegaBOB: Definitely a good idea!

thlorenz:
Wondering what the status on this is (it's still in draft mode).
If you want me to have another closer look please rebase on master.

Also we should probably think about some smoke-test that would exercise the TS SDK to the point where we detect at least that code is invalid and/or that files are missing.

Could be similar to what I do in [solita](https://github.com/metaplex-foundation/solita/blob/master/test/utils/verify-code.ts).

If so I could pull those utils into a separate package so we can use them in here for that.

(Would be a separate PR obviously).

anvog777:
Looking forward for this. Glad to help.

stegaBOB:
Time to do a super fun manual rebase :)

samuelvanderwaal:
@stegaBOB this has been replaced by #669 and can be closed, correct?

stegaBOB:
yup we can close this

PR #507, opened 2022-06-02
fix: don't rely on explicit export of TransactionCtorFields type

steveluscher:
This is our bad. We:

1. Over-export fundamentally private types from @solana/web3.js _just_ so that TypeDoc can generate documentation for them. This has to stop.
2. Didn't stop to think that someone might have imported those types for use.

In this PR I replace the recently-renamed `TransactionCtorFields` type with a type that _derives_ it, for all current and future versions of web3.js. This will not break again, no matter how many times the parameter gets renamed.

Fixes: https://github.com/metaplex-foundation/js/issues/202

### DIFF
--- a/auction/js/src/transactions/CancelBid.ts
+++ b/auction/js/src/transactions/CancelBid.ts
@@ -12,7 +12,6 @@ import {
   SystemProgram,
   SYSVAR_CLOCK_PUBKEY,
   SYSVAR_RENT_PUBKEY,
-  TransactionCtorFields,
   TransactionInstruction,
 } from '@solana/web3.js';
 import { AuctionProgram } from '../AuctionProgram';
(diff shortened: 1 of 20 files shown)


steveluscher:
Please also ship https://github.com/metaplex-foundation/metaplex-program-library/pull/507.

lorisleiva (approved):

stegaBOB (approved):

austbot (changes requested):
broken integ

steveluscher:
Interesting.

🤔

```
src/Transaction.ts(4,15): error TS2502: 'options' is referenced directly or indirectly in its own type annotation.
```

steveluscher:
Seems like the answer may be to do exactly what I hoped not to have to do. https://stackoverflow.com/a/48640951/802047

stegaBOB:
The metaplex/js is being deprecated soon and the js-next repo will be replacing it. I think the issues in this mpl repo are from old code that isn't being auto generated yet. It may just be easier to wait a few days for us to be ready to replace the js repo and we can probably generate the code for these contracts too.

PR #510, opened 2022-06-03
Using swap remove for mint index

febo:
This PR adds a feature flag to enable the swap-remove mint index generation, which offers constant time generation + shuffles the order of indices.


candy-machine/program/src/processor/mint.rs
```diff
@@ -684,34 +686,68 @@ pub fn get_good_index(
...
+            + 4
+            + (items_available as usize) * CONFIG_LINE_SIZE
+            + 4
+            + ((items_available
+                .checked_div(8)
+                .ok_or(CandyError::NumericalOverflowError)?
+                + 1) as usize)
+            + 4;
+        // calculates the mint index and retrieves the value at that position
+        let mint_index = indices_vec_start + index * 4;
```
austbot, line 714: index could be much larger, where are you clamping the int to be within the range of acceptable values

austbot (changes requested):

candy-machine/program/src/processor/mint.rs
febo: The `index` is the reminder of the division by `candy_machine.data.items_available` (line `497`).

febo:
> is this ready for a real review @febo

@austbot It is ready!

stegaBOB:
For big candy machines, I think this may be a breaking change. It seems like realloc can only modify the size by up to 10kb, which may cause older clients to panic on massive candy machines.

febo:
> For big candy machines, I think this may be a breaking change. It seems like realloc can only modify the size by up to 10kb, which may cause older clients to panic on massive candy machines.

@stegaBOB Good point - it seems that the limit is per call, so if we break the increase request into multiple `realloc` calls it should be ok. I will test this modification.

stegaBOB:
I'm quite interested to see if that would work. The current account allocation/realloc limit is 10kb for validator memory reasons and how it has to be preassigned before the instruction if I understand it correctly. If realloc can be called multiple times this would appear to get around said limit.

febo:
@stegaBOB Done - the process is divided in two parts. The funding for the `realloc` happens in the `initialize`, since we need the `system_program` account to do that; the actual `realloc` happens in the `add_config_lines` multiple times until the account has the expected size. This only applies if the account does not have the expected size, i.e., it was created by a previous version of the CLI.

There is only one edge case that we can't avoid: if you try to add the config lines in reverse order (the last index first), you will get an error since we might not be able to increase the account to the size required in a single `realloc` - the CLIs do not try to do that, this can only happen if you "manually" call `add_config_lines`. There is a check on the program for that and you will get an informative error.

cc: @austbot

samuelvanderwaal:
This has an API change with a couple new errors so will require updating the JS SDK but we're going to hold off until we get Sarah's bot tax fixes in and deployed to mainnet.

stegaBOB:
We need to update candy machine to anchor 0.25 before this goes in because of the Solana 1.9 reallocate account vulnerability.

febo:
This has been incorporated into Candy Machine Core.

PR #511, opened 2022-06-04
Candy machine lock settings

jpbogle:
cc @gchatz22 as a second contributor

This is purely a suggestion - curious to get thoughts from others

This makes use of a different protocol called token-manager to handle the lockup and release. More info can be found here -> https://github.com/cardinal-labs/cardinal-token-manager.

Adding “**LockupSettings**” to candy machine.

LockupSettings allows for specifying a time-based lockup after mint. It supports duration, locking the token for X seconds after it is minted, or expiration allowing for a set datetime for all tokens to be released.

There are a few use cases for this
1. Collections that do NOT want their collection being listed on marketplaces before the mint is over. Somewhat of a hot topic but it seems people are upset when the collection gets listed and the floor price moves during mint, because this can affect the rest of the mint (positively and negatively).
2. Long term lockups (think 6 months) forcing projects to deliver on their promises before collecting royalties and have the project actually produce value for holders before they sell.
3. Having pre-sale and public-sale be released at the same time. This allows for public and WL minters to get equal chance at listing.

Release of the tokens is permissionless and can be triggered by anyone. We are currently running this every 5 seconds to invalidate all expired token-managers. There is a small reward (the sol in the account) to incentivize others to trigger the release. Token manager has received 1 audit.

Note about the current solana clock drift -- for using LockupType::Duration nothing changes since its purely added time to the current solana clock at the time of mint. For LockupType::Expiration it means that the creator of the candy machine should be aware that if the clock is drifted, the time specified in UTC seconds should be aligned to Solana clock that can be triggered once solana clock reaches that time.

PR includes 3 commits
1. implementation
- The space allocated for `EndSettings` was only 10, not sure why this was the case since it should be size of 16, same for `LockupSettings` so the PR includes setting that to 10. As far as I'm aware this should be backwards compatible since existing CMs wont use this extra part of the struct, and new ones will just get allocated the extra space needed to use it.

2. client gen

3. amman tests addition
- @thlorenz using accountsCluster and executable:true instead of local path for cloning the program was not working, please advise on this - have experienced on other repos as well not sure if it's missing some config for a remote clone


candy-machine/program/src/processor/mint.rs
```diff
@@ -608,6 +622,152 @@ pub fn handle_mint_nft<'info>(
...
+        }
+        let time_invalidator_program = &ctx.remaining_accounts[remaining_accounts_counter];
+
+        remaining_accounts_counter += 1;
+        if ctx.remaining_accounts.len() <= remaining_accounts_counter {
+            return err!(CandyError::LockupSettingsMissingTokenManagerProgram);
+        }
+        let token_manager_program = &ctx.remaining_accounts[remaining_accounts_counter];
+        // If we add more extra accounts later on we need to uncomment the following line out.
+        // remaining_accounts_counter += 1;
```
jpbogle, line 678: not sure why this was commented out or what the harm is for just adding + one to avoid confusion for future additions, but it was like this above so we left it

candy-machine/program/src/constants.rs
```diff
@@ -22,7 +22,7 @@ pub const CONFIG_ARRAY_START: usize = 8 + // key
     8 + // price
     8 + // items available
     9 + // go live
-    10 + // end settings
+    16 + // end settings
```
jpbogle, line 25: confused why this was set to 10, checking out the space `std::mem::size_of::<EndSettings>` returns 16

candy-machine/js/.ammanrc.js
```diff
@@ -1,8 +1,28 @@
...
+        '../../../cardinal-token-manager/target/deploy/cardinal_token_manager.so',
+      ),
+    },
+    {
+      label: 'Time Invalidator',
+      programId: tokenManager.programs.timeInvalidator.TIME_INVALIDATOR_ADDRESS,
+      deployPath: path.join(
+        '../../../cardinal-token-manager/target/deploy/cardinal_time_invalidator.so',
+      ),
+    },
```
jpbogle, line 25: Needs to be replaced but this didnt work for local clone
```js
    {
      label: 'Token Manager',
      programId: tokenManager.programs.tokenManager.TOKEN_MANAGER_ADDRESS,
      executable: true
    },
```

thlorenz:
Only the later versions of amman support remote accounts. Candy machine is still using `0.1.` https://github.com/metaplex-foundation/metaplex-program-library/pull/511/files#diff-b47d37355a8e52dd1b64475a8f2682884c041b17e5344f19c4df4ab3c9f225b5R53.

So you'd need to update to the latest version (and use `amman-client` in the tests) see [here](https://github.com/metaplex-foundation/js-next/blob/thlorenz/feat/upload/test/helpers/amman.ts) where we use latest amman in our js-next code.

@stegaBOB maybe able to clarify as well (he added that feature).

candy-machine/program/src/processor/mint.rs
stegaBOB: Commented out because of the dead code lint

stegaBOB:
I don't think this would be mergeable to be honest. Looks like quite a big breaking change that would break all current existing candy machines. The big issue is the change in the config array start length. I actually have a PR out that basically does this all natively with the token metadata program itself. https://github.com/metaplex-foundation/metaplex-program-library/pull/505

Its still a WIP and I'm still adding other features.

stegaBOB:
The main reason for the incompatibility is the large array of config lines uses the size of the config array start to determine where to begin looking. As for the end settings size, we will have to look into that? As far as I know it should serialize to 10 bytes using borsh rules? But from what it sounds like perhaps thats not correct?

candy-machine/program/src/constants.rs
stegaBOB: Does this check the already deserialized size though? I think the size would be different after it is serialized in borsh, no? Almost all our types show different sizes than the ones we calculated in the constants.rs. I feel like candy machines with all the settings active would have broken by now (which we have tested and they shouldn't break?)

jpbogle:
@stegaBOB ya good point -- this was just proposing / seeing if you are open to us merging a time-based lockup feature in some fashion, we have updated the PR to follow the feature flag convention that is backwards compatible with the existing state.

Though same for collections feature flag, why is the UUID not zeroed out always on initialize? Seems like if I accidentally passed in a UUID with a 1 in the front I would be inadvertently turning on collections feature flag. 

We followed a similar design for this feature being a separate PDA and turning on the feature flag to enforce that account is passed in

>Note: I think std::men::size_of takes the in memory size which would be the deserialized type. The borsh serialized type should be different.
Interesting ya can see changing this would break just wasn't sure why it was 10 but never realized the mem size would be different.

stegaBOB:
Re: UUID on initialization. All of our first party tools as well as all of the tools I'm aware of have been setting the UUID as the first 6 characters in the base 58 pubkey (which can't be 1 as it's base 58). It hasn't been enforced at a program level yet, although I believe I have included that inforcement in the freeze PR. if someone sets it as a 1 it's sorta been considered a user error.

candy-machine/program/src/constants.rs
jpbogle: reverted back to 10 since you are right config size offset change would not be backwards compatible with existing reads without a v3

candy-machine/program/src/processor/mint.rs
jpbogle: ah makes sense, in this case the linting doesn't complain just leaving it as a += 1 maybe because its a &mut
https://github.com/metaplex-foundation/metaplex-program-library/pull/511/files#diff-a5b4d95737a7474da174757a8ff945dad5df3db8b0d6b0f8feecf4c91d2bca44R908

jpbogle:
>Could we do it in a way where instead of calling all the cardinal CPI, we just set the delegate/freeze authority to be what is configured in the lockup settings and call freeze?

If im understanding correctly this is mostly just because you want LockupSettings to be generic so that anyone can "hook" into it? I get it, maybe it can just be part of lockup settings and use the Type to vector on different lockup_managers so Cardinal can just be one implementation of one, but afaict a CPI would be required because the account its being delegated to has to trigger freeze -- only way around this i can think of without changing token_metadata was maybe to freeze the token_account BEFORE making the master_edition, then create_master_edition and then delegate to the token to the specified account in LockupSettings.lockup_manager so anyone making the candy machine can specify who can unfreeze it later on, but you cant delegate after its frozen so basically it has to CPI so the account configured in lockups can sign to call freeze

--
>im kind of worried about the extra CPI calls as we are running low (Or were brefore) on compute.

The CI we added passes with solana 1.9.28 the new compute limits should give this plenty of room given that it is now 1.4mm compute per instruction instead of 200k. One discussion we had was that if compute was an issue it could be modeled using that instruction sysvar to say that if freeze settings are set there HAS to be another instruction after, but in this case, these CPIs are still well under the new compute budget so imo its okay, and this code is only executed if lockup settings are set

--
>I think your time invalidator is really cool however. The other thing about this is that we are learing to avoid the single mutable state account. In this case the royalties may need to be sharded into different fee accounts how would your system hadle the time invalidated lockup of those accounts?

The way this feature is currently designed, it shouldn't really affect royalties because what the token is non-tradeable during lockup right so there will be no royalties to collect. Maybe i am misinterpreting but it seems like some "royalty escrow/lockup" is kind of a separate feature that could be implemented via the royalties going to an shared wallet similar to hydra with a lockup

-- 
>I really think we need the time based lockups of royalties which is great that you have posted that, fees and tokens. We were thinking a super simple approach if just having candy machine be freeze authority of the token and freeze it, and then royalties go into pairwise pdas depending on the size of the mint, and then there is an explicit instruction admins loop over to get the royalties and fees out which is only active after the lockup predicate passes(mint out) or the time expiration has passed.

Same thing here aren't there no royalties if the token is frozen. As for primary sale fees, having lockup there separately makes sense. It seems like that can be achieved already by just having the `wallet` of the candy_machine be a wallet PDA with a release date unless you are referring to having it be different per-mint.

Are you are suggesting there is a desire to have them be "tiered" or per-mint and there needs to be tracked per-mint? Is there a system in mind if lockup would be different per-mint?

--
>Would you consider some of those changes so that Cardinals mechanism cans still be a lockup option? But it keeps it general based on the freeze authority yall built with us?

The feature that seems to make clear sense to add would be that the time-invalidator is already per-mint it just happens they are either duration or set expiration and all the same but there is a concept of extension so it makes to have them be extended differently per-mint and collect a different fee per-mint if that can be modeled in the state of lockupsettings

stegaBOB:
Do we have a slack connection or something? Would love to hop in a call at some point and discuss this in more detail instead of GitHub comments.

stegaBOB:
The thing I'm mostly concerned about for using the lockup manager is the number of accounts needed for freezing. That's going to hit the 1232 (I think) byte limit super fast.

jpbogle:
> The thing I'm mostly concerned about for using the lockup manager is the number of accounts needed for freezing. That's going to hit the 1232 (I think) byte limit super fast.

you mean in the future right? since right now integration tests pass with these accounts so its under the limit currently

stegaBOB:
Sure but this will require most of the use cases to split the candy machine mints into 2 txns. Integration tests pass because it splits it up (they aren't detailed enough to test splits and stuff)

jpbogle:
Just came back to add some new things
- Renamed the LockupType's to be more specific (Duration -> DurationSeconds and Expiration -> ExpirationUnixTimestamp) 
- Updated to allow adding Compute program ID before the mint instructions to request more compute
- Allowed for init_if_needed on lockup settings so someone can update / change it for different tiers of mints
- Added test for fixed expiration timstamp

jpbogle:
thanks for the comment @austbot

PR #512, opened 2022-06-05
switches types so idel gen happens with no issue

austbot:
# Problem
The imported creators type from token metadata has no anchor serialize, or account trait impl and this stops it from getting proper treatment from the idl generator.

# Solution
create a new type that gets saved in the runtime that can be properly idl'ed


thlorenz (commented):
LGTM

PR #513, opened 2022-06-06
Update fixed price sale anchor version to 0.24

VadimGrozinok:
## What
Update used anchor_lang lib to 0.24
## Why
Old anchor version had issue in init_if_needed function
## How
Update Cargo.toml and fix all the problems involved


austbot (approved):
I think my creator change worked but this is maybe better to keep the same type everywhere

samuelvanderwaal (approved):

PR #514, opened 2022-06-06
Fix/permissionless sale bug

### LINKED ISSUE #480: [Bug]: Permissionless execute_sale with any wallet not working

auction-house

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

_This is a follow-up from Discord where I was told it seems to be a bug._

The auction house documentation states that the execute sale command is permissionless (if the 'requires sign-off' is false) and that **anyone** can call this command. 

I am trying to call this command with a wallet that is neither auction house authority, buyer nor seller. And I am getting an error `NoPayerPresent (6012)`. I tried to add the wallet as a fee payer to the transaction, I can see on solscan that I added the wallet as a fee payer to the instruction's account list, but it does not help. 

I looked at the code of the Solana program and it seems that the fee payer may be either the authority or buyer or seller (https://github.com/metaplex-foundation/metaplex-program-library/blob/facd7cb11903bab086165e6619bbf0cc6e28c272/auction-house/program/src/utils.rs#L93) 

I am using the latest auction house CLI.

Use case: a backend server runs the execute sale command but we do not want to have a private key of the authority stored on a server due to security reasons

### Relevant log output

_No response_

High (immediate attention needed)

stranzhay:
closes #480


roederw:
Blocking this until we finish discussions about some other items that might conflict with this.

PR #530, opened 2022-06-10
chore: publish mpl-auctioneer

exromany:
@thlorenz  is it enough to publish the package to npm?

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -8,6 +8,7 @@
     "token-metadata/js",
     "metaplex/js",
     "auction-house/js",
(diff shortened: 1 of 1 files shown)


stegaBOB:
Looks like auctioneer is already on the packages. I also don't think we have CI yet set up for publishing to NPM

samuelvanderwaal:
It's now published, but it's a very early version with no README or docs: https://www.npmjs.com/package/@metaplex-foundation/mpl-auctioneer

PR #533, opened 2022-06-14
Adding reserve price to example auctioneer.

### LINKED ISSUE #449: [Feature]: Add Reserve Price to Auctioneer

auction-house

### Feature

As part of the example Auctioneer's base functionality, it should be possible to set a reserve (minimum) price for the listing.

### Standard Change?

No

- A reserve price should be added as an argument to the sell instruction
- A reserve price field should be added to ListingConfig
- A check should be performed to verify all bids are above the reserve price

_No response_

### Other context

_No response_

blockiosaurus:
Closes #449


auctioneer/program/src/sell/config.rs
```diff
@@ -2,7 +2,7 @@ use anchor_lang::prelude::*;
 use solana_program::clock::UnixTimestamp;
 
 pub const BID_SIZE: usize = 8 + 1 + 32;
-pub const LISTING_CONFIG_SIZE: usize = 8 + 1 + 8 + 8 + BID_SIZE + 1;
+pub const LISTING_CONFIG_SIZE: usize = 8 + 1 + 8 + 8 + BID_SIZE + 1 + 8;
```
austbot, line 5: will this require a re alloc for existing bids if the size is smaller ?

austbot (commented):
This looks really good, one question

auctioneer/program/src/sell/config.rs
blockiosaurus: It would, but my intention is to knock out all of these tickets this week before anyone really starts using Auctioneer. I guess I could add some padding as well so this week's changes don't break anything.

samuelvanderwaal (approved):

PR #536, opened 2022-06-15
fix(token-metadata): change set collection size args for shank support

samuelvanderwaal:


stegaBOB (approved):
lgtm!

token-metadata/js/src/generated/instructions/SetCollectionSize.ts
```diff
@@ -7,14 +7,15 @@
...
 import * as beet from '@metaplex-foundation/beet';
 import * as web3 from '@solana/web3.js';
+import { SetCollectionSizeArgs, setCollectionSizeArgsBeet } from '../types/SetCollectionSizeArgs';
 
 /**
  * @category Instructions
  * @category SetCollectionSize
  * @category generated
  */
 export type SetCollectionSizeInstructionArgs = {
```
thlorenz, line 17: Now I'm not so sure anymore why we had to change this. 
The code seems valid to me with a generic `instructionArgs` name and the related beet below seems legit as well.

Did that generated code not work for some reason?

thlorenz: If so please file an issue with solita as we'd need to fix it

samuelvanderwaal: 1) It wasn't Solita it was Shank. The generated IDL didn't work with the docs which reads from it to automatically create the instructions. It's possible I messed up something there but that's what originally got me to ask you about whether Shank supports single args with no struct type.

2) I don't like the name being "instructionArgs" as it's not very descriptive; I want "size" in there somewhere.

stegaBOB: So regardless of whether we can fix the shank issue, based on #2 it seems like a worthwhile change. I'm curious to see what the root of the issue with shank was though.

PR #540, opened 2022-06-16
Add versioning and build actions to workflows

jshiohaha:
## Context
* This change wires up the actions I previously added for verifiable builds and NPM & Crates versioning
  * https://github.com/metaplex-foundation/metaplex-program-library/pull/488
  * https://github.com/metaplex-foundation/metaplex-program-library/pull/515
* When a PR is merged, it triggers a `push` event to master. In this case, the workflow checks for (1) changes in the package and (2) a `push` event before trying to publish a new version and build, in the case of Rust

## Changes
* Most changes are adding the top-level actions for publishing version updates for JS (`publish-to-npm`) & Rust (`publish-to-cratess`) and publishing builds for Rust (`build-and-publish`)
* During a push event, we don't have access to the associated PR number. So, I added a custom `get-pr-for-commit` action that takes the event's commit SHA and relates it to a PR, if it exists.
* I also created an action called `setup-aws-env` that sets up common AWS env vars so that I didn't have to copy pasta in every single workflow

## Future
* After getting a feel for the comment-to-version packages these workflows enable, we can add more granular comment parsing like `patch:js` or `patch:rust:auction-house` 
* Maybe add the link to a build artifact on the original PR
* Do we need extra work/infra to sync deployed binary and the compatible IDL?

### DIFF
--- a/.github/workflows/sdk-auction-house.yml
+++ b/.github/workflows/sdk-auction-house.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [master]
 
+permissions:
+  id-token: write
+  contents: write
+
 jobs:
   changes:
     runs-on: ubuntu-latest
@@ -24,6 +28,7 @@ jobs:
             - 'core/**'
           package:
             - 'auction-house/**'
+
   build-lint-and-test-auction-house:
     needs: changes
     if: ${{ needs.changes.outputs.core == 'true' || needs.changes.outputs.package == 'true' }}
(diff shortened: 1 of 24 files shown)


.github/workflows/sdk-auction-house.yml
```diff
@@ -33,6 +38,26 @@ jobs:
       - uses: actions/checkout@v1
       - uses: ./.github/actions/yarn-install-and-verify
         with: 
-          cache_id: sdk-auction  
-          working_dir: ./auction/js
+          cache_id: sdk-auction-house
+          working_dir: ./auction-house/js
```
jshiohaha, line 42: This seemed like a typo? I can revert if this was intended.

austbot: yes you are right

jshiohaha:
Oh cool, it was closed! For viz, this is the commit message

* for NPM: https://github.com/metaplex-foundation/metaplex-program-library/blob/master/.github/actions/publish-to-npm/action.yaml#L71
* for Crates: https://github.com/metaplex-foundation/metaplex-program-library/blob/master/.github/actions/publish-to-crates/action.yaml#L85

jshiohaha:
Retroactively linking the related issues

* https://github.com/metaplex-foundation/metaplex-program-library/issues/124
* https://github.com/metaplex-foundation/metaplex-program-library/issues/123

PR #542, opened 2022-06-18
Adding minimum bid increment

### LINKED ISSUE #451: [Feature]: Add automatic time extension to Auctioneer

auction-house

### Feature

As part of the example Auctioneer's base functionality, it should be possible to enable a time extension (bids executed towards the end of the auction extend the auction end time) on Auctioneer listings.

### Standard Change?

No

- Time extension delta and time extension period arguments should be added to the sell instruction
- Time extension delta and time extension period fields should be added to ListingConfig
- The end time should be pushed back by the delta if a bid is seen in the time extension period.

_No response_

### Other context

_No response_

blockiosaurus:
Closes #451

### DIFF
--- a/auctioneer/js/idl/auctioneer.json
+++ b/auctioneer/js/idl/auctioneer.json
@@ -556,6 +556,18 @@
         {
           "name": "reservePrice",
           "type": "u64"
+        },
+        {
+          "name": "minBidIncrement",
+          "type": "u64"
+        },
+        {
+          "name": "timeExtPeriod",
+          "type": "u32"
+        },
+        {
+          "name": "timeExtDelta",
+          "type": "u32"
         }
       ]
     },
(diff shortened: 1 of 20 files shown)


auctioneer/js/idl/auctioneer.json
```diff
@@ -731,6 +743,18 @@
           {
             "name": "reservePrice",
             "type": "u64"
+          },
```
austbot, line 746: Should we make them all options so that when we extend this in the future its not wewiord that only some of them are options?

blockiosaurus: I figured their default values of 0 took care of it and the SDK would just send down the default unless otherwise specified, but you're right, options probably make more sense in the Rust code.

blockiosaurus: OK, done.

roederw (approved):

PR #545, opened 2022-06-22
Fix burning editions

samuelvanderwaal:
v1.3.1

Prevents print editions and master editions with a supply > 0 from being burned in `burn_nft` until we decide how to handle burning print editions. (If it's decided to do this it should be in a new handler e.g. `burn_print_nft`). 

Adds two new errors.

Fixes a bug where collection items belonging to unsized collections couldn't be burned.

This changes the TypeScript package slightly by adding two new errors to the IDL and changing some account names and descriptions. 

Some comments and explanations on the changes:

This verbosity 

```rust
    // Ensure this is a Master Edition and not a Print.

    // Scope this so the borrow gets dropped and doesn't conflict with the mut borrow
    // later in the handler when overwriting data.
    {
        let edition_account_data = edition_info.try_borrow_data()?;

        // First byte is the object key.
        let key = edition_account_data[0];
        if key != Key::MasterEditionV1 as u8 && key != Key::MasterEditionV2 as u8 {
            return Err(MetadataError::NotAMasterEdition.into());
        }

        // Next eight bytes are the supply, which must be converted to a u64.
        let supply_bytes = array_ref![edition_account_data, 1, 8];
        let supply = u64::from_le_bytes(*supply_bytes);

        // Cannot burn Master Editions with existing prints in this handler.
        if supply > 0 {
            return Err(MetadataError::MasterEditionHasPrints.into());
        }
    }
```

is to save compute while checking for both versions of `MasterEdition`. If we only have to worry about MasterEditionV2 this can be replaced with:

```rust
let master_edition = MasterEditionV2::from_account_info(edition_info).map_err(|| MetadataError::NotAMasterEdition)?;

if master_edition.supply > 0 {
            return Err(MetadataError::MasterEditionHasPrints.into());
}
```

This function fails if the collection parent is unsized (doesn't have CollectionDetails populated).

```rust
decrement_collection_size(&mut collection_metadata, collection_metadata_info)?;
```

So I pulled the logic out from it to allow decrementing the size but allowing burning anyway if it's unsized. 

I'm not sure how good my error handling is (paging @thlorenz) but I have eight tests for this handler now. Let me know if there's some edge case I'm missing and I can write more tests.

### DIFF
--- a/token-metadata/js/idl/mpl_token_metadata.json
+++ b/token-metadata/js/idl/mpl_token_metadata.json
@@ -1671,10 +1671,10 @@
           "desc": "Token account to close"
         },
         {
-          "name": "editionAccount",
+          "name": "masterEditionAccount",
           "isMut": true,
           "isSigner": false,
-          "desc": "MasterEdition2 or Edition Account of the NFT"
+          "desc": "MasterEdition2 of the NFT"
         },
         {
           "name": "splTokenProgram",
@@ -3326,6 +3326,16 @@
       "code": 108,
       "name": "MissingEditionAccount",
       "msg": "This mint account has an edition but none was provided."
+    },
+    {
+      "code": 109,
+      "name": "NotAMasterEdition",
+      "msg": "This edition is not a Master Edition"
+    },
+    {
+      "code": 110,
+      "name": "MasterEditionHasPrints",
+      "msg": "This Master Edition has existing prints"
     }
   ],
   "metadata": {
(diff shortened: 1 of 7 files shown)


blockiosaurus (dismissed):
Looks good to me!

roederw:
ship it

thlorenz:
> I'm not sure how good my error handling is 
There _is_ error handling .. that's the important part :)

thlorenz (approved):

austbot (approved):
looks good to me.
patch

PR #547, opened 2022-06-22
Stranzhay/pl 8 add bot tax to gated token failures

### LINKED ISSUE #475: [Feature]: Add bot tax to Gated token failures

candy-machine

### Feature

Instead of failing when no gatekeeper token is present, the Candy Machine should (potentially) apply the bot tax and succeed.

* Concerns are that this would be more prone to catching users and applying the bot tax to them.

### Standard Change?

No

* Discuss and implement if necessary

_No response_

### Other context

_No response_

stranzhay:
closes #475


candy-machine/program/src/processor/mint.rs
```diff
@@ -229,17 +235,30 @@ pub fn handle_mint_nft<'info>(
...
                 BOT_FEE,
+                None,
             )?;
             return Ok(());
         }
         let gateway_token_info = &ctx.remaining_accounts[remaining_accounts_counter];
+
         remaining_accounts_counter += 1;
 
+        let gateway_verification_result = {
```
stranzhay, line 251: I added this block today and from the output, its saying that the programId is invalid from gateway_token_info. Could also explain why there is no data being created on the account creation. The PDA has been cross checked with the JS version and it is correct. The programId being passed in from the test is also correct as far I can tell.

candy-machine/program/tests/initialize.rs
```diff
@@ -54,3 +76,140 @@ async fn init_default_success() {
...
+    let mut context = candy_machine_program_test().start_with_context().await;
+    let context = &mut context;
+
+    let gatekeeper_network = Keypair::new();
+    let gatekeeper_authority = Keypair::new();
+
+    let client = RpcClient::new("https://metaplex.devnet.rpcpool.com".to_string());
+
+    let gateway_account_pubkey =
+        Pubkey::from_str("gatem74V238djXdzWnJf94Wo1DcnuGkfijbf3AuBhfs").unwrap();
```
samuelvanderwaal, line 91: Nit/cool thing to know:
You can replace all the 

`Pubkey::from_str("constant").unwrap()`

with the `pubkey` macro:

```rust
use solana_program::pubkey;
...
let gateway_account_pubkey = pubkey!("gatem74V238djXdzWnJf94Wo1DcnuGkfijbf3AuBhfs");
 ```
 
 For tests it doesn't really matter so you don't need to change this if you don't want to, but compile time checking this for prod code is much better. (We don't do this consistently in our codebase yet, but we should.)

**Review by samuelvanderwaal — COMMENTED** [review]
The test only checks one branch of the `if...else if` statement for Gateway tokens, right? It checks that bots get punished for trying with an expired token, but what about writing a test for if they try with an invalid token?

File: candy-machine/program/tests/utils/candy_manager.rs [hunk 3]
```diff
@@ -219,6 +221,97 @@ impl Clone for TokenInfo {
...
+pub struct GatekeeperInfo {
+    pub set: bool,
+    pub network_expire_feature: Option<Pubkey>,
+    pub gateway_app: Pubkey,
+    pub gateway_token_info: Pubkey,
+    pub gatekeeper_config: GatekeeperConfig,
+}
+
+impl GatekeeperInfo {
+    #[allow(dead_code)]
```
samuelvanderwaal, line 260: :-(

samuelvanderwaal (approved):

samuelvanderwaal (approved):

candy-machine/program/tests/utils/candy_manager.rs
stegaBOB: lol

stegaBOB (approved):
LGTM

PR #548, opened 2022-06-23
Make Auction House JS lib consistent with Token Metadata

lorisleiva:
Currently, the Auction House autogenerated library uses the autogenerated code and wraps it in a big class that also contains PDA helpers.

This not only breaks tree-shaking because the entire package is loaded into static fields on the class, but it also is inconsistent with how we handle autogenerated libraries with other programs such as the Token Metadata program.

This PR fixes this by removing that custom class with static fields and exporting the generated files in the same way other programs do.

Note that this is a breaking change and so we will need a new major version when publishing this. Note that this also removes all PDA helpers. However, I think this is the right thing to do as these should also be autogenerated as explained in https://github.com/metaplex-foundation/solita/issues/61.


thlorenz (approved):
LGTM, ship it!

PR #549, opened 2022-06-23
refactor(token-metadata): set-collection-size

samuelvanderwaal:
A few more changes to be included in v1.3.1:

* ~~Set Collection Size allows setting size for both Some and None variants of the collection_details field~~.
* Add overflow checks to release profile in Cargo.toml
* Add optional account to Shank annotations for Set Collection SIze
* Perform checked addition and subtraction for collection size math
* Regenerate the JS SDK

Pivot:

There are attacks possible with allowing `set-collection-size` to always update the size. Instead, it's going to only allow unsized collection Parents to be set (If the collection details is not populated or set to `None`).  This means a collection size can only be set once, after that a new collection has to be created and the items migrated over. This is annoying for users who mess up but makes it so the on-chain size can actually be trusted once it's set correctly.

### DIFF
--- a/token-metadata/Cargo.toml
+++ b/token-metadata/Cargo.toml
@@ -3,3 +3,6 @@ members=[
   "program",
   "cli"
 ]
+
+[profile.release]
+overflow-checks = true
--- a/token-metadata/program/Cargo.toml
+++ b/token-metadata/program/Cargo.toml
@@ -8,6 +8,8 @@ license = "AGPL-3.0"
 edition = "2021"
 readme = "README.md"
 
+[profile.release]
+overflow-checks = true
 
 [features]
 no-entrypoint = []
--- a/token-metadata/program/Cargo.toml
+++ b/token-metadata/program/Cargo.toml
@@ -8,9 +8,6 @@ license = "AGPL-3.0"
 edition = "2021"
 readme = "README.md"
 
-[profile.release]
-overflow-checks = true
-
 [features]
 no-entrypoint = []
 test-bpf = []
--- a/token-metadata/program/src/processor.rs
+++ b/token-metadata/program/src/processor.rs
@@ -1809,8 +1809,11 @@ pub fn process_burn_nft(program_id: &Pubkey, accounts: &[AccountInfo]) -> Progra
         if let Some(ref details) = collection_metadata.collection_details {
             match details {
                 CollectionDetails::V1 { size } => {
-                    collection_metadata.collection_details =
-                        Some(CollectionDetails::V1 { size: size - 1 });
+                    collection_metadata.collection_details = Some(CollectionDetails::V1 {
+                        size: size
+                            .checked_sub(1)
+                            .ok_or(MetadataError::NumericalOverflowError)?,
+                    });
                     clean_write_metadata(&mut collection_metadata, collection_metadata_info)?;
                 }
             }
@@ -1869,7 +1872,7 @@ pub fn set_collection_size(
             }
         }
     } else {
-        return Err(MetadataError::NotACollectionParent.into());
+        metadata.collection_details = Some(CollectionDetails::V1 { size });
     }
 
     clean_write_metadata(&mut metadata, parent_nft_metadata_account_info)?;
--- a/token-metadata/program/src/processor.rs
+++ b/token-metadata/program/src/processor.rs
@@ -1863,14 +1863,9 @@ pub fn set_collection_size(
         }
     }
 
-    if let Some(details) = metadata.collection_details {
-        match details {
-            CollectionDetails::V1 {
-                size: _current_size,
-            } => {
-                metadata.collection_details = Some(CollectionDetails::V1 { size });
-            }
-        }
+    // Only unsized collections can have the size set, and only once.
+    if metadata.collection_details.is_some() {
+        return Err(MetadataError::SizedCollection.into());
     } else {
         metadata.collection_details = Some(CollectionDetails::V1 { size });
     }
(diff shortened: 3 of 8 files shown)


token-metadata/program/Cargo.toml
```diff
@@ -8,6 +8,8 @@ license = "AGPL-3.0"
 edition = "2021"
 readme = "README.md"
 
+[profile.release]
+overflow-checks = true
```
thlorenz, line 12: Does this have an effect when running in `ebpf` VM?
I thought I read something somewhere that it doesn't .. not 100% sure tough

thlorenz (approved):
LGTM pending the caveat about release overflow-checks.

Also would be nice to have some test showing behavior in both cases, i.e. with `collection_details` and without.

token-metadata/program/Cargo.toml
samuelvanderwaal: That is a good question. I'll look into it.

thlorenz: Obviously now I cannot find anything about this .. so maybe it's not an issue at all :)

samuelvanderwaal:
> LGTM pending the caveat about release overflow-checks.
> 
> Also would be nice to have some test showing behavior in both cases, i.e. with `collection_details` and without.

I'm happy to add tests, but can you elaborate? The `collection_details` field is always going to be populated because 

```rust
let mut metadata = Metadata::from_account_info(parent_nft_metadata_account_info)?;
```

[uses](https://github.com/metaplex-foundation/metaplex-program-library/blob/d8d6e1c29fcf4254c728e0ea134f1e00f00a0225/token-metadata/program/src/state.rs#L258) the [custom deser function](https://github.com/metaplex-foundation/metaplex-program-library/blob/d8d6e1c29fcf4254c728e0ea134f1e00f00a0225/token-metadata/program/src/deser.rs#L9) which ensures that the value is at least set to None.

thlorenz:
> I'm happy to add tests, but can you elaborate? 

There is an `if` [statement in the code](https://github.com/metaplex-foundation/metaplex-program-library/blob/107242c5638751ebbeed5c128cf3ff9f8f9ecffa/token-metadata/program/src/processor.rs#L1866-L1876) where it checks if `metadata.collection_details` is set or not.

Depending on that it takes different paths, a branch was added, so would be nice to have a test that that branch hits/behaves as expected.

Not sure if there are tests present for the previous behavior, just saying we should test anytime we add logic.

token-metadata/program/Cargo.toml
samuelvanderwaal: Seems to be used by other Solana devs: https://github.com/coral-xyz/anchor/pull/1758

samuelvanderwaal:
> > I'm happy to add tests, but can you elaborate?
> 
> There is an `if` [statement in the code](https://github.com/metaplex-foundation/metaplex-program-library/blob/107242c5638751ebbeed5c128cf3ff9f8f9ecffa/token-metadata/program/src/processor.rs#L1866-L1876) where it checks if `metadata.collection_details` is set or not.
> 
> Depending on that it takes different paths, a branch was added, so would be nice to have a test that that branch hits/behaves as expected.
> 
> Not sure if there are tests present for the previous behavior, just saying we should test anytime we add logic.

This branch can be replaced with just:

```rust
    metadata.collection_details = Some(CollectionDetails::V1 { size });
```

and maybe should be, since both forks of the branch do the same thing. The only upside is that having the match in there means if we add a V2 later, this code stops compiling which helps with code refactoring when updating versions.

thlorenz:
> This branch can be replaced with just:  ...

I'm not saying that the code should be changed, I'm talking about a unit test that ensures that the code behaves as expected since there is new logic added to the code.

danenbm (dismissed):

token-metadata/program/src/processor.rs
```diff
@@ -1869,7 +1872,7 @@ pub fn set_collection_size(
             }
         }
     } else {
-        return Err(MetadataError::NotACollectionParent.into());
+        metadata.collection_details = Some(CollectionDetails::V1 { size });
```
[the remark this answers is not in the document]
samuelvanderwaal: > > This branch can be replaced with just:  ...
> 
> I'm not saying that the code should be changed, I'm talking about a unit test that ensures that the code behaves as expected since there is new logic added to the code.

I'll add a test.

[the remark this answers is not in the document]
samuelvanderwaal: > Would it make sense to add a log message here just to note that collection details are being added that were not previously here?

I think this isn't necessary anymore with the refactor in the latest commit. If you disagree, lmk and I'll add a log message.

thlorenz (dismissed):

stegaBOB (approved):

token-metadata/Cargo.toml
```diff
@@ -2,4 +2,7 @@
 members=[
   "program",
   "cli"
-]
\ No newline at end of file
+]
+
+[profile.release]
+overflow-checks = true
```
thlorenz, line 8: Not fully understanding why it has to be in this Cargo.toml (vs. inside the program), but maybe this is how our workspaces function?

Would hate to have this compile with this actually turned off and get deployed that way. Is it possible to add this in both places just to be sure?

samuelvanderwaal: When you put it inside the `Cargo.toml` inside of `token-metadata/program` it gives you a warning that it will only be applied if it's in the workspace root `Cargo.toml` file. The matches [the docs](https://doc.rust-lang.org/cargo/reference/profiles.html).

PR #550, opened 2022-06-23
Make changes label more specific to each package workflow

jshiohaha:
@thlorenz pointed out that the `changes` label in all workflows is the same, making it hard to differentiate between packages (image below). This PR proposes a change to update labels from `changes` to a package and workflow specific, e.g. `<package>-<scope>-changes`.

Specifically, in `./github/workflows/sdk-token-metadata.yml`, `changes` is now `token-metadata-sdk-changes`.

image


jshiohaha:
Okay, false alarm. Labeling shows up as expected again 🤷🏼‍♂️

PR #553, opened 2022-06-24
fix: for anchor-client 0.24 works well

Tonyce:
with anchor 0.24，decode/encode check the account name，and its `camelcase(name, { pascalCase: true })`

https://github.com/coral-xyz/anchor/blob/master/ts/src/coder/borsh/accounts.ts#L45-L64
https://github.com/coral-xyz/anchor/blob/master/ts/src/coder/borsh/accounts.ts#L97-L100

### DIFF
--- a/candy-machine/program/src/processor/collection/remove_collection.rs
+++ b/candy-machine/program/src/processor/collection/remove_collection.rs
@@ -4,7 +4,7 @@ use solana_program::program::invoke;
 
 use crate::{
     cmp_pubkeys, constants::COLLECTIONS_FEATURE_INDEX, remove_feature_flag, CandyError,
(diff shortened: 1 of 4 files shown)


Tonyce:
https://github.com/metaplex-foundation/metaplex/issues/2082 ，The problem is similar

stegaBOB:
This is a breaking change and not an issue at the program level. Bumping the JS anchor client version will require some slight changes on the client side. Technically this is still camel case, even though acronyms should technically be lowercase after the first letter according to the Rust naming guidelines. This is specific to Rust though and is more ambiguous elsewhere.

Tonyce:
There are plans to upgrade to anchor 0.24 for candy mathine cli ？

PR #554, opened 2022-06-24
Add serde

omertxyz:
example: Ok("{\"UpdateMetadataAccount\":{\"data\":null,\"update_authority\":\"4uQeVj5tqViQh7yWWGStvkEG1Zmhx6uasJtWCJziofM\",\"primary_sale_happened\":true}}")

### DIFF
--- a/token-metadata/program/Cargo.toml
+++ b/token-metadata/program/Cargo.toml
@@ -24,6 +24,8 @@ spl-associated-token-account = { version = "~1.0.5", features = ["no-entrypoint"
 thiserror = "~1.0"
 borsh = "~0.9.2"
 shank = { version = "~0.0.2" }
+serde = "~1.0.136"
+serde_with = "~1.12.0"
 
 [dev-dependencies]
 solana-sdk = "~1.10.19"
--- a/token-metadata/program/Cargo.toml
+++ b/token-metadata/program/Cargo.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 [features]
 no-entrypoint = []
 test-bpf = []
+serde-feature = ["serde", "serde_with"]
 
 [dependencies]
 num-derive = "~0.3"
@@ -24,8 +25,8 @@ spl-associated-token-account = { version = "~1.0.5", features = ["no-entrypoint"
 thiserror = "~1.0"
 borsh = "~0.9.2"
 shank = { version = "~0.0.2" }
-serde = "~1.0.136"
-serde_with = "~1.12.0"
+serde = { version = "~1.0.136", optional = true }
+serde_with = { version = "~1.12.0", optional = true }
 
 [dev-dependencies]
 solana-sdk = "~1.10.19"
--- a/token-metadata/program/src/deprecated_instruction.rs
+++ b/token-metadata/program/src/deprecated_instruction.rs
@@ -8,15 +8,16 @@ use solana_program::{
     pubkey::Pubkey,
     sysvar,
 };
+use serde::{Serialize, Deserialize};
 
 #[repr(C)]
-#[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone)]
+#[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone, Serialize, Deserialize)]
 pub struct MintPrintingTokensViaTokenArgs {
     pub supply: u64,
 }
 
 #[repr(C)]
-#[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone)]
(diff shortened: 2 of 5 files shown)


token-metadata/program/Cargo.toml
```diff
@@ -24,6 +24,8 @@ spl-associated-token-account = { version = "~1.0.5", features = ["no-entrypoint"
 thiserror = "~1.0"
 borsh = "~0.9.2"
 shank = { version = "~0.0.2" }
+serde = "~1.0.136"
```
austbot, line 27: We need to make th3se optional based on crate feature

austbot: Examples

```sea-orm = { optional = true```

```
[features]
json_types = ["serde", "serde_json"]
sql_types = ["sea-orm"]
```

File: token-metadata/program/src/deprecated_instruction.rs [hunk 2]
```diff
@@ -8,15 +8,16 @@ use solana_program::{
     pubkey::Pubkey,
     sysvar,
 };
+use serde::{Serialize, Deserialize};
 
 #[repr(C)]
-#[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone)]
+#[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone, Serialize, Deserialize)]
```
austbot, line 14: same here we can avoid this code getting into the smart contract via this

```#[cfg_attr(feature = "example", derive(Debug))]```

**Review by austbot — CHANGES_REQUESTED** [review]
sopme comments to make this work for our contracts

File: token-metadata/program/src/deprecated_instruction.rs (hunk 2 above)
Comment (omertxyz) [reply token-metadata/program/src/deprecated_instruction.rs:14]: Unfortunately this doesn't work with serde_as macro (tried it already): https://docs.rs/serde_with/1.10.0/serde_with/guide/serde_as/index.html#gating-serde_as-on-features

File: token-metadata/program/Cargo.toml (hunk 1 above)
Comment (omertxyz) [reply token-metadata/program/Cargo.toml]: @austbot fixed

**Review by austbot — APPROVED** [review]

**Review by danenbm — APPROVED** [review]

File: token-metadata/program/src/instruction.rs [hunk 3]
```diff
@@ -12,27 +12,37 @@ use solana_program::{
...
+    serde_with::{As, DisplayFromStr},
+};
 
 #[repr(C)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Debug, Clone)]
 /// Args for update call
 pub struct UpdateMetadataAccountArgs {
     pub data: Option<Data>,
+    #[cfg_attr(feature = "serde", serde(with = "As::<Option<DisplayFromStr>>"))]
```
danenbm, line 27: Is this syntax required since serde is gated as a crate feature, otherwise it could just be `as = DisplayFromStr`?

omertxyz: Yes, serde_with doesn't work as is with features, see: https://docs.rs/serde_with/1.10.0/serde_with/guide/serde_as/index.html#gating-serde_as-on-features

token-metadata/program/src/state.rs
```diff
@@ -587,6 +604,7 @@ impl ReservationListV2 {
 }
 
 #[repr(C)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
samuelvanderwaal, line 607: Why is the feature added here but not for `ReservationListV2`?

samuelvanderwaal: And not for `ReservationListV1`?

samuelvanderwaal: Also, `EditionMarker` is missing it.

omertxyz: I might have missed it, but I added it to all the structs I needed to for it to compile, I'm happy to add it to them too

samuelvanderwaal: Are you actually using the `Reservation` structs? 

`EditionMarker` definitely should have it.

omertxyz: Since I was only using MetadataInstruction, I added only to the structs that the compiler complained didn't have Serde attr on them. But I'll add to these too

omertxyz: It seems like ReservationListV2 has some serialize function already defined, so it rejects my addition of Serde, how do you want to handle that?

samuelvanderwaal: I think you can skip all the `Reservation` structs and just add it to `EditionMarker` and we'll be good. Thanks for this PR!

samuelvanderwaal (approved):

PR #557, opened 2022-06-26
Add combination authority

LUK3ARK:
If we add this the vault authority can be responsible for combination and the vault can only be combined when specified with the REAL price account. Please please please I have a contract ready :)


staccDOTsol:
Goodmorning @austbot I ask your attention here, too, equally as important.

Issue #552, opened 2022-06-24
[Bug]: the code is contradictory


Tonyce:
### Which package is this bug report for?

token-metadata

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/processor.rs#L1476-L1490

### Priority this issue should have

Low (slightly annoying)

samuelvanderwaal:
You are correct; thanks for the callout.

Tonyce:
the some superfluous and contradictory code also in other files

PR #560, opened 2022-06-27
fix(token-metadata): remove superfluous and contradictory code

samuelvanderwaal:
Addresses #552. The tests passed because the code doesn't ever get reached. No IDL or API changes.


austbot (approved):
This is the size PR I LIKE

roederw (approved):

PR #577, opened 2022-06-30
build: reduce workflow runtime by modifying IDL version directly

jshiohaha:
### Context
* after chatting with @thlorenz about #558, he suggested editing the IDL directly since we are only updating version based on the associated crate's version.
* nothing changes here, functionally. the main win of this change is we remove a bunch of `cargo install <pkg>` commands and reduce overall workflow runtime. at worst, i think this will reduce runtime from 15-17 min to 5-7 min 🕐 
* the modified package's SDK is validated when the PR is (re)opened and synchronized (aka updated) based on this change #576 

### Testing from jshiohaha/metaplex-program-library fork
PR (see bot commits): https://github.com/jshiohaha/metaplex-program-library/pull/18
Action for `version patch`: https://github.com/jshiohaha/metaplex-program-library/runs/7139333038?check_suite_focus=true

### DIFF
--- a/.github/actions/make-version-changes/action.yaml
+++ b/.github/actions/make-version-changes/action.yaml
@@ -33,24 +33,16 @@ runs:
       run: |
         cargo install cargo-release
 
-    # cache and install shank cli
-    - uses: actions/cache@v2
-      name: Cache Shank CLI
-      id: cache-shank-cli
-      with:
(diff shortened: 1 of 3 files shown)


jshiohaha:
merged in master to resolve conflicts

PR #579, opened 2022-07-01
refactor(token-metadata): create TokenMetadataAccount trait

samuelvanderwaal:
### Safe Deserialize

This PR adds a new trait, `TokenMetadataAccount` which is implemented for all account types owned by the Token Metadata program. The primary method of this trait, `safe_deserialize`, is a type-safe deserialization function that ensures all Token Metadata Accounts can only be deserialized into their correct types. It does this by checking both the `Key` field on each account and also the expected account length. It uses the standard `BorshDeserialize` implementation for the actual deserialization.

```rust
use borsh::maybestd::io::Error as BorshError;

pub trait TokenMetadataAccount {
    fn key() -> Key;

    fn size() -> usize;

    fn pad_length(buf: &mut Vec<u8>) {
        let padding_length = Self::size() - buf.len();
        buf.extend(vec![0; padding_length]);
    }

    fn safe_deserialize<T: BorshDeserialize + TokenMetadataAccount>(
        mut data: &[u8],
    ) -> Result<T, BorshError> {
        if !is_correct_account_type(data, T::key(), T::size()) {
            return Err(BorshError::new(ErrorKind::Other, "DataTypeMismatch"));
        }

        let result: T = T::deserialize(&mut data)?;

        Ok(result)
    }
}
```

The `safe_deserialize` and `pad_length` methods are generic and the default implementations will work for every account the trait is implemented on as long as that account also implements `BorshDeserialize`. Only the `key` and `size` methods must be manually implemented for every account. 

All `from_account_info` methods have been updated to use this `safe_deserialize` method as well as to independently check that each account that is deserialized is actually owned by the Token Metadata program. 

```rust
impl Metadata {
    pub fn from_account_info(a: &AccountInfo) -> Result<Metadata, ProgramError> {
        let metadata = Metadata::safe_deserialize(&a.data.borrow_mut())
            .map_err(|_| MetadataError::DataTypeMismatch)?;

        // Check that this is a `token-metadata` owned account.
        assert_owned_by(a, &ID)?;

        Ok(metadata)
    }
}
```

The other benefit of `safe_deserialize` is that it provides clients and other consumers of the `mpl-token-metadata` crate a type-safe way to deserialize an account solely from a mutable buffer of data, without having to pass in the expected `Key` and account size.
An example of how a consumer of `mpl-token-metadata` might use `safe_deserialize`:

```rust
use mpl_token_metadata::state::{Metadata, TokenMetadataAccount};

. . .
let buffer = client.get_account_data(pubkey)?;
let metadata: Metadata = Metadata::safe_deserialize(&mut buffer)?;
```

### Try From Slice Checked

The utility function `try_from_slice_checked` has been kept for backwards compatibility but should be deprecated in preference to `safe_deserialize`.

### State Tests

A new file `state_test.rs` has been added and all state tests moved over to it. In a future PR, I'd like to refactor the folder structure to be more modular and split up `state.rs`, which has grown overly long, into separate files, perhaps based around each token account type.


token-metadata/program/src/deser.rs
```diff
@@ -6,7 +6,10 @@ use solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, pubke
 // This function is used in a custom deserialization implementation for the
 // `Metadata` struct, so should never have `msg` macros used in it as it may be used client side
 // either in tests or client code.
-pub fn meta_deser(buf: &mut &[u8]) -> Result<Metadata, borsh::maybestd::io::Error> {
+//
+// It does not check `Key` type or account length and should only be used through the custom functions
+// `from_account_info` and `deserialize` implemented on the Metadata struct.
+pub fn meta_deser_unchecked(buf: &mut &[u8]) -> Result<Metadata, borsh::maybestd::io::Error> {
```
austbot, line 12: Perhaps we should only pub(crate) so people dont have a foot gun

token-metadata/program/src/state.rs
```diff
@@ -257,15 +259,31 @@ pub struct Metadata {
 
 impl Metadata {
     pub fn from_account_info(a: &AccountInfo) -> Result<Metadata, ProgramError> {
-        let md: Metadata = meta_deser(&mut a.data.borrow_mut().as_ref())?;
+        let data = a.data.borrow_mut();
```
austbot, line 262: Should from account info on metadata check owner in every case? I think it should. It would let us remove owner checks in a-lot of other spots.

token-metadata/program/src/state.rs
```diff
@@ -776,3 +794,244 @@ impl EditionMarker {
...
+        );
+
+        let err = Metadata::deserialize(&mut buf.as_ref()).unwrap_err();
+        assert_eq!(err.kind(), ErrorKind::Other);
+        assert_eq!(err.to_string(), "DataTypeMismatch");
+
+        let err = Metadata::from_account_info(&account_info).unwrap_err();
+        assert_eq!(err, MetadataError::DataTypeMismatch.into());
+    }
+}
```
austbot, line 1037: Wrt ownership checks in account metadata add a test to verify the wrong owner(program) cannot be deserialized

token-metadata/program/src/state.rs
samuelvanderwaal: Do you mean check that the account is owned by `token-metadata` program?

token-metadata/program/src/state.rs
```diff
@@ -776,3 +794,244 @@ impl EditionMarker {
         Ok(())
     }
 }
+
+#[cfg(test)]
+
+mod metadata_deserialization {
+    use solana_sdk::{signature::Keypair, signer::Signer};
```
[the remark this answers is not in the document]
samuelvanderwaal: They do depend on that, but that should be a safe assumption given that users aren't able to set their own `Key` types, right?

token-metadata/program/src/state.rs
samuelvanderwaal: Will do.

token-metadata/program/src/state.rs
samuelvanderwaal: Yeah, this makes sense.

token-metadata/program/src/state.rs
samuelvanderwaal: `fail_to_deserialize_metadata_with_wrong_owner` test added and other tests updated to use `token-metadata` program ID

token-metadata/program/src/utils.rs
```diff
@@ -144,7 +144,7 @@ pub fn assert_initialized<T: Pack + IsInitialized>(
 }
 
 /// Create account almost from scratch, lifted from
-/// https://github.com/solana-labs/solana-program-library/tree/master/associated-token-account/program/src/processor.rs#L51-L98
+/// <https://github.com/solana-labs/solana-program-library/tree/master/associated-token-account/program/src/processor.rs#L51-L98>
```
samuelvanderwaal, line 147: This makes the link clickable and makes a warning go away when generating docs.

token-metadata/program/src/state.rs
```diff
@@ -776,3 +802,278 @@ impl EditionMarker {
...
+        let error = Metadata::from_account_info(&md_account_info).unwrap_err();
+        assert_eq!(error, MetadataError::IncorrectOwner.into());
+
+        // `deserialize` should succeed because it just deserializes bytes, with no account context.
+        let metadata = Metadata::deserialize(&mut buf.as_ref()).unwrap();
+        assert_eq!(metadata, expected_metadata);
+    }
+
+    #[test]
+    fn fail_to_deserialize_master_edition() {
```
danenbm, line 895: Not super familiar with all the account types, but what about other ones from the `Key` enum such as `MasterEditionV1` and `ReservationListV2`?

Or are the tests sufficient with `MasterEditionV2`, `EditionV1`, `UseAuthorityRecord`, and `CollectionAuthorityRecord`, `EditionMarker`.

Just curious why that subset?  Not sure if the tests are meant to be exhaustive and if they are or not.

token-metadata/program/src/deser.rs
danenbm: +1 to pub(crate)

token-metadata/program/src/state.rs
samuelvanderwaal: `ReserverationV1`, `ReservationV2`, and `MasterEditionV1` are not widely in use (or at all?) as they're from old really early versions of the code. I figured the subset of the most common was sufficient to demonstrate the code works as expected but it wouldn't be hard to make the list exhaustive if you think I should.

danenbm: I think its fine, proving a couple cases showed you got the business logic of the check correct.  Any more is probably overkill, like proving that enums work properly.

token-metadata/program/src/utils.rs
```diff
@@ -830,14 +830,16 @@ pub fn assert_token_program_matches_package(token_program_info: &AccountInfo) ->
     Ok(())
 }
 
+pub fn is_correct_account_type(data: &[u8], data_type: Key, data_size: usize) -> bool {
+    (data[0] == data_type as u8 || data[0] == Key::Uninitialized as u8) && (data.len() == data_size)
```
danenbm, line 834: I think I see why this function has to pass `data[0] == Key::Uninitialized` as well as the actual correct `Key`.
1.  Looks like in several places `create_or_allocate_account_raw()` is called on various account types.
1. Right after potentially creating the account, `from_account_info()` is called, before the `key` is set a few lines later.

wdyt about having a `from_uninitialized_account_info()` that is used in those specific cases?  Maybe also another trait function: `safe_uninitialized_deserialize()` that **requires** the key is `Key::Uninitialized` but verifies the size info.

This might be overkill but seems like it makes `is_correct_account_type` a stronger check, and then the few places where uninitialized accounts are deserialized are handled in an explicit way.

danenbm (commented):
Overall I really like the direction you are going with this PR, using a trait so that the correct key and size are bundled with the type rather than parameters that have to be chosen correctly whenever other functions such as `from_account_info()` are implemented.

I think this is a good object oriented approach.  The `TokenMetadataAccount` trait provides some guarantees on what the different account types can provide.

token-metadata/program/src/state.rs
```diff
@@ -176,13 +219,34 @@ pub struct UseAuthorityRecord {
...
+        Key::UseAuthorityRecord
+    }
+
+    fn size() -> usize {
+        USE_AUTHORITY_RECORD_SIZE
+    }
+}
+
 impl UseAuthorityRecord {
     pub fn from_account_info(a: &AccountInfo) -> Result<UseAuthorityRecord, ProgramError> {
```
danenbm, line 243: Why can't `from_account_info()` also be a deefault implementation in the  `TokenMetadataAccount` trait?  It seems like it always calls `safe_deserialize()` then check owner.  Maybe I'm missing some other detail though.

token-metadata/program/src/state.rs
```diff
@@ -75,6 +83,29 @@ pub const USE_AUTHORITY_RECORD_SIZE: usize = 18; //8 byte padding
 
 pub const COLLECTION_AUTHORITY_RECORD_SIZE: usize = 11; //10 byte padding
 
+pub trait TokenMetadataAccount {
+    fn key() -> Key;
+
+    fn size() -> usize;
+
+    fn pad_length(buf: &mut Vec<u8>) {
+        let padding_length = Self::size() - buf.len();
```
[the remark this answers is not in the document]
samuelvanderwaal: Good call.

token-metadata/program/src/state.rs
```diff
@@ -75,6 +83,29 @@ pub const USE_AUTHORITY_RECORD_SIZE: usize = 18; //8 byte padding
...
+        buf.extend(vec![0; padding_length]);
+    }
+
+    fn safe_deserialize<T: BorshDeserialize + TokenMetadataAccount>(
```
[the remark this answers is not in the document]
samuelvanderwaal: `safe_deserialize` needs to be an associated function because you're trying to get the type from raw bytes, you're not starting with the type. So the typical use case is something like this:

```rust
use mpl_token_metadata::state::{Metadata, TokenMetadataAccount};

. . .
let buffer = client.get_account_data(pubkey)?;
let metadata: Metadata = Metadata::safe_deserialize(&mut buffer)?;
```

I suppose you could create a default or empty `Metadata` type and then you could use it as a method:

```rust
use mpl_token_metadata::state::{Metadata, TokenMetadataAccount};

. . .
let md_empty = Metadata::new();
let buffer = client.get_account_data(pubkey)?;
let metadata = md_empty.safe_deserialize(&mut buffer)?;
```
but this seems awkward and unnecessary.

The same thing goes from `from_account_info` (which I agree should be part of the Trait). Am I missing something?

token-metadata/program/src/deser.rs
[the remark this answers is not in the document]
samuelvanderwaal: Agreed.

token-metadata/program/src/state.rs
samuelvanderwaal: It can but it makes things a bit more messy as it requires manually specifying the account type in a lot of places and it also causes issues with [this](https://github.com/metaplex-foundation/metaplex-program-library/blob/ac94ca76c9f1e371cee2e8e169435c209b48f150/token-metadata/program/src/state.rs#L285) because of the other trait derived on these accounts, `MasterEdition`.

token-metadata/program/src/state.rs
```diff
@@ -104,6 +135,18 @@ pub struct Data {
     pub creators: Option<Vec<Creator>>,
 }
 
+impl Default for Data {
```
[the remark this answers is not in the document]
samuelvanderwaal: I don't think it matters much.

token-metadata/program/src/utils.rs
samuelvanderwaal: With green field design this does make sense but feels like overkill for the current state of the program since we still check the account length and none of our accounts have the same length. I'm already having to stop myself from refactoring large swaths of `token-metadata` because it's out-of-scope for this PR which needs to get merged so v1.3 can get to mainnet. However, if you think this is important I can add it.

token-metadata/program/src/state.rs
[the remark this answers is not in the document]
samuelvanderwaal: Also, none of these functions actually need any information from `self` so it seems anti-pattern to make them methods.

token-metadata/program/src/state.rs
samuelvanderwaal: I figure out how to get around the issues with the function and added in all the types, so this works.

token-metadata/program/src/state.rs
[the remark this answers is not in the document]
danenbm: I do think maybe in general it is ok to have a `&self` reference even if you don't know if you need the info from an instance, because when you are just defining a trait you don't restrict the implementation details.  For example:

```
pub trait Summary {
    fn summarize(&self) -> String;
}

struct Something {
    summary: String
}

impl Summary for Something {
    fn summarize(&self) -> String {
        self.summary.clone()
    }
}

struct SomethingElse;

impl Summary for SomethingElse {
    fn summarize(&self) -> String {
        "This summary is hardcoded".to_string()
    }
}
```

I also think `&self` implies you are dealing with an instance and only really want it to be able to be called on an instance.  And I agree with your logic that you are **not** starting with an instance and it doesn't make sense to instantiate one, so I would say you are right.

Also, it looks like you removed the trait bound on the function in the trait (which makes sense to me also).

token-metadata/program/src/utils.rs
danenbm: Yeah that is fine, its probably not worth it.

token-metadata/program/src/state.rs
danenbm: Ok great!

danenbm (dismissed):

token-metadata/program/src/state.rs
```diff
@@ -82,6 +89,41 @@ pub const USE_AUTHORITY_RECORD_SIZE: usize = 18; //8 byte padding
...
+        }
+
+        let result: T = T::deserialize(&mut data)?;
+
+        Ok(result)
+    }
+
+    fn from_account_info<T: BorshDeserialize>(a: &AccountInfo) -> Result<T, ProgramError>
+where {
+        let ua: T = Self::safe_deserialize(&a.data.borrow_mut())
```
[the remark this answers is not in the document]
austbot: im dumb is below !

token-metadata/program/src/state.rs
```diff
@@ -433,10 +530,33 @@ pub struct Edition {
...
+        Key::EditionV1
+    }
+
+    fn size() -> usize {
+        MAX_EDITION_LEN
+    }
+}
+
 impl Edition {
     pub fn from_account_info(a: &AccountInfo) -> Result<Edition, ProgramError> {
```
austbot, line 554: why doesnt this just use the trait?

token-metadata/program/src/state.rs
```diff
@@ -600,11 +730,11 @@ impl ReservationList for ReservationListV2 {
 
 impl ReservationListV2 {
     pub fn from_account_info(a: &AccountInfo) -> Result<ReservationListV2, ProgramError> {
```
austbot, line 732: same here?

austbot (commented):
Just a question of why we have dupped from_account infor methods on Edition

token-metadata/program/src/state.rs
samuelvanderwaal: It should; good catch. Will fix.

token-metadata/program/src/state.rs
samuelvanderwaal: Will fix.

token-metadata/program/src/state.rs
```diff
@@ -191,16 +233,27 @@ pub struct UseAuthorityRecord {
...
+        Key::UseAuthorityRecord
+    }
+
+    fn size() -> usize {
+        USE_AUTHORITY_RECORD_SIZE
+    }
+}
+
+impl UseAuthorityRecord {
     pub fn from_bytes(b: &[u8]) -> Result<UseAuthorityRecord, ProgramError> {
```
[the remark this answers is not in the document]
samuelvanderwaal: It effectively performs the same thing as `safe_deserialize`.

token-metadata/program/src/state.rs
```diff
@@ -82,6 +89,41 @@ pub const USE_AUTHORITY_RECORD_SIZE: usize = 18; //8 byte padding
 
 pub const COLLECTION_AUTHORITY_RECORD_SIZE: usize = 11; //10 byte padding
 
+pub trait TokenMetadataAccount {
```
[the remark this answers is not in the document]
samuelvanderwaal: Indeed

token-metadata/js/src/generated/accounts/EditionMarker.ts
```diff
@@ -63,6 +64,18 @@ export class EditionMarker implements EditionMarkerArgs {
     return EditionMarker.fromAccountInfo(accountInfo, 0)[0];
   }
 
+  /**
+   * Provides a {@link web3.Connection.getProgramAccounts} config builder,
+   * to fetch accounts matching filters that can be specified via that builder.
+   *
+   * @param programId - the program that owns the accounts we are filtering
+   */
+  static gpaBuilder(
```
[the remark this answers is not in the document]
samuelvanderwaal: Thorsten at Work

samuelvanderwaal (commented):
version patch

austbot (approved):

PR #584, opened 2022-07-01
Stranzhay/pl 56 optional receipts need to be able to be closed 

### LINKED ISSUE #295: [Bug]: Optional Receipts Need to be able to be closed

auction-house

### Issue description

Receipt Accounts are Expensive and Auction House instruction needs to be made to allow the authority/maker of the receipt to close them to recoup their cost. 

Perhaps adding the bidder that paid for the accounts into the struct would allow them or the auction house authority to close them. 

### Relevant log output

_No response_

Medium (should be fixed soon)

stranzhay:
closes #295


auction-house/program/src/receipt/mod.rs
```diff
@@ -552,23 +563,29 @@ pub fn print_purchase_receipt<'info>(
...
-    bid_receipt.purchase_receipt = Some(purchase_receipt_account.key());
-
-    bid_receipt.try_serialize(&mut *bid_receipt_data)?;
+    let curr_bid_receipt_lamp = bid_receipt_info.lamports();
+    **bid_receipt_info.lamports.borrow_mut() = 0;
+    sol_memset(*bid_receipt_data, 0, BID_RECEIPT_SIZE);
+    **bookkeeper.lamports.borrow_mut() = bookkeeper
+        .lamports()
+        .checked_add(curr_bid_receipt_lamp)
+        .ok_or(AuctionHouseError::NumericalOverflow)?;
```
kespinola, line 589: Closing the listing and bid receipt when execute sale is fired messes up indexing of these accounts with geyser since when the account is passed to the geyser on account update the data will be empty. The indexer won't be able to correctly process the account. It has the address but the discriminator is cleared as well.

Close receipt should be a separate permissionless instruction that sends funds to the bookkeeper (as it is now) but only let the instruction run if canceled_at or purchase_receipt written to the bid receipt or listing receipt.

This would allow indexers to process the account change and then at a later time close our the accounts after they've been indexed properly. 

At Holaplex, we are speccing v2 of our indexer and will likely track address to accounts type so when data is empty can still lookup the table associated with the account by its address.

kespinola (changes requested):

auction-house/program/src/receipt/mod.rs
kespinola: Holaplex accounts for cancel offer, cancel listings, and execute sale instructions now so close away.

https://github.com/holaplex/indexer/tree/dev/crates/indexer/src/geyser/instructions

austbot (changes requested):
this is not backward compatible, also may not be needed

PR #591, opened 2022-07-05
fix(candy-machine): add missing `mut`

lincot:
- `SetCollection`: `authority` and `payer` need to be mutable to invoke [ApproveCollectionAuthority](https://docs.metaplex.com/programs/token-metadata/instructions#approve-a-new-collection-authority). also made `payer` to pay for account creation because it's more appropriate than `authority`
- `SetCollectionDuringMint`: `metadata` and `payer` need to be mutable to invoke [SetAndVerifyCollection](https://docs.metaplex.com/programs/token-metadata/instructions#set-and-verify-the-collection)
- `WithdrawFunds`: `authority` needs to be mutable to receive funds. also removing address check since there is an identical `has_one` above

### DIFF
--- a/candy-machine/js/idl/candy_machine.json
+++ b/candy-machine/js/idl/candy_machine.json
@@ -1,5 +1,5 @@
 {
-  "version": "4.0.0",
+  "version": "4.0.1",
   "name": "candy_machine",
   "instructions": [
     {
@@ -140,7 +140,7 @@
         },
         {
           "name": "authority",
-          "isMut": false,
+          "isMut": true,
           "isSigner": true
         },
         {
@@ -150,7 +150,7 @@
         },
         {
           "name": "payer",
-          "isMut": false,
+          "isMut": true,
           "isSigner": true
         },
         {
@@ -333,12 +333,12 @@
         },
         {
           "name": "metadata",
-          "isMut": false,
+          "isMut": true,
           "isSigner": false
         },
         {
           "name": "payer",
-          "isMut": false,
+          "isMut": true,
           "isSigner": true
         },
         {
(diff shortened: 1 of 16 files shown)


stegaBOB:
Can you regenerate the IDL and SDK please using shank?

lincot:
@stegaBOB generating IDL doesn't seem to change anything, how do I generate SDK?

it may be visual with shank, but it changes anchor's IDL and fixes CPI

stegaBOB:
How are candy machines currently working?

stegaBOB:
This feels like a breaking change, but we know for a fact that candy machines currently are able to function. Not sure how that is

lincot:
`RemoveCollection` also needs mutable `authority` due to [RevokeCollectionAuthority](https://docs.metaplex.com/programs/token-metadata/instructions#revoke-an-existing-collection-authority), even though I doubt that RevokeCollectionAuthority, as well as ApproveCollectionAuthority, really needs mutable update authority

lincot:
I guess `RemoveCollection` also needs mutable `authority` for [RevokeCollectionAuthority](https://docs.metaplex.com/programs/token-metadata/instructions#revoke-an-existing-collection-authority)

also I doubt that token-metadata's `ApproveCollectionAuthority` truly needs mutable update authority

stegaBOB:
The authority in set collection doesn't need to be marked mutable, but I believe you're right in that all the others do. The reason the candy machine program still works now is I believe is because anchor marks payer as mutable on the client side automatically, and that the set collection during mint can't be run in its own txn, so the mutable writable metadata in the mint ix takes precedence.

PR #594, opened 2022-07-06
Stranzhay/pl 90 move hydra into mpl

stranzhay:


jshiohaha:
oops, you can ignore the `Verify package SDK` check for now. the workflow is picking up `['.github', 'hydra']` as top-level programs when it should only pick up `['hydra']`. i will update!

stranzhay:
@austbot its under hydra -> program already and then hydra -> cli and hydra -> js

auction-house/program/sugar.log
```diff
@@ -0,0 +1,2 @@
+{"v":0,"name":"sugar","msg":"Lend me some sugar, I am your neighbor.","level":30,"hostname":"Sarahs-MacBook-Pro-2.local","pid":3347,"time":"2022-07-09T02:02:11.842556Z","target":"sugar","line":110,"file":"src/main.rs"}
```
austbot, line 1: whats going here? I think this is an accident

stranzhay: not sure how this got in here at all.

PR #604, opened 2022-07-11
Fix serde -> serde-feature in the code

omertxyz:


austbot (dismissed):

samuelvanderwaal:
@omertxyz what is the issue this PR fixes? Using v1.3.2 with feature = ["serde_feature"] seems to work fine now.

omertxyz:
@samuelvanderwaal does it actually work for ser/de or it just compiles? There was a mismatch between the feature name defined in cargo.toml and the one in the code, I think serde_feature probably doesn't do anything, am I wrong?

austbot (approved):
version patch

PR #605, opened 2022-07-12
Forcing verification of canonical PDAs and bumps.

blockiosaurus:


austbot (approved):

lincot:
there is no need in the bump arguments then.
we can just assign to anchor's bumps like this:
```rust
auction_house.treasury_bump = *ctx.bumps.get("auction_house_treasury").unwrap();
```

PR #607, opened 2022-07-13
closed the ListingConfig account

2001asjad:
### The Following tasks have been performed
1. Running the `ExecuteSales` successfully in the Auctioneer.
2. Closed the `ListingConfig` account by setting its lamports to 0.
3. Added that SOL to the seller's account.

### Current problem
Currently, in the added code, we are getting this error in the editor:
```
no method named `lamports` found for reference `&std::boxed::Box<anchor_lang::prelude::Account<'info, sell::config::ListingConfig>>` in the current scope

method not found in `&std::boxed::Box<anchor_lang::prelude::Account<'info, sell::config::ListingConfig>>`
```

This PR closed this [issue](https://linear.app/metaplex/issue/P-3/close-listingconfigs-on-auctioneer-sales)


auctioneer/program/src/execute_sale/mod.rs
```diff
@@ -221,5 +224,25 @@ pub fn auctioneer_execute_sale<'info>(
 
     invoke_signed(&ix, &cpi_accounts.to_account_infos(), &[&auctioneer_seeds])?;
 
+    let close_listing_config = listing_config.lamports();
```
blockiosaurus, line 227: So I took a look at this and the `listing_config` is an `Account` (https://docs.rs/anchor-lang/0.25.0/anchor_lang/accounts/account/struct.Account.html) which is just a wrapper around `AccountInfo`, which does have a `lamports` field. To access that `AccountInfo` you need to call `to_account_info()` to unwrap `listing_config`.

2001asjad:
Hey @blockiosaurus , The methods have been solved, the only glitch now is the third argument in the call:
```
let ix = anchor_lang::solana_program::system_instruction::transfer(
        &ctx.accounts.listing_config.key(),
        &ctx.accounts.seller.key(),
        lamports,
    );
```

2001asjad:
Hey @blockiosaurus , I have made the changes, what I was thinking that we don't need the portion:
```
 anchor_lang::solana_program::program::invoke(
        &ix,
        &[
            ctx.accounts.listing_config.to_account_info(),
            ctx.accounts.seller.to_account_info(),
        ],
    );
    
```
As the `transfer` function seems to be doing the job.

PR #619, opened 2022-07-20
fix: drop spl-associated-token-account to 1.0.3

samuelvanderwaal:
I missed this last dependency conflict on my other changes. [spl-associated-token-account 1.0.5](https://crates.io/crates/spl-associated-token-account/1.0.5/dependencies) depends on spl-token ^3.3.0 which causes conflicts when clients try to use the Rust package with other Solana dependencies which are pinned to spl-token 3.2. We need to stay on 3.2 until we bump everything to the 1.10 series.


austbot (approved):

samuelvanderwaal (commented):
version patch

PR #621, opened 2022-07-21
build: remove most filters in version workflow

jshiohaha:
We're giving power back to the people and allowing them to make version updates whenever they want!

If accepted, this change will do a few things: 

1. Consider all changed files. I previously filtered out files like `Cargo.toml` and `package.json`, but @samuelvanderwaal pointed out dependency changes might necessitate a version bump. Example workflow that failed: https://github.com/metaplex-foundation/metaplex-program-library/runs/7435095133
2. Add `contractVersion` to each packages' `package.json` and update that value after updating a rust package version. We read the version directly from the relative `Cargo.toml`.
3. Exit the workflow when no version changes are needed. Previously the workflow took an unnecessarily long time to run by installing Rust, Solana, and Anchor when they weren't needed.

### Testing
* Candy machine Cargo.toml PR: https://github.com/jshiohaha/metaplex-program-library/pull/102
 * Workflow that adds a version commit to PR: https://github.com/jshiohaha/metaplex-program-library/runs/7454524257?check_suite_focus=true
 * Resulting commit: https://github.com/jshiohaha/metaplex-program-library/pull/102/commits/96f199e9ccbf8f10c3d4a6ffc160eb5a424c32e3
* Workflow that exits early with no changes 🥳 https://github.com/jshiohaha/metaplex-program-library/actions/runs/2713509330


jshiohaha:
closing in favor of https://github.com/metaplex-foundation/metaplex-program-library/pull/704

PR #624, opened 2022-07-23
build: add workflow to integrate sugar cli test

jshiohaha:
Per Linear task, add [Sugar CLI test script](https://github.com/metaplex-foundation/sugar/blob/main/script/sugar-cli-test.sh) from [Sugar](https://github.com/metaplex-foundation/sugar) repository.

Right now, it's manually invoked and runs against candy machine deployed on devnet (default config). 

We need to decide what (if any) events should invoke the workflow, e.g. push to master with candy machine change. I think we could also convert it to an action and integrate with the existing candy machine program workflow.

### DIFF
--- a/.github/workflows/run-sugar-cli-test.yml
+++ b/.github/workflows/run-sugar-cli-test.yml
@@ -1,0 +1,48 @@
+name: Run Sugar CLI Test
+
+on:
+  workflow_dispatch:
+
+env:
+  SOLANA_VERSION: 1.9.22
+  RUST_TOOLCHAIN: stable
+
+jobs:
+  run-sugar-cli-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/install-linux-build-deps
+      - uses: ./.github/actions/install-solana
+        with:
+          solana_version: ${{ env.SOLANA_VERSION }}
+      - uses: ./.github/actions/install-rust
+        with:
+          toolchain: ${{ env.RUST_TOOLCHAIN }}
+
+      - run: solana-keygen new -o $GITHUB_WORKSPACE/kp --no-bip39-passphrase
+        shell: bash
+
+      - run: solana config set --url https://api.devnet.solana.com --keypair $GITHUB_WORKSPACE/kp
+        shell: bash
+
+      # note: must specify config before this command, or directly specify network. otherwise, airdrop might fail.
+      - run: solana airdrop 1 $(solana address -k $GITHUB_WORKSPACE/kp)
+        shell: bash
+
+      - run: solana balance $(solana address -k $GITHUB_WORKSPACE/kp)
+        shell: bash
+
+      # install the publicly available binary for the lifetime of the workflow
+      - name: Install Stable Sugar
+        run: bash <(curl -sSf https://sugar.metaplex.com/install.sh)
+        shell: bash
+
+      # we need to actually reference the test script to run it, so we'll clone the repo and launch the script from there
+      - name: Pull in test script from Sugar repository (outside of MPL)
+        shell: bash
+        run: cd .. && git clone https://github.com/metaplex-foundation/sugar.git && cd sugar
+
+      - name: Run test script from sugar directory
+        shell: bash
+        run: cd ../sugar && echo | ./script/sugar-cli-test.sh


.github/workflows/run-sugar-cli-test.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          solana_version: ${{ env.SOLANA_VERSION }}
+      - uses: ./.github/actions/install-rust
+        with:
+          toolchain: ${{ env.RUST_TOOLCHAIN }}
+
+      - run: solana-keygen new -o $GITHUB_WORKSPACE/kp --no-bip39-passphrase
+        shell: bash
+
+      - run: solana config set --url https://api.devnet.solana.com --keypair $GITHUB_WORKSPACE/kp
```
jshiohaha, line 26: should i swap this with a non-public endpoint?

austbot: this endpoint sucks alot we should make it a non public or genesys go one

austbot: Actually I think we should run a local validator if thats possible so we eliminate network issues as false test failures,

austbot (changes requested):

.github/workflows/run-sugar-cli-test.yml
jshiohaha: ok cool, that's what i was thinking too - will swap out for a genesys go endpoint (https://devnet.genesysgo.net)

jshiohaha: good point on the local validator - i'll ask someone more familiar with sugar if that's possible. but, at first glance, it looks like we only expect to run against devnet or mainnet https://github.com/metaplex-foundation/sugar/blob/main/script/sugar-cli-test.sh#L181-L194

if not, another option for to mitigate false negatives due to network conditions is this [retry action](https://github.com/marketplace/actions/retry-step).

jshiohaha:
newest version runs against a local test validator, using amman to load candy machine and token metadata programs. the current version of the workflow still loads programs from devnet - alternatively, we could build and reference local `.so` files for candy machine and token metadata, but that would greatly extend build time.

if accepted we need to add the following secrets for the workflow to work (configured on AWS console)
* SUGAR_CLI_TEST_AWS_ACCESS_KEY
* SUGAR_CLI_TEST_AWS_SECRET_KEY
* SUGAR_CLI_TEST_AWS_REGION

tested in my local fork: https://github.com/jshiohaha/metaplex-program-library/runs/7627238343

.github/.config/sugar-cli-test.amman.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const base = require('../../.base-ammanrc.js');
+
+const validator = {
+  ...base.validator,
+  programs: [],
+  accountsCluster: 'https://devnet.genesysgo.net',
+  accounts: [
```
thlorenz, line 9: To not break CI due to changes on devnet you may wanna include the program `.so` files instead.
That is unless you _do_ want to break if devnet becomes incompat with what sugar assumes.
But then maybe pulling from main-net is the better thing to do?

Not sure .. either way pulling it from either net prevents us from testing features that haven't been deployed yet.

.github/workflows/run-sugar-cli-test.yml
```diff
@@ -0,0 +1,69 @@
...
+      - run: solana config set --url http://127.0.0.1:8899 --keypair $GITHUB_WORKSPACE/kp
+        shell: bash
+
+      # `sugar-cli-test.amman.js` specifies how to load candy machine and token metadata bytes into the local validator
+      - name: Install and start amman with custom config
+        shell: bash
+        run: |
+          cd candy-machine/js
+          YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn && npm install --location=global @metaplex-foundation/amman
+          cat $GITHUB_WORKSPACE/.github/.config/sugar-cli-test.amman.js > .ammanrc.js
```
thlorenz, line 43: Why isn't this `.ammanrc.js` already in the right place? Isn't it needed to run tests locally?

.github/workflows/run-sugar-cli-test.yml
```diff
@@ -0,0 +1,69 @@
...
+          amman start > /dev/null &
```
thlorenz, line 44: Not sure if you wanna totally ignore amman logs since they might be useful to triage failures.
Why not `amman start > /tmp/amman-logs` and then after tests completed (or only when they failed) `echo "amman output"; cat /tmp/amman-logs`?

thlorenz: Also you'd need to do some `&2>1` magic (or similar .. always have to google that) to both capture stderr and stdout.

thlorenz (changes requested):

.github/workflows/run-sugar-cli-test.yml
jshiohaha: Good point - I thought that we needed to be within an existing directory with an `.ammanrc.js`, but it looks like the relative import for the base `ammanrc.js` works + we install amman globally. I will remove the cd + cat.

.github/workflows/run-sugar-cli-test.yml
jshiohaha: Good idea! I will add redirection for both stdout and stderr to some temp log file. After some googling, I think it's smth like

```
amman start > <tmp-dir>/amman-logs 2>&1
```

.github/.config/sugar-cli-test.amman.js
jshiohaha: Yeah, I think this is a more philosophical question w.r.t. what we want to test against. I can see an argument for all three. If testing against local changes, we will need to produce a `.so` file for candy machine and token metadata unless those `.so` files were previously cached somewhere, like AWS.

Maybe @austbot can weigh in here.

jshiohaha: Another path is to provide the option to choose build and use local, use devnet, or use mainnet?

austbot (approved):

PR #632, opened 2022-07-28
closed the listing config account

2001asjad:
## What needs to be done
ListingConfigs are no longer needed once the listing has been closed. The ListingConfig account should therefore be closed when ExecuteSale runs successfully.

## Workflow
1.) Run the ExecuteSale successfully.
2.) Once they run successfully the ListingConfig account is closed.
3.) Its lamports are retuned to the seller account.
4.) At last the tests to check if the LisitnConfig account is closed also added.
## Test Results
Screenshot 2022-08-05 at 4 05 35 PM

Solves the issue : [P-3](https://linear.app/metaplex/issue/P-3/close-listingconfigs-on-auctioneer-sales).

### DIFF
--- a/auctioneer/program/tests/execute_sale.rs
+++ b/auctioneer/program/tests/execute_sale.rs
@@ -369,6 +369,14 @@ async fn execute_sale_success() {
     let rent_exempt_min: u64 = rent.minimum_balance(listing_config_account.data.len());
 
     assert_eq!(seller_before.lamports + fee_minus + rent_exempt_min, seller_after.lamports);
+
+    let listing_config_closed = context
+          .banks_client
+          .get_account(listing_config_address)
+          .await
+          .unwrap();
+    
+    assert_eq!(listing_config_closed.is_none(), true);
 }
 
 #[tokio::test]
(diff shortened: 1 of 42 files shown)


auctioneer/program/tests/execute_sale.rs
```diff
@@ -536,6 +548,13 @@ async fn execute_sale_two_bids_success() {
         .unwrap();
     assert_eq!(buyer1_token_before.is_none(), true);
 
+    let listing_config_account = context
+          .banks_client
+          .get_account(listing_config_address)
+          .await
+          .unwrap()
```
blockiosaurus, line 555: Awesome! So last request. We know that if this block is called after the account is closed (Line 558) that the first unwrap returns an Option that is None. So the last check we need is something at the very end that retrieves this account again, unwraps it once, and checks and verifies that Option == None and the account is officially closed.

2001asjad: Done!

blockiosaurus (dismissed):

danenbm (approved):

PR #641, opened 2022-08-03
[token-metadata] H4-fix

samuelvanderwaal:


stranzhay (approved):

austbot (approved):
This looks good to me, 

my only question is a logic one should we allow them to set FungibleAsset if the mint has a higher than one max supply. I dont think that answer should be scoped to this PR

samuelvanderwaal:
> This looks good to me,
> 
> my only question is a logic one should we allow them to set FungibleAsset if the mint has a higher than one max supply. I dont think that answer should be scoped to this PR

I don't follow?

I know that Phantom has asked about being able to set Fungible Assets that have decimals greater than zero but I'm not sure how else to objectively differentiate on-chain between these two types. Maybe we let people choose whether it's a FungibleAsset or a FungibleToken and only determine the other types automatically?

PR #643, opened 2022-08-03
fix(token-metadata): only allow update if existing collection has bee…

stranzhay:
Changing to make sure that collection cannot be updated if it is verified. The user will have to unverify which will force the collection size to decrement. Use case here

Given:
 - A Collection NFT A (size: 1)
 - A Collection NFT B (size: 0)
 - A Regular NFT from Collection NFT A and verified
When:
 - We update the collection of Regular NFT by passing Collection NFT B to the UpdateMetadataAccountV2 instruction
Then:
 - The Regular NFT is now from Collection NFT B and not verified
 - We expect:
    - Collection NFT A (size: 0) -> Since it's no longer from that collection
    - Collection NFT B (size: 0) -> Since we did not verify it yet
 - We get:
    - Collection NFT A (size: 1) -> The update instruction did not decrement the size of Collection NFT A
    - Collection NFT B (size: 0) -> All good here

### DIFF
--- a/token-metadata/program/src/assertions/collection.rs
+++ b/token-metadata/program/src/assertions/collection.rs
@@ -19,6 +19,7 @@ pub fn assert_collection_update_is_valid(
     // If incoming verified is true. Confirm incoming and existing are identical
     let is_incoming_data_valid = !is_incoming_verified_true
         || (existing.is_some()
+            && !existing.as_ref().unwrap().verified
             && incoming.as_ref().unwrap().verified == existing.as_ref().unwrap().verified
             && incoming.as_ref().unwrap().key == existing.as_ref().unwrap().key);
 
--- a/token-metadata/program/src/assertions/collection.rs
+++ b/token-metadata/program/src/assertions/collection.rs
@@ -11,11 +11,18 @@ use crate::{
 
 pub fn assert_collection_update_is_valid(
     edition: bool,
+    existing: &Option<Collection>,
     incoming: &Option<Collection>,
 ) -> Result<(), ProgramError> {
     let is_incoming_verified_true = incoming.is_some() && incoming.as_ref().unwrap().verified;
 
-    if is_incoming_verified_true && !edition {
+    // If incoming verified is true. Confirm incoming and existing are identical
+    let is_incoming_data_valid = !is_incoming_verified_true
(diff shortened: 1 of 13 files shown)


samuelvanderwaal:
This looks good. Can you update the description with a couple sentences explaining what changes you made and why? That will help the second reviewer get up-to-speed faster.

token-metadata/program/src/assertions/collection.rs
```diff
@@ -11,18 +11,11 @@ use crate::{
 
 pub fn assert_collection_update_is_valid(
     edition: bool,
-    existing: &Option<Collection>,
     incoming: &Option<Collection>,
 ) -> Result<(), ProgramError> {
     let is_incoming_verified_true = incoming.is_some() && incoming.as_ref().unwrap().verified;
 
-    // If incoming verified is true. Confirm incoming and existing are identical
-    let is_incoming_data_valid = !is_incoming_verified_true
```
austbot, line 20: why was this removed

token-metadata/program/src/processor.rs
```diff
@@ -348,7 +348,7 @@ pub fn process_update_metadata_accounts_v2(
                 true,
             )?;
             metadata.data = compatible_data;
-            assert_collection_update_is_valid(false, &metadata.collection, &data.collection)?;
+            assert_collection_update_is_valid(false, &data.collection)?;
```
austbot, line 351: wont this fail in a lot of cases where we expecte them to send the same collection that exists in the update?

I think we need to only fail in ```token-metadata/program/src/assertions/collection.rs``` if the id is different then what was existing

austbot (changes requested):
I dont think this will work and in many cases will break users.

token-metadata/program/src/processor.rs
samuelvanderwaal: They don't need to send the same collection if they're not changing anything, they just send `None` for the `Option`.

samuelvanderwaal: Oh yeah, that doesn't work this way currently, but maybe that's how it should be?

austbot (dismissed):

austbot (approved):

PR #644, opened 2022-08-03
fix(token-metadata): [H3] adding signer check after pda check

stranzhay:


austbot (approved):
This looks good, I even double checked the assertions I wrote a long time ago.

token-metadata/js/src/generated/instructions/RevokeCollectionAuthority.ts
```diff
@@ -62,7 +62,7 @@ export function createRevokeCollectionAuthorityInstruction(
     {
       pubkey: accounts.delegateAuthority,
       isWritable: true,
-      isSigner: true,
+      isSigner: false,
```
austbot, line 65: this is concerning.

austbot: this should have already been like this for a long time.

austbot (approved):
Approved but we need to get to the bottom of why solita is only now changing the signer.

token-metadata/js/src/generated/instructions/RevokeCollectionAuthority.ts
samuelvanderwaal: This is resolved via a Google Meet. It's my fault for merging a PR without updating the JS SDK but we'll be doing that after we get all these fixes in.

samuelvanderwaal (approved):

austbot (approved):

PR #646, opened 2022-08-04
Bubblegum support and refactor `assert_data_valid` in `token-metadata`

danenbm:
### Notes
- Remove requirement that one of the creators is the update authority
when metadata account is being created.
- Flatten some of the logic and use positive conditionals that
should be easier to understand.
- Use an iterator and hashmap instead of nested for loop to look for
duplicates.
- Allow Bubblegum program to create metadata with Verified Creators
 (feature currently hardcoded to inactive).

### Testing
- Added test to allow update authority to not be creato.
- Still working on testing the Bubblegum CPI (I don't think I can do that with the Banks client so looking to test it in Candyland repo).

### DIFF
--- a/token-metadata/program/src/utils.rs
+++ b/token-metadata/program/src/utils.rs
@@ -4,8 +4,8 @@ use crate::{
     error::MetadataError,
     pda::find_master_edition_account,
     state::{
-        get_reservation_list, CollectionDetails, Data, DataV2, Edition, EditionMarker, Key,
-        MasterEditionV1, MasterEditionV2, Metadata, TokenMetadataAccount, TokenStandard, Uses,
+        get_reservation_list, CollectionDetails, Creator, Data, DataV2, Edition, EditionMarker,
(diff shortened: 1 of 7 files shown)


token-metadata/program/src/utils.rs
```diff
@@ -55,79 +55,81 @@ pub fn assert_data_valid(
...
-                        .ok_or(MetadataError::NumericalOverflowError)?;
+        if creators.is_empty() {
+            return Err(MetadataError::CreatorsMustBeAtleastOne.into());
+        }
 
-                    if creator.address == *update_authority {
-                        found = true;
-                    }
+        // Do not allow duplicate creators.
+        let u: Vec<&Creator> = creators.into_iter().unique().collect();
```
danenbm, line 69: This check won't work as-is because it would allow for creators with different struct member fields to count as unique entries when it really needs to be comparing on `creator.address`.

danenbm:
Reworking based on some other project needs.  Closing this PR.

PR #651, opened 2022-08-04
Master edition account PDA not validated

stranzhay:


token-metadata/program/src/assertions/collection.rs
```diff
@@ -85,6 +86,18 @@ pub fn assert_collection_verify_is_valid(
             return Err(MetadataError::CollectionNotFound.into());
         }
     }
+
+    assert_derivation(
+        &crate::id(),
+        collection_mint,
```
austbot, line 92: shouldnt this be edition_account_info pda check ?

austbot (changes requested):

token-metadata/program/src/assertions/collection.rs
stranzhay: according to the doc those seeds are for edition_acount_info   ["metadata", program_id, mint, "edition"]

stranzhay: oh i see what you're saying.

token-metadata/program/tests/verify_collection.rs
```diff
@@ -1275,3 +1275,168 @@ mod verify_collection {
...
+            None,
+        )
+        .await
+        .unwrap_err();
+
+    assert_custom_error!(err, MetadataError::IncorrectOwner);
+}
+
+#[tokio::test]
+async fn fail_verify_collection_derived_key_invalid() {
```
austbot, line 1354: test name should include information that its the edition key that is invalid

austbot (commented):
This seems correct to me but your api changes for solita are out of date some how

austbot (approved):

PR #656, opened 2022-08-05
[M-1]: check for already verified/unverified items

samuelvanderwaal:


token-metadata/program/src/error.rs
```diff
@@ -448,6 +448,14 @@ pub enum MetadataError {
     /// 113 - Edition Account Doesnt Match Collection
     #[error("Edition account aoesnt match collection ")]
     CollectionMasterEditionAccountInvalid,
+
+    /// 114 - Item is already verified.
+    #[error("Item is already verified.")]
+    AlreadyVerified,
+
+    /// 115 - Item is already verified.
```
austbot, line 456: copy pasta comment error

token-metadata/program/tests/verify_sized_collection_item.rs
```diff
@@ -1284,3 +1284,336 @@ mod verify_sized_collection_item {
...
+            CollectionDetails::V1 { size } => size,
+        }
+    } else {
+        panic!("Expected CollectionDetails::V1");
+    };
+    assert_eq!(size, 1);
+}
+
+#[tokio::test]
+async fn fail_unverify_already_unverified() {
```
austbot, line 1414: I think this test would be better if the collection had a size, to truly test this, I think this test is fine but if you want to go above and beyond have a collection size of 3 then unverified 1 and ensure there are size 2 after an extra failing try to inverify

austbot (approved):
This seems good to me, the tests look good aswell

token-metadata/program/src/processor.rs
```diff
@@ -1520,6 +1534,13 @@ pub fn set_and_verify_sized_collection_item(
     let mut metadata: Metadata = Metadata::from_account_info(metadata_info)?;
     let mut collection_metadata: Metadata = Metadata::from_account_info(collection_info)?;
 
+    // Don't reverify already verified items, otherwise we end up with invalid size data.
+    if let Some(ref collection) = metadata.collection {
```
danenbm, line 1538: I think in this case it would be equivalent but I usually would do:
`if let Some(collection) = &metadata.collection` instead.

Do you have a reason to use `ref` keyword?
I don't know everything about this, but curious if you do know more about it.  Like is `ref` preferred.  I was thinking not needed because of this:
https://rust-lang.github.io/rfcs/2005-match-ergonomics.html

token-metadata/program/tests/verify_sized_collection_item.rs
danenbm: This might be one portion of what Austin is already saying, but I would like to add to this comment specifically:

I looked briefly through the existing `verify_sized_collection_item` tests, and I am seeing all the good checks to be defensive against missing or incorrect data or signers, etc.  and I'm seeing success case with collection size of 1.

But to me a piece that we should specifically test is that `CollectionDetailsV1::size` is working properly, i.e., incrementing and decrementing in general, including even decrementing past 0 case (probably not incrementing past `u64::max` unless you can hardcode it somehow).

I could just be missing where these size checks are occurring.

danenbm (dismissed):
This change looks good to me and this testing covers this PR well.  I do think there are some other aspects of sized collections for which I think we should consider expanding testing, unless I am missing where it is currently being done.

token-metadata/program/tests/verify_sized_collection_item.rs
samuelvanderwaal: Done.

token-metadata/program/src/error.rs
samuelvanderwaal: Good catch. This was actually copilot, IIRC. Modern solutions, same problems.

samuelvanderwaal:
> This change looks good to me and this testing covers this PR well. I do think there are some other aspects of sized collections for which I think we should consider expanding testing, unless I am missing where it is currently being done.

Check out the `collection_sizes.rs` test and the `size_tracking` module in it. I think that has what you are looking for but probably wouldn't hurt to have some more tests in it.

token-metadata/program/tests/verify_sized_collection_item.rs
danenbm: Talked offline with @samuelvanderwaal and he showed me that the `size_tracking` module in `collection_sizes.rs` is what I was looking for.  Comment resolved.

token-metadata/program/src/processor.rs
samuelvanderwaal: In this case it's not actually necessary. I'm surprised `clippy` didn't catch it. For the other two functions thre needs to be a reference on one side or the other and it doesn't really matter but I suppose using `&` instead of `ref` is more idiomatic so I fixed that.

danenbm (approved):
Looks great!

PR #658, opened 2022-08-08
Stranzhay review

stranzhay:


jshiohaha:
Linking here for visibility - given that these changes will bump anchor and solana versions, here are my proposed changes so that CI works with the version bumps https://github.com/metaplex-foundation/metaplex-program-library/pull/668

PR #663, opened 2022-08-09
Some changes to get Bubblegum to build

danenbm:
### Notes
- `anchor build --skip-lint` builds from root of
metaplex-program-library and  builds bubblegum.so.
- I am not sure if this is the correct setup for some
of these project files and need an in-depth review,
especially for some of the js/solita stuff.


bubblegum/program/Cargo.toml
```diff
@@ -21,7 +21,7 @@ anchor-spl = { version = "0.25.0" }
 spl-token = { version = "3.3.0", features = [ "no-entrypoint" ] }
 spl-associated-token-account = { version = "1.0.5", features = [ "no-entrypoint" ] }
 mpl-token-metadata = { version = "1.3.3", features = ["no-entrypoint"] }
-gummyroll = { path = "../gummyroll", features = ["cpi"] }
+gummyroll = { git = "https://github.com/jarry-xiao/candyland.git", features = ["cpi"] }
```
austbot, line 24: youi may need a path here

austbot (approved):

bubblegum/program/Cargo.toml
danenbm: It actually does `anchor build` already build without a path.  I tried adding a path just for posterity, but it says:
```
Only one of `git` or `path` is allowed.
```

PR #599, opened 2022-07-08
upgrade to anchor 0.25, solana 1.10.25 and spl-token 3.3

dovahcrow:

PR #668, opened 2022-08-09
build: pull out get solana version logic into an action and reuse across workflows

jshiohaha:
### Context
There is an [outstanding PR](https://github.com/metaplex-foundation/metaplex-program-library/pull/599) to upgrade Anchor to 0.25.x and Solana to 1.10.x. To avoid breaking the CI, we need to also upgrade the Solana version used across our workflows. Previously, it was hard-coded but is now dynamic based on network versions. 

### Changes
* Remove all hard-coded solana version environment variables 🤗
* Pull out @blockiosaurus's dynamic getVersion logic from AH / Auctioneer workflows to be re-used across all program workflows. The benefit of this approach is that we can continuously test against versions on the specific networks - as opposed to hard-coded values 👌🏼
* The get version action will only return unique values so that we don't run workflows over the same versions
  * Right now, stable = mainnet, unstable = devnet?
* In `.github/workflows/pr-comment.yml`, extract should execute job logic to its own job so we can re-use the output var and don't have repeated multi-boolean statements 
* Update `.github/workflows/publish-on-pr-merge.yml` so that it doesn't fail if there is no PR attached to the merged commit (e.g. direct push)
* Update `.github/workflows/verify-lib-on-pr-open.yml` so that it only uses the first version to generate the package's library

### Todo 🔨
* ~Debug `Program Token Metadata` failure: https://github.com/jshiohaha/metaplex-program-library/runs/7739820555?check_suite_focus=true~ retry worked, likely a transient failure
* Debug `Integration Token Vault` failure: https://github.com/jshiohaha/metaplex-program-library/runs/7739821465?check_suite_focus=true
  * The workflow doesn't seem to work with the latest Solana versions. Everything builds as expected, but 4 tests fail. Since this program is on the path to deprecation ☠️, I'm going to mostly leave the workflow as is - just upgrade to the highest version I could with it still passing.
  * Note; this needs the changes in https://github.com/metaplex-foundation/metaplex-program-library/pull/599 to pass

### DIFF
--- a/.github/actions/get-solana-versions/action.yaml
+++ b/.github/actions/get-solana-versions/action.yaml
@@ -1,0 +1,24 @@
+name: 'Get Solana network versions'
+description: 'Returns unique solana versions (mainnet-beta, devnet)'
+
+outputs:
+  value:
+    description: Solana devnet and mainnet versions
+    value: ${{ steps.fetch-solana-versions.outputs.value }}
+
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/checkout@v3
+    - name: Invoke getVersions to get current Solana network versions
+      id: fetch-solana-versions
+      shell: bash
+      run: |
+        sudo apt-get install -y jq && \
+        MAINNET=$(curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":1, "method":"getVersion"}' | jq '.result["solana-core"]') && \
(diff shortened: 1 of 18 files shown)


joshsc63:
This would cause an issue with our current workflows as the latest version of solana introduces changes that cause failures. Instead of automatically pulling the version from dev/main net, we may want to have a central version defined that can be easily changed

blockiosaurus:
> This would cause an issue with our current workflows as the latest version of solana introduces changes that cause failures. Instead of automatically pulling the version from dev/main net, we may want to have a central version defined that can be easily changed

But the real world use case is us running on the actual cluster version. If there's a breakage then it's something we should be addressing in the protocol code.

PR #669, opened 2022-08-09
Feat: Candy Machine Freeze

stegaBOB:
This feature allows freezing NFTs after mint!

This PR adds 4 handlers and modifies an existing handler, withdraw_funds (it needs to be modified regardless).

`set_freeze`: This handler sets the freeze conditions on a Candy Machine. It can only be called prior to any NFTs minting out.

`remove_freeze`: This handler disables freeze. 

`unlock_funds`: This handler sends the locked up funds from the freeze to the authority or a provided token account if CM `token_mint` is enabled. It only works if all NFTs have been unfrozen already. I think it also closes the freeze PDA. 

`thaw_nft`: This is a permissionless handler that allows anyone to thaw their NFT (or anyone elses from a frozen Candy Machine) if it is eligible to be unfrozen. That eligibility is determined by the following conditions:
  - The set `freeze_time` or 7 days, whichever comes sooner, after the first NFT has been minted has elapsed.
  - The freeze is disabled by the authority of the Candy Machine.
  - The Candy Machine has been minted out.

This PR replaces #505


candy-machine/program/src/processor/mint.rs
```diff
@@ -571,6 +579,7 @@ pub fn handle_mint_nft<'info>(
         ctx.accounts.rent.to_account_info(),
         candy_machine_creator.to_account_info(),
     ];
+
```
austbot, line 582: Where do you enforce that the mint has freeze authority set correctly ? 
Also I think Create Master Edition forces mint supply of 1 but do you want to short circuit that to save compute ?

candy-machine/program/src/processor/mint.rs
```diff
@@ -634,6 +643,87 @@ pub fn handle_mint_nft<'info>(
...
+        // If we add more extra accounts later on we need to uncomment the following line out.
+        // remaining_accounts_counter += 1;
+        let mut freeze_pda: Account<FreezePDA> = Account::try_from(freeze_pda_info)?;
+
+        assert_is_ata(nft_token_account_info, &payer.key(), &mint_pubkey)?;
+        let seeds: &[&[u8]] = &[FreezePDA::PREFIX.as_bytes(), candy_pubkey.as_ref()];
+        let (expected_freeze_key, freeze_bump) = Pubkey::find_program_address(seeds, &crate::id());
+        assert_keys_equal(&expected_freeze_key, &freeze_pda.key())?;
+        // redundant check
+        freeze_pda.assert_from_candy(&candy_pubkey)?;
```
austbot, line 662: whats the deal here?

candy-machine/program/src/processor/update.rs
```diff
@@ -51,17 +63,18 @@ pub fn handle_update_candy_machine(
     }
 
     let old_uuid = candy_machine.data.uuid.clone();
-    candy_machine.wallet = ctx.accounts.wallet.key();
+    if is_feature_active(&old_uuid, FREEZE_FEATURE_INDEX) && candy_machine.token_mint != token_mint
```
austbot, line 66: good catch, but also do you think we should restrict changing the wallet? I do

candy-machine/program/src/processor/withdraw.rs
```diff
@@ -1,45 +1,42 @@
...
-    let snapshot: u64 = pay.lamports();
-
-    **pay.lamports.borrow_mut() = 0;
-
-    **authority.lamports.borrow_mut() = authority
-        .lamports()
-        .checked_add(snapshot)
-        .ok_or(CandyError::NumericalOverflowError)?;
+    let candy_machine = &ctx.accounts.candy_machine;
+    if is_feature_active(&candy_machine.data.uuid, FREEZE_FEATURE_INDEX) {
```
austbot, line 23: I think you should stull be able to withdraw only after the freeze is up

austbot (changes requested):
1. Why a 1 week limit
2. We need to be able to close the freeze pda
3. Withdraw should be allowed but only after freeze is up, or mints are all unfrozen.
I think funds should be locked until all assets are frozen because then it incentives the creators to unfreewze for people thereby not making us need to build a unfreezzer service people need to rely on

candy-machine/program/src/processor/initialize.rs
```diff
@@ -42,6 +42,8 @@ pub fn handle_initialize_candy_machine(
         items_redeemed: 0,
     };
 
+    candy_machine.data.uuid = "000000".to_string();
```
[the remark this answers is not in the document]
stegaBOB: Yep. We should be initializing to no features active to prevent people from "activating" a feature. Since the UUID is being passed in by the user.

candy-machine/program/src/processor/freeze/thaw_nft.rs
```diff
@@ -0,0 +1,107 @@
...
+
+pub fn handle_thaw_nft(ctx: Context<ThawNFT>) -> Result<()> {
+    let freeze_pda = &mut ctx.accounts.freeze_pda;
+    let candy_machine = &mut ctx.accounts.candy_machine;
+    let current_timestamp = Clock::get()?.unix_timestamp;
+    let can_thaw = if candy_machine.data_is_empty() {
+        true
+    } else {
+        let candy_struct: Account<CandyMachine> =
+            Account::try_from(&candy_machine.to_account_info())?;
```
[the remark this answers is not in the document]
stegaBOB: Also the account may not exist since withdraw could have happened :)

candy-machine/program/src/processor/freeze/thaw_nft.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[account(address = mpl_token_metadata::id())]
+    token_metadata_program: UncheckedAccount<'info>,
+    system_program: Program<'info, System>,
+}
+
+pub fn handle_thaw_nft(ctx: Context<ThawNFT>) -> Result<()> {
+    let freeze_pda = &mut ctx.accounts.freeze_pda;
+    let candy_machine = &mut ctx.accounts.candy_machine;
+    let current_timestamp = Clock::get()?.unix_timestamp;
+    let can_thaw = if candy_machine.data_is_empty() {
```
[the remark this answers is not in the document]
stegaBOB: Only after everything is thawed. We could add a check that closes at the end I guess, but I'm slightly weary of that failing for whatever reason.

candy-machine/program/src/processor/freeze/set_freeze.rs
```diff
@@ -0,0 +1,43 @@
...
+    // > Only needed if spl token mint is enabled
+    // freeze_ata
+}
+
+pub fn handle_set_freeze(ctx: Context<SetFreeze>, freeze_time: i64) -> Result<()> {
+    let candy_machine = &mut ctx.accounts.candy_machine;
+    candy_machine.assert_not_minted(CandyError::NoChangingFreezeDuringMint)?;
+    let freeze_pda = &mut ctx.accounts.freeze_pda;
+    if freeze_time > MAX_FREEZE_TIME {
+        return err!(CandyError::MaxFreezeIsOneWeek);
```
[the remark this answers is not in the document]
stegaBOB: Sorta needs to be a balance between being fair to the creator and the collector. Some NFT mints will just never mint out.

candy-machine/program/src/processor/withdraw.rs
stegaBOB: The issue is there's no way to do this without forcing the freezepda to be passed in. I think this tradeoff is worth it, and really only requires passing in the remove freeze ix before this.

candy-machine/program/src/processor/update.rs
stegaBOB: It shouldn't actually need to, since a non-withdrawn candy machine account will be passed in to the unfreeze ix. Initially I duplicated the wallet in the freezePDA but I dont think thats needed anymore

candy-machine/program/src/processor/mint.rs
stegaBOB: Just a redundant check I guess. Not strictly needed, but I had a helper function lying around that I wanted to use lol. We should be able to remove this if we want.

candy-machine/program/src/processor/mint.rs
```diff
@@ -467,23 +475,24 @@ pub fn handle_mint_nft<'info>(
...
-        punish_bots(
-            CandyError::CandyMachineEmpty,
-            payer.to_account_info(),
-            ctx.accounts.candy_machine.to_account_info(),
-            ctx.accounts.system_program.to_account_info(),
-            BOT_FEE,
-        )?;
-        return Ok(());
-    }
+    let wallet_to_use = if is_feature_active(&candy_machine.data.uuid, FREEZE_FEATURE_INDEX) {
```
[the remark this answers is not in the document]
stegaBOB: This isnt really a check here. Its just setting the treasury wallet to make sure its sent to the lockup if freeze is active.

candy-machine/program/src/processor/mint.rs
```diff
@@ -20,15 +21,17 @@ use solana_program::{
...
         CONFIG_LINE_SIZE, CUPCAKE_ID, EXPIRE_OFFSET, GUMDROP_ID, PREFIX,
     },
     utils::*,
-    CandyError, CandyMachine, CandyMachineData, ConfigLine, EndSettingType, WhitelistMintMode,
-    WhitelistMintSettings,
+    CandyError, CandyMachine, CandyMachineData, ConfigLine, EndSettingType, FreezePDA,
+    WhitelistMintMode, WhitelistMintSettings,
 };
 
 /// Mint a new NFT pseudo-randomly from the config array.
```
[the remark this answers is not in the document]
stegaBOB: ah the FreezePDA? good catch

candy-machine/program/src/processor/mint.rs
```diff
@@ -87,6 +90,10 @@ pub struct MintNFT<'info> {
     // > Only needed if candy machine has token mint
     // token_account_info
     // transfer_authority_info
+    // > Only needed if freeze token until after mint is set to true
+    // freeze_pda
```
stegaBOB, line 94:
```suggestion
    // freeze_pda (writable)
```

candy-machine/program/src/processor/freeze/thaw_nft.rs
```diff
@@ -0,0 +1,107 @@
...
+                    Transfer {
+                        from: freeze_pda.to_account_info(),
+                        to: payer.to_account_info(),
+                    },
+                ),
+                freeze_pda.freeze_fee,
+            )?;
+        }
+    } else {
+        msg!("Token account is not frozen!");
```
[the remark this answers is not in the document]
stegaBOB: We haven't been doing bot taxes on non mint commands. I think what I was doing was allowing this handler to always be run from both the owner and a random other party. The issue is if a random party runs it they cant revoke delegate authority (only owner can do this). So you can run it again and it would basically just revoke.

candy-machine/program/src/processor/freeze/remove_freeze.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+pub fn handle_remove_freeze<'info>(
+    ctx: Context<'_, '_, '_, 'info, RemoveFreeze<'info>>,
+) -> Result<()> {
+    let candy_machine = &mut ctx.accounts.candy_machine;
+    let freeze_pda = &mut ctx.accounts.freeze_pda;
+    let authority = &mut ctx.accounts.authority;
+    if let Some(mint) = &candy_machine.token_mint {
+        if ctx.remaining_accounts.len() != 3 {
```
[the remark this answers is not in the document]
stegaBOB: This should actually be compatible yep. Since they're all either required or not required, if we were to change this and someone didn't pass in the accounts from the older client the program would just treat it as None but if it would be passed in it would work the same as remaining accounts.

candy-machine/program/src/processor/collection/remove_collection.rs
```diff
@@ -30,9 +30,8 @@ pub struct RemoveCollection<'info> {
 pub fn handle_remove_collection(ctx: Context<RemoveCollection>) -> Result<()> {
     let mint = ctx.accounts.mint.to_account_info();
     let candy_machine = &mut ctx.accounts.candy_machine;
-    if candy_machine.items_redeemed > 0 {
-        return err!(CandyError::NoChangingCollectionDuringMint);
-    }
+    candy_machine.assert_not_minted(CandyError::NoChangingCollectionDuringMint)?;
```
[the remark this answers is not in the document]
stegaBOB: ty! I actually think we should change this function to take an actual AnchorError. That should allow it to correctly display line numbers from there.

candy-machine/program/src/processor/mint.rs
stegaBOB: Hm we'd have to do some weird logic way earlier in the transaction to catch this I think. I dont believe its safe to bot tax here. As of now this should still be safe since the freeze should fail later on, but good point yeah.

stegaBOB:
> 3\. Withdraw should be allowed but only after freeze is up, or mints are all unfrozen.
>         I think funds should be locked until all assets are frozen because then it incentives the creators to unfreewze for people thereby not making us need to build a unfreezzer service people need to rely on

What do you mean by freeze is up?

candy-machine/program/src/processor/mint.rs
stegaBOB: Another thing though. Not actually sure it makes sense to check this. You would have to be maliciously sending in instructions to make this the case. I'm under the impression that bot tax should be used to prevent bots trying to mint from killing the network, and not to prevent a malicious actor from just bringing it down.

stegaBOB: If youre just asking if freeze is set in general, the freeze ix from token metadata should just fail later on if it doesnt have freeze authority.

candy-machine/program/src/processor/freeze/remove_freeze.rs
```diff
@@ -0,0 +1,35 @@
...
+    freeze_pda: Account<'info, FreezePDA>,
+}
+
+pub fn handle_remove_freeze(ctx: Context<RemoveFreeze>) -> Result<()> {
+    let candy_machine = &mut ctx.accounts.candy_machine;
+    let freeze_pda = &mut ctx.accounts.freeze_pda;
+    freeze_pda.allow_thaw = true;
+    remove_feature_flag(&mut candy_machine.data.uuid, FREEZE_FEATURE_INDEX);
+
+    // Closes the account to enable editing if minting hasn't started
```
austbot, line 26: shouldnt we also close if everything is unfrozen

candy-machine/program/src/processor/freeze/unlock_funds.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+pub fn handle_unlock_funds<'info>(
+    ctx: Context<'_, '_, '_, 'info, UnlockFunds<'info>>,
+) -> Result<()> {
+    let candy_machine = &mut ctx.accounts.candy_machine;
+    let freeze_pda = &mut ctx.accounts.freeze_pda;
+    let authority = &mut ctx.accounts.authority;
+    if !freeze_pda.allow_thaw {
+        freeze_pda.allow_thaw = true;
```
austbot, line 34: what does allow thaw do?

austbot (approved):
This is good. All my concerns are taken care of

candy-machine/program/src/processor/freeze/unlock_funds.rs
stegaBOB: It flips whether or not you can thaw the nft if none of the other conditions are met. Just another way for the creator to let people thaw. They can call `remove_freeze` and then they can begin to thaw everything.

candy-machine/program/src/processor/freeze/remove_freeze.rs
stegaBOB: Thaw is done in the `unlock_funds` handler.

samuelvanderwaal (approved):

PR #675, opened 2022-08-11
chore(candy-machine): bump mpl-token-metadata version and fix new cli…

samuelvanderwaal:
To support sized collections in Sugar, we need to rely on `mpl-token-metadata` 1.3+ which requires `mpl-candy-machine` to bump its version of that crate and import the `TokenMetadataAccount` trait.

This PR also fixes the new clippy lint warnings introduced in Rust v1.63.0.

This should result in a minor bump in the`mpl-candy-machine` crate version, IMO, to v4.1.0.


samuelvanderwaal (commented):
version minor

stranzhay (approved):

stegaBOB:
Why did so much change?

candy-machine/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metaplex-foundation/mpl-candy-machine",
-  "version": "4.3.0",
+  "version": "4.4.0",
   "programVersion": "4.0.2",
```
stegaBOB, line 4: Should this be 4.1?

stegaBOB (changes requested):

candy-machine/js/package.json
samuelvanderwaal: No, the JS SDKs are versioned independently.

samuelvanderwaal:
> Why did so much change?

About half of the Rust changes are new clippy lints released in new version of clippy that came with Rust v1.63.0.

Is there a specific change that is concerning? We can do a huddle review if you wan to walk through the changes together.

stegaBOB:
I was talking about the lockfile. Did something else get bumped or something?

stegaBOB (approved):

PR #680, opened 2022-08-15
Allow Bubblegum program to create metadata with verified creators

danenbm:
### Notes
- Make sure that `mint_authority` is owned by Bubblegum program and
is a signer before allowing creator to be verified without all
checks in place.
- Feature currently hardcoded to inactive.
- Needed to support Bubblegum `decompress` instruction.
- Also simplifity `allow_direct_creator_writes` parameter to be only
at a lower level where it will be used.

### DIFF
--- a/token-metadata/program/src/utils.rs
+++ b/token-metadata/program/src/utils.rs
@@ -891,6 +891,16 @@ pub const SEED_AUTHORITY: Pubkey = Pubkey::new_from_array([
     0x92, 0x17, 0x2c, 0xc4, 0x72, 0x5d, 0xc0, 0x41, 0xf9, 0xdd, 0x8c, 0x51, 0x52, 0x60, 0x04, 0x26,
     0x00, 0x93, 0xa3, 0x0b, 0x02, 0x73, 0xdc, 0xfa, 0x74, 0x92, 0x17, 0xfc, 0x94, 0xa2, 0x40, 0x49,
 ]);
+
+// This equals the program address of the Bubblegum program:
+// "BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY"
+// This allows the Bubblegum program to add verified creators since they were verified as part of
+// the Bubblegum program.
+pub const BUBBLEGUM_PROGRAM_ADDRESS: Pubkey = Pubkey::new_from_array([
+    0x98, 0x8b, 0x80, 0xeb, 0x79, 0x35, 0x28, 0x69, 0xb2, 0x24, 0x74, 0x5f, 0x59, 0xdd, 0xbf, 0x8a,
+    0x26, 0x58, 0xca, 0x13, 0xdc, 0x68, 0x81, 0x21, 0x26, 0x35, 0x1c, 0xae, 0x07, 0xc1, 0xa5, 0xa5,
+]);
+
 /// Create a new account instruction
 pub fn process_create_metadata_accounts_logic(
     program_id: &Pubkey,
@@ -963,6 +973,19 @@ pub fn process_create_metadata_accounts_logic(
 
     let mut metadata: Metadata = Metadata::from_account_info(metadata_account_info)?;
     let compatible_data = data.to_v1();
+
+    // This allows the Bubblegum program to create metadata with verified creators since they were
+    // verified already by the Bubblegum program.
+    const BUBBLEGUM_ACTIVATED: bool = false;
+    let allow_direct_creator_writes = if BUBBLEGUM_ACTIVATED
+        && mint_authority_info.owner == &BUBBLEGUM_PROGRAM_ADDRESS
+        && mint_authority_info.is_signer
+    {
+        true
+    } else {
+        allow_direct_creator_writes
+    };
+
     assert_data_valid(
         &compatible_data,
         &update_authority_key,
--- a/token-metadata/program/src/deprecated_processor.rs
+++ b/token-metadata/program/src/deprecated_processor.rs
@@ -18,7 +18,6 @@ pub fn process_deprecated_create_metadata_accounts<'a>(
     program_id: &'a Pubkey,
     accounts: &'a [AccountInfo<'a>],
     data: Data,
-    allow_direct_creator_writes: bool,
     is_mutable: bool,
 ) -> ProgramResult {
     let account_info_iter = &mut accounts.iter();
@@ -50,7 +49,7 @@ pub fn process_deprecated_create_metadata_accounts<'a>(
             collection: None,
             uses: None,
         },
-        allow_direct_creator_writes,
+        false,
         is_mutable,
         false,
         false,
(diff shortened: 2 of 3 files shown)


token-metadata/js/idl/mpl_token_metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.3.3",
+  "version": "1.3.4",
```
austbot, line 2: did you manually change these?

token-metadata/program/src/utils.rs
```diff
@@ -891,6 +891,16 @@ pub const SEED_AUTHORITY: Pubkey = Pubkey::new_from_array([
     0x92, 0x17, 0x2c, 0xc4, 0x72, 0x5d, 0xc0, 0x41, 0xf9, 0xdd, 0x8c, 0x51, 0x52, 0x60, 0x04, 0x26,
     0x00, 0x93, 0xa3, 0x0b, 0x02, 0x73, 0xdc, 0xfa, 0x74, 0x92, 0x17, 0xfc, 0x94, 0xa2, 0x40, 0x49,
 ]);
+
+// This equals the program address of the Bubblegum program:
+// "BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY"
+// This allows the Bubblegum program to add verified creators since they were verified as part of
+// the Bubblegum program.
+pub const BUBBLEGUM_SEED_AUTHORITY: Pubkey = Pubkey::new_from_array([
```
austbot, line 899: you can use the Pubkeys! macros now for this

token-metadata/program/src/utils.rs
```diff
@@ -963,6 +973,19 @@ pub fn process_create_metadata_accounts_logic(
 
     let mut metadata: Metadata = Metadata::from_account_info(metadata_account_info)?;
     let compatible_data = data.to_v1();
+
+    // This allows the Bubblegum program to create metadata with verified creators since they were
+    // verified already by the Bubblegum program.
+    const BUBBLEGUM_ACTIVATED: bool = false;
```
austbot, line 979: was this your testing const?

austbot (changes requested):
nits but looks good.

token-metadata/js/idl/mpl_token_metadata.json
danenbm: No I did not manually change.  I believe this just changed when i ran `yarn api:gen`

token-metadata/program/src/utils.rs
danenbm: This is here because we cannot have this existing yet in any real deployed code, because Bubblegum is not updated to fully check signers for setting the verify flag.  So it is being added as a disabled feature.  Does that answer your question?

token-metadata/program/src/utils.rs
danenbm: The way it is now matches the `pub const SEED_AUTHORITY:` above it in code.  I can change both to use `Pubkeys!` macro.

danenbm: I tried changing it to use `pubkeys!` and it builds in my IDE which uses regular `cargo build`, but when I `cargo build-bpf` I am getting errors:
```
error[E0107]: this struct takes 0 lifetime arguments but 1 lifetime argument was supplied
   --> /home/danenbm/.cargo/registry/src/github.com-1ecc6299db9ec823/termcolor-1.1.2/src/lib.rs:375:20
    |
375 |     StdoutLock(io::StdoutLock<'a>),
    |                    ^^^^^^^^^^---- help: remove these generics
    |                    |
    |                    expected 0 lifetime arguments

error[E0107]: this struct takes 0 lifetime arguments but 1 lifetime argument was supplied
   --> /home/danenbm/.cargo/registry/src/github.com-1ecc6299db9ec823/termcolor-1.1.2/src/lib.rs:376:20
    |
376 |     StderrLock(io::StderrLock<'a>),
    |                    ^^^^^^^^^^---- help: remove these generics
    |                    |
    |                    expected 0 lifetime arguments

   Compiling percent-encoding v2.1.0
   Compiling fnv v1.0.7
For more information about this error, try `rustc --explain E0107`.
error: could not compile `termcolor` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error[E0425]: cannot find function `is` in this scope
  --> /home/danenbm/.cargo/registry/src/github.com-1ecc6299db9ec823/atty-0.2.14/src/lib.rs:98:6
   |
98 |     !is(stream)
   |      ^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: build failed

```

So I reverted the change back to how it was.

austbot: ah okay thqt may just be a client thing

token-metadata/program/src/utils.rs
austbot: yes

austbot (dismissed):

token-metadata/program/src/utils.rs
alexandruag: Just wondering, was the initial suggestion about the `pubkeys!` macro, or about `pubkey!` (singular)? For example, looks like `pub const BUBBLEGUM_SEED_AUTHORITY: Pubkey = solana_program::pubkey!("BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY");` builds fine with `cargo-bpf` as well (I think the macro is in sdk as well).

alexandruag (approved):

token-metadata/program/src/utils.rs
```diff
@@ -963,6 +973,19 @@ pub fn process_create_metadata_accounts_logic(
...
+    let allow_direct_creator_writes = if BUBBLEGUM_ACTIVATED
+        && mint_authority_info.key == &BUBBLEGUM_SEED_AUTHORITY
```
danenbm, line 981: This actually won't work as-is, because the Bubblegum program ID will not be the `mint_authority` and the program cannot be a signer.  @samuelvanderwaal noticed this as we reviewed it offline.

We also cannot just hardcode the mint_authority PDA information because `mint_authority` has a dynamic seed based on the mint account, which itself has dynamic seeds.

Sam came up with a potential solution that is simple and elegant in that it will not require changing the structure of the PDAs in Bubblegum:
1. Change the name `BUBBLEGUM_SEED_AUTHORITY` to `BUBBLEGUM_PROGRAM_ADDRESS` because that is what it is.
2. The `mint_authority` is still a signer, and since it is a PDA of Bubblegum, instead verify that the `mint_authority_info` is owned by the Bubblegum program.

danenbm: Fixed in latest version

samuelvanderwaal (approved):

alexandruag (approved):

samuelvanderwaal (approved):

austbot (approved):

austbot (approved):
version patch

jshiohaha:
Highlighting here - it looks like the version workflow failed because the changes were merged & branch was deleted before the versioning workflow could finish. https://github.com/metaplex-foundation/metaplex-program-library/runs/8023777077?check_suite_focus=true

PR #681, opened 2022-08-16
Adding `verify_creator` and `unverify_creator` to Bubblegum

danenbm:
- Analogous to `SignMetadata` and `RemoveCreatorVerification` in
token-metadata program.
- Adding new `BubblegumError` variants.
- Add code to prevent minting with `verified` set to true unless
creator is a signer.


danenbm:
> Workflow `Verify package library` found differences when running `yarn api:gen` in the JS lib for `bubblegum`. Please see the job for more details: https://github.com/metaplex-foundation/metaplex-program-library/actions/runs/2867438467.

Ran `yarn api:gen`

bubblegum/program/src/lib.rs
```diff
@@ -603,9 +726,12 @@ pub mod bubblegum {
         let merkle_slab = &ctx.accounts.merkle_slab;
 
         // The mint authority must sign if it is not equal to the tree authority
-        if mint_authority.key() != ctx.accounts.authority.key() {
+        let signer = if mint_authority.key() != ctx.accounts.authority.key() {
```
danenbm, line 729: This particular concept needs close look and review and please feel free to ping me with offline discussion about if I'm doing this wrong.

Basically if the mint_authority is a signer it can be used to verify a creator, but not the tree authority if the tree authority is not a signer.  I'm not 100% sure I fully understand if that is correct behavior.

bubblegum/program/src/lib.rs
```diff
@@ -603,9 +726,12 @@ pub mod bubblegum {
...
             assert!(mint_authority.is_signer);
-        }
+            Some(mint_authority.key())
+        } else {
+            None
+        };
```
danenbm, line 734: This particular concept needs close look and review and please feel free to ping me with offline discussion about if I'm doing this wrong.

Basically if the mint_authority is a signer it can be used to verify a creator, but not the tree authority if the tree authority is not a signer. I'm not 100% sure I fully understand if that is correct behavior.

bubblegum/program/src/lib.rs
```diff
@@ -516,6 +546,99 @@ fn process_mint_v1<'info>(
...
+    if !creators.iter().any(|c| c.address == creator) {
+        return Err(BubblegumError::CreatorNotFound.into());
+    }
+
+    // User-provided creator Vec must result in same user-provided creator hash.
+    let creator_data = creators
+        .iter()
+        .map(|c| [c.address.as_ref(), &[c.verified as u8], &[c.share]].concat())
+        .collect::<Vec<_>>();
+    let calculated_creator_hash = keccak::hashv(
```
austbot, line 580: maybe this is a rust thing but usng the same var confused me

bubblegum/program/src/lib.rs
austbot: this is to ensure if the authority, which is a PDA of bubble gum and is the tree config I guess doesnt sign then we cannot mint, this is to ensure the right approved mint requester is signing.

austbot (approved):
This seems like it works, in absence of test which I know are coming I dont 100% know.

bubblegum/program/src/lib.rs
danenbm: Yeah I will change them to make it more clear.

Done in next version.

bubblegum/program/src/lib.rs
danenbm: Ok I read your comments and talked to Noah and studied the code more, and I understand now.

bubblegum/program/src/lib.rs
```diff
@@ -516,6 +547,99 @@ fn process_mint_v1<'info>(
...
+            .as_ref(),
+    );
+
+    let asset_id = get_asset_id(&merkle_slab.key(), nonce);
+    let previous_leaf = LeafSchema::new_v0(
+        asset_id,
+        owner.key(),
+        delegate.key(),
+        nonce,
+        data_hash,
```
danenbm, line 616: TODO: I need to rehash data after updating creator array, because this hash also contains creator array.  This change in progress now.

danenbm: Fixed in next version

bubblegum/program/src/lib.rs
```diff
@@ -92,11 +93,29 @@ pub struct Burn<'info> {
...
+    #[account(
+        seeds = [merkle_slab.key().as_ref()],
+        bump,
+    )]
+    pub authority: Account<'info, TreeConfig>,
+    /// CHECK: This account is checked in the instruction
+    pub owner: UncheckedAccount<'info>,
     /// CHECK: This account is chekced in the instruction
+    pub delegate: UncheckedAccount<'info>,
+    pub creator: Signer<'info>,
```
danenbm, line 114: Wondering should I add a `pub payer: Signer<'info>,` account to these new Creator instructions.  In token-metadata we don't have a `payer` account for creator verification, but we do have one for all the Collection verification functions.

austbot (approved):

bubblegum/program/src/lib.rs
austbot: I think its assumed the creato has to do this themsef thats why we dont have a payer.

bubblegum/program/src/lib.rs
```diff
@@ -516,6 +547,99 @@ fn process_mint_v1<'info>(
...
+        .map(|c| [c.address.as_ref(), &[c.verified as u8], &[c.share]].concat())
+        .collect::<Vec<_>>();
+    let calculated_creator_hash = keccak::hashv(
+        provided_creator_data
+            .iter()
+            .map(|c| c.as_slice())
+            .collect::<Vec<&[u8]>>()
+            .as_ref(),
+    );
+    assert_eq!(creator_hash, calculated_creator_hash.to_bytes());
```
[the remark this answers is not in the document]
danenbm: 👍 Sounds good, will work on this as part of the Linear task

PR #682, opened 2022-08-17
Feat: candy machine sized collections

stegaBOB:
This is technically a breaking change. Because anchor enforces that the account is writable, the `mut` constraint that I added wouldn't actually work. I think the only way to make this work is to not mark it as mutable at the program level but deploy the IDL and rust crate with the mutable constraint set.


stegaBOB:
I made this based off of #669, so this shouldnt be merged in and will need to be sorta rebased once that PR is in.

samuelvanderwaal (dismissed):

samuelvanderwaal:
LGTM, make sure to run the version bump before merging.

stegaBOB:
version minor

stegaBOB (commented):
version minor

candy-machine/program/src/processor/collection/set_collection_during_mint.rs
```diff
@@ -95,6 +99,36 @@ pub fn handle_set_collection_during_mint(ctx: Context<SetCollectionDuringMint>)
     if !cmp_pubkeys(&collection_pda.mint, &collection_mint.key()) {
         return Ok(());
     }
+
+    let collection_metadata: Metadata =
+        Metadata::safe_deserialize(&ctx.accounts.collection_metadata.data.borrow_mut())?;
+
+    let collection_instruction = if collection_metadata.collection_details.is_some() {
```
austbot, line 106: you MUST check if the account info is writable .

austbot (changes requested):

candy-machine/program/src/processor/collection/set_collection_during_mint.rs
stegaBOB: I'll add a custom error here to be more descriptive.

samuelvanderwaal (approved):

austbot (approved):

PR #685, opened 2022-08-22
build: add build:docs command for each package and use in an auto-build workflow

jshiohaha:
### Changes

* Update every `package.json` `build:docs` script to call typedoc and output to a common ancestor `docs` dir
  * Output is similar to what we see in https://github.com/metaplex-foundation/beet
* Add workflow file to update docs when there is a change in the `js` dir - logic inspired by https://github.com/metaplex-foundation/beet/blob/master/sh/update-docs

### How the workflow works

1. On push to master, we get the commit SHA and try to associate with a PR
2. We map files changed to the higher level packages, e.g. candy-machine, token-metadata, etc.
3. We checkout the `gh-pages` and sync with `master`
4. For every changed package, we run `yarn build:docs` to re-generate that package's docs
5. We call `git add` so that git tracks the updates
6. We call `git push` to push the changes to the `gh-pages` branch
   * 🚧 note: pushing directly to `master` from a workflow doesn't work because it's write-protected. Since `gh-pages` is not write-protected, I think we should be able to push directly to that branch as long as token permissions allow writes.

### Testing

* token-metadata: https://github.com/jshiohaha/metaplex-program-library/runs/7945457822?check_suite_focus=true
* candy-machine: https://github.com/jshiohaha/metaplex-program-library/runs/7945236866?check_suite_focus=true

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,106 @@
+name: Publish docs
+
+on:
+  push:
+    branches: [master]
+
+permissions:
+  id-token: write
+  contents: write
+
+jobs:
+  dump-context:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Dump GitHub context
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: echo "$GITHUB_CONTEXT"
+
+  get-changes-scope:
+    runs-on: ubuntu-latest
+    outputs:
+      changed-packages: ${{ steps.map-changed-files-to-pkg.outputs.result }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get PR for push event HEAD commit
+        uses: ./.github/actions/get-pr-for-commit
+        id: get-pr-for-commit
+        with:
+          commit-sha: ${{ github.sha }}
+
+      - name: List changed files
+        uses: ./.github/actions/list-changed-files
+        id: list-changed-files
+        with:
+          pull-number: ${{ steps.get-pr-for-commit.outputs.pull-number }}
+
+      - name: Get scope of changed packages
+        uses: actions/github-script@v4
+        id: map-changed-files-to-pkg
+        with:
+          script: |
+            const files = ${{ steps.list-changed-files.outputs.changed-files }}
+            // yarn doc flow only generates docs for JS
+            const regexp = /[a-zA-Z\-]+\/(js)/g
+            const uniqueFilesObj = files
+              .filter(f => regexp.test(f))
+              .reduce((p, file) => {
+                const pkgForFile = file.split("/")[0];
+                if (!p[pkgForFile]) p[pkgForFile] = 1
+                return p
+              }, {})
+            const result = Array.from(Object.keys(uniqueFilesObj))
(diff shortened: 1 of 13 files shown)


auction-house/js/package.json
```diff
@@ -10,7 +10,7 @@
     "postversion": "git add package.json && git commit -m \"chore: update $npm_package_name to v$npm_package_version\" && git tag $npm_package_name@$npm_package_version",
     "prepublishOnly": "yarn check:publish-ready",
     "postpublish": "git push origin && git push origin --tags",
-    "build:docs": "typedoc",
+    "build:docs": "docsdir=../../docs/auction-house; rimraf $docsdir && typedoc --out $docsdir",
```
thlorenz, line 13: Pretty sure that `env_var=` won't work on windows (in case peeps wanna run this locally).
Not a huge deal though, esp. considering we don't have many windows devs :)

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,106 @@
...
+          pull-number: ${{ steps.get-pr-for-commit.outputs.pull-number }}
+
+      - name: Get scope of changed packages
+        uses: actions/github-script@v4
+        id: map-changed-files-to-pkg
+        with:
+          script: |
+            const files = ${{ steps.list-changed-files.outputs.changed-files }}
+            // yarn doc flow only generates docs for JS
+            const regexp = /[a-zA-Z\-]+\/(js)/g
```
thlorenz, line 47: Confused about this regex as the files that'll be updated will be TypeScript, i.e. `*.ts` files.

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,106 @@
...
+        id: list-changed-files
+        with:
+          pull-number: ${{ steps.get-pr-for-commit.outputs.pull-number }}
+
+      - name: Get scope of changed packages
+        uses: actions/github-script@v4
+        id: map-changed-files-to-pkg
+        with:
+          script: |
+            const files = ${{ steps.list-changed-files.outputs.changed-files }}
```
thlorenz, line 45: If I understand correctly this is to ensure that we only run the docs update inside the folders where the SDK files actually changed?

thlorenz (changes requested):
Aside from the regex question LGTM, so let's figure that part out and we can merge.
Nice work!

.github/workflows/publish-docs.yml
jshiohaha: Yeah, I previously used this regex to identify packages - which were in the form `<pkg>/js` or `<pkg>/program`. But, I think regex matching on `*.ts` should work too.

auction-house/js/package.json
jshiohaha: Oh, shoot - good to know! I didn't even know that about our Windows counterparts 😰 the alternative (that I know of) is to repeat the path to docs.

If you're ok with possible OS compat, I will leave as is?

.github/workflows/publish-docs.yml
thlorenz: Ah ok so this matches on the folder .. got it.

auction-house/js/package.json
thlorenz: Given this runs in CI it's probably fine

PR #687, opened 2022-08-23
 Adding `verify_collection`, `unverify_collection`, and `set_and_verify_collection` to Bubblegum

danenbm:
### Notes
- Similar functionality as token-metadata versions of these functions.
- Uses some existing assertion functions from token-metadata.
- Adding new `BubblegumError` variants.
- Add code to prevent minting in Bubblegum with collection `verified` set to true.
- Regenerating APIs.
- Make use of same collection authority delegate system as used in token-metadata.  In fact it calls `assert_has_collection_authority()` from token-metadata to verify.
- For  `set_and_verify_collection()` require that either the tree authority signed this transaction, or the tree authority is the collection update authority.
- Add `payer` to both creator and collection verification instructions.

### Discussion items
1. Discuss if we want to allow minting with collection verified.
   a. At first I thought we did but it looks like this is currently not allowed in existing token-metadata program.
   b. Also it will require adding additional accounts to `mint_v1()` which does not seem like what we want right now.
   c. A solution that might work well is in a follow-on PR add a separate `mint_to_collection_v1()` instruction that takes all the required collection accounts.
   d. Alternatively we could just say for minting to a collection you just have to do `mint_v1()` followed by `verify_collection()`.  To be honest in this case we don't even need `set_and_verify_collection()` because we would just set it as unverified in `mint_v1()`.
1. This one is more of an FYI on a variance from token-metadata.  For `set_and_verify_collection()`, this PR requires that either the tree authority signed this transaction, or the tree authority is the collection update authority which means the leaf update is approved via proxy, when we later call `assert_has_collection_authority()`.
   a. This is similar to logic in token-metadata for `set_and_verify_collection()` except this logic also allows the tree authority (which we are treating as the leaf metadata authority) to be different than the collection authority (actual or delegated).  The token-metadata program required them to be the same.


bubblegum/program/src/lib.rs
```diff
@@ -119,6 +122,30 @@ pub struct CreatorVerification<'info> {
...
+    )]
+    pub authority: Account<'info, TreeConfig>,
+    /// CHECK: This account is checked in the instruction
+    pub owner: UncheckedAccount<'info>,
+    /// CHECK: This account is checked in the instruction
+    pub delegate: UncheckedAccount<'info>,
+    pub collection_authority: Signer<'info>,
+    /// CHECK: This account is checked in the instruction
+    pub collection_mint: UncheckedAccount<'info>,
+    pub collection_metadata: Box<Account<'info, TokenMetadata>>,
```
austbot, line 139: this needs to be mut so we can increment size

austbot: via CPI to token metadata

bubblegum/program/src/lib.rs
```diff
@@ -665,6 +704,120 @@ fn process_creator_verification<'info>(
...
+    // If the NFT has collection data, we set it to the correct value after doing some validation.
+    if let Some(collection) = &mut message.collection {
+        // Create a token-metadata Metadata struct and ONLY populate the collection field.
+        // This is being created only to interact with some token-metadata assert functions.
+        let mut mpl_metadata = mpl_token_metadata::state::Metadata::default();
+        mpl_metadata.collection = Some(collection.adapt());
+
+        // Collection verify assert from token-metadata program.
+        assert_collection_verify_is_valid(
+            &mpl_metadata,
```
austbot, line 748: surprised this works, but makes alot of sense, good reuse of code

bubblegum/program/src/lib.rs
```diff
@@ -665,6 +704,120 @@ fn process_creator_verification<'info>(
...
+        // Collection authority assert from token-metadata.
+        assert_has_collection_authority(
+            &collection_authority,
+            &collection_metadata,
+            collection_mint.key,
+            collection_authority_record,
+        )?;
+
+        // This handler can only verify non-sized NFTs so far.
+        if !sized && collection_metadata.collection_details.is_some() {
```
austbot, line 770: we can rebuild him

bubblegum/program/src/lib.rs
```diff
@@ -665,6 +704,120 @@ fn process_creator_verification<'info>(
...
+            return Err(BubblegumError::SizedCollection.into());
+        }
+
+        // Update collection in metadata args.
+        collection.verified = verify;
+
+        // Calculate new data hash.
+        let metadata_args_hash = keccak::hashv(&[message.try_to_vec()?.as_slice()]);
+        let updated_data_hash = keccak::hashv(&[
```
austbot, line 779: very nice

bubblegum/program/src/lib.rs
```diff
@@ -665,6 +704,120 @@ fn process_creator_verification<'info>(
...
+        // This handler can only verify non-sized NFTs so far.
+        if !sized && collection_metadata.collection_details.is_some() {
+            return Err(BubblegumError::SizedCollection.into());
+        }
+
+        // Update collection in metadata args.
+        collection.verified = verify;
+
+        // Calculate new data hash.
+        let metadata_args_hash = keccak::hashv(&[message.try_to_vec()?.as_slice()]);
```
austbot, line 778: im worried there is not enough validation on this message

austbot (changes requested):

bubblegum/program/src/lib.rs
danenbm: This works because the types all match:
-  mpl_metadata: I instantiate the NFT's metadata with the previous two lines of code (more on this at bottom).
- collection_metadata can be deserialized because `AccountSerialize` and `AccountDeserialize` were implemented for the struct `TokenMetadata(mpl_token_metadata::state::Metadata)` in metaplex_anchor.rs.
- collection_mint: This is just used for its key.
- edition_account: This is just account info, and it is actually deserialized inside of `assert_collection_verify_is_valid()`, but that token-metadata function knows how to do that.

Regarding the mpl_metadata that I instantiate, I could get away with not having to create that if I change the prototype of `assert_collection_verify_is_valid()` since it really only uses the collection information not any of the rest of the NFT's metadata.

See this idea for an example: https://github.com/metaplex-foundation/metaplex-program-library/pull/690

The reason I did not do it that way is because then I would have to wait for token-metadata to be published again before this contract picks up the updated prototype, and I am not sure of our timeline for token-metadata publishing.

bubblegum/program/src/lib.rs
danenbm: Yes I found that out (that it needed to be mut) in my local copy where I'm working on sized collection support.

Although rather than CPI to token metadata, I think I may be able to just increment it and reserialize it in the bubblegum contract, because `AccountSerialize` and `AccountDeserialize` were implemented for the struct `TokenMetadata(mpl_token_metadata::state::Metadata)` in metaplex_anchor.rs.

bubblegum/program/src/lib.rs
danenbm: Working on sized collection, will update this PR soon.

bubblegum/program/src/lib.rs
danenbm: Are you referring to the new variable name :smile: ?

bubblegum/program/src/lib.rs
danenbm: Yes you are absolutely right.  I need to validate that the metadata they gave me results in the existing data hash, and then ONLY flip the collection verified flag and use that for the new metadata.

I did this for creator verification but forgot to do it here.  Thank you for catching that.  We will make sure to cover this case as part of the testing task which is coming up right after this PR.

danenbm: Fixed in latest version.

- Verify that user-provided metadata results in same hash
as user-provided `data_hash`.
- This is done to ensure that creator and collection
verification functions only modify the verified flag in the
metadata and do not pick up other extraneous changes.
- This was already being done with the creator hash in
creator verification.
- Also made change to use helper functions to do creator
and data hashing in one place.

bubblegum/program/src/lib.rs
danenbm: Updated in latest version.

bubblegum/program/src/lib.rs
danenbm: Added support for sized collection.

In token-token-metadata we have `verify_collection()`, `verify_sized_collection_item()`, `unverify_collection()`, and `unverify_sized_collection_item()`.

For Bubblegum I am combining functionality for both sized and unsized collection verification by autodetecting in the Bubblegum program whether its a sized or unsized collection.

bubblegum/program/src/lib.rs
danenbm: Actually I take that back I think you were right and it will have to be a CPI to token-metadata as that program is the account owner.

danenbm: Converting this to a CPI now

danenbm: Converted to use CPI in latest version.

bubblegum/program/src/lib.rs
danenbm: I my latest version I have to update token-metadata anyways, so the changes from https://github.com/metaplex-foundation/metaplex-program-library/pull/690 are now included in this PR.

danenbm:
> Workflow `Verify package library` found differences when running `yarn api:gen` in the JS lib for `bubblegum`. Please see the job for more details: https://github.com/metaplex-foundation/metaplex-program-library/actions/runs/2921325909.

Reran API gen for token-metadata and bubblegum and picked up the changes.

samuelvanderwaal (changes requested):

token-metadata/program/src/processor.rs
```diff
@@ -230,6 +231,10 @@ pub fn process_instruction<'a>(
             msg!("Instruction: Set Collection Size");
             set_collection_size(program_id, accounts, args)
         }
+        MetadataInstruction::BubblegumSetCollectionSize(args) => {
+            msg!("Instruction: Bubblegum Program Set Collection Size");
+            set_collection_size(program_id, accounts, args)
```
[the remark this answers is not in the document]
danenbm: Fixed

token-metadata/program/src/instruction.rs
```diff
@@ -510,6 +510,14 @@ pub enum MetadataInstruction {
     #[account(3, optional, name="collection_authority_record", desc="Collection Authority Record PDA")]
     SetCollectionSize(SetCollectionSizeArgs),
 
+    /// Set size of an existing collection.
```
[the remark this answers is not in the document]
danenbm: Fixed

bubblegum/program/src/lib.rs
```diff
@@ -119,6 +126,42 @@ pub struct CreatorVerification<'info> {
...
+    pub tree_delegate: UncheckedAccount<'info>,
+    pub collection_authority: Signer<'info>,
+    /// CHECK: This account is checked in the instruction
+    pub collection_mint: UncheckedAccount<'info>,
+    pub collection_metadata: Box<Account<'info, TokenMetadata>>,
+    /// CHECK: This account is checked in the instruction
+    pub edition_account: UncheckedAccount<'info>,
+    /// CHECK: This is just used as a signing PDA.
+    #[account(
+        seeds = [COLLECTION_CPI_PREFIX.as_ref()],
```
danenbm, line 153: I just put a string here for a seed.  In actuality I don't think I need any seed, this is really only a signing PDA and the only thing that matters is that it is owned by Bubblegum and can be a signer for a CPI into token-metadata.

That being said, do any of you have any feedback or suggestions for different PDA seeds?  I could leave this as-is, or have `seeds = []`, or also include `merkle_slab.key()` like in most of the other Bubblegum PDAs.

bubblegum/program/Cargo.toml
```diff
@@ -19,7 +19,7 @@ anchor-lang = { version = "0.25.0", features = [ "init-if-needed" ] }
 anchor-spl = { version = "0.25.0" }
 spl-token = { version = "3.3.0", features = [ "no-entrypoint" ] }
 spl-associated-token-account = { version = "1.0.5", features = [ "no-entrypoint" ] }
-mpl-token-metadata = { version = "1.3.3", features = ["no-entrypoint"] }
+mpl-token-metadata = {path = "../../token-metadata/program" , features = ["no-entrypoint"] }
```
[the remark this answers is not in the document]
alexandruag: A practice that's encountered quite often in monorepos hosting multiple crates is to specify both a path and version dependency for local dependencies (more details [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations)). What ends up happening is that things are resolved locally while developing (so it's more streamlined to work with changes across the repo), but at publishing time, the local path dependency is automatically ignored.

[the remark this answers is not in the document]
danenbm: Latest version of code has:
```
- mpl-token-metadata = { version = "1.3.3", features = ["no-entrypoint"] }
+ mpl-token-metadata = { version = "1.3.5", path = "../../token-metadata/program" , features = ["no-entrypoint"] }
```
So I am using the local path as well as the version number.  Still wondering if it is OK to update token-metadata version even though we are not publishing a new version.

bubblegum/program/src/lib.rs
samuelvanderwaal: In this case, I feel that you only need a single seed, and that the seed should indicate what the PDA's purpose is. Basically, self-documentation. I think "signer" or "program_signer" makes sense but don't have a super strong opinion otherwise. If you think it's possible you'd want multiple program signers in the future, then you'd want to add a second seed to differentiate what kind of signer or what purpose it is for.

bubblegum/program/Cargo.toml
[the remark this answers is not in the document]
samuelvanderwaal: Why doesn't path and version 1.3.4 work?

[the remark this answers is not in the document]
danenbm: That does, I just didn't like it since 1.3.4 doesn't actually contain my changes, but I just removed the path from it for now.

bubblegum/program/src/lib.rs
danenbm: I'm using "collection_cpi" for now, can change if desired

danenbm:
> Workflow `Verify package library` found differences when running `yarn api:gen` in the JS lib for `token-metadata`. Please see the job for more details: https://github.com/metaplex-foundation/metaplex-program-library/actions/runs/2959164987.

I have regenerated api for token-metadata

danenbm:
> Workflow `Verify package library` found differences when running `yarn api:gen` in the JS lib for `bubblegum`. Please see the job for more details: https://github.com/metaplex-foundation/metaplex-program-library/actions/runs/2959146367.

I have regenerated api for bubblegum

samuelvanderwaal (approved):

austbot (approved):

bubblegum/program/src/lib.rs
```diff
@@ -574,19 +634,12 @@ fn process_creator_verification<'info>(
...
-            .collect::<Vec<&[u8]>>()
-            .as_ref(),
-    );
-    assert_eq!(creator_hash, calculated_creator_hash.to_bytes());
+    let incoming_creator_hash = hash_creators(&message.creators)?;
+    assert_eq!(creator_hash, incoming_creator_hash);
+
+    // User-provided metadata must result in same user-provided data hash.
+    let incoming_data_hash = hash_metadata(&message)?;
+    assert_eq!(data_hash, incoming_data_hash);
```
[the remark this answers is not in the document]
danenbm: yes will do this in follow on PR.

bubblegum/program/src/lib.rs
```diff
@@ -606,32 +659,16 @@ fn process_creator_verification<'info>(
         })
         .collect::<Vec<Creator>>();
 
+    // Calculate new creator hash.
+    let updated_creator_hash = hash_creators(&message.creators)?;
```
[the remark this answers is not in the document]
danenbm: I believe this should be accomplished by `message.creators = updated_creator_vec;` below before redoing the data hash.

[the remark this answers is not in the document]
danenbm: I missed this, but now we have found it during initial test writing.  Fixed in https://github.com/metaplex-foundation/metaplex-program-library/pull/719

PR #688, opened 2022-08-23
nft-packs: adding IDL and generated SDK

thlorenz:
After updating shank, beet and solita to support all features that nft-packs requires I 
generated its SDK.

This was done in a few steps:

1. added shank dependency and `.solitarc.js`
2. annotated all accounts with `ShankAccount`
3. annotated all instructions with the required accounts
4. added solita dep
5. generated IDL and SDK

### NOTE: Regarding step _3._ 

Someone more familiar with the nft-packs program should carefully review that the accounts are
correctly annotated.
I derived them from the comments that were present and cross-checked with the related
processing code, but would like to make sure that we don't introduce errors here that would
make some SDK instructions unusable.

see [this commit](https://github.com/metaplex-foundation/metaplex-program-library/pull/688/commits/901b7f41306086e1513b6d44596469ff04ec0933)


.github/workflows/sdk-nft-packs.yml
```diff
@@ -0,0 +1,38 @@
...
+    outputs:
+      core: ${{ steps.filter.outputs.core }}
+      package: ${{ steps.filter.outputs.package }}
+    steps:
+    - uses: actions/checkout@v2
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
```
jshiohaha, line 22: If you want to reference `core` as an output, I think you need to add this to the filters

```
core:
  - 'core/**'
```

Existing example here: https://github.com/metaplex-foundation/metaplex-program-library/blob/master/.github/workflows/sdk-token-vault.yml#L10-L29

jshiohaha (approved):
Action related changes almost perfect as far as I can tell - 1 small comment.

blockiosaurus (dismissed):
Looks good! I'm good to merge once @jshiohaha approves as well.

.github/workflows/sdk-nft-packs.yml
thlorenz: I purposely removed that reference since this SDK does not depend on core .. actually most don't anymore.

jshiohaha: Makes sense. Then, I think you need to remove other places referencing `core`

* Line 13: `core: ${{ steps.filter.outputs.core }}`
* Line 28: `needs.changes.outputs.core == 'true'`

thlorenz: Yes good catch .. was an oversight .. fixed.

jshiohaha (approved):

blockiosaurus (approved):

PR #690, opened 2022-08-24
Modify `assert_collection_verify_is_valid` to make it more usable with Bubblegum

danenbm:


danenbm:
This functionality is now included in https://github.com/metaplex-foundation/metaplex-program-library/pull/687

PR #699, opened 2022-08-28
Update bubblegum to use spl version of Compression

austbot:
Compression primitive contracts have been moved form the jarry candy land repo into SPL. This PR connects the two. We change all candy land references to point at the new names. This should not change functionality.

This PR Also blocks my work on Blockbuster (ironically)


bubblegum/program/Cargo.toml
```diff
@@ -15,13 +15,14 @@ cpi = ["no-entrypoint"]
...
-anchor-lang = { version = "0.25.0", features = [ "init-if-needed" ] }
+anchor-lang = { version = "0.25.0", features = ["init-if-needed"] }
 anchor-spl = { version = "0.25.0" }
-spl-token = { version = "3.3.0", features = [ "no-entrypoint" ] }
-spl-associated-token-account = { version = "1.0.5", features = [ "no-entrypoint" ] }
+spl-token = { version = "3.3.0", features = ["no-entrypoint"] }
+spl-associated-token-account = { version = "1.0.5", features = ["no-entrypoint"] }
 mpl-token-metadata = { version = "1.3.3", features = ["no-entrypoint"] }
-gummyroll = { git = "https://github.com/jarry-xiao/candyland.git", features = ["cpi"] }
+spl-compression = { git = "https://github.com/solana-labs/solana-program-library", features = ["cpi"] }
```
alexandruag, line 23: I think it's a good practice for `git` dependencies to also refer to a specific commit using `rev = <hash>` in the specification, so we don't inadvertently end up consuming something unexpected just because a branch advances.

bubblegum/program/Cargo.toml
```diff
@@ -15,13 +15,14 @@ cpi = ["no-entrypoint"]
...
 bytemuck = "1.8.0"
+solana-program = "^1.10.35"
```
alexandruag, line 25: As a bit of trivia, the caret syntax is equivalent to the default one, so `"^1.10.35"` is the same as `"1.10.35"`.

alexandruag (commented):
Looks good; haven't double checked the new discriminators but we would catch if something's wrong when adding the tests.

bubblegum/program/src/utils.rs
```diff
@@ -67,15 +68,15 @@ pub fn replace_leaf<'info>(
     let authority_pda_signer = &[&seeds[..]];
     let cpi_ctx = CpiContext::new_with_signer(
         gummyroll_program.clone(),
-        gummyroll::cpi::accounts::Modify {
+        spl_compression::cpi::accounts::Modify {
             authority: authority.clone(),
-            merkle_roll: merkle_roll.clone(),
-            candy_wrapper: candy_wrapper.clone(),
+            merkle_tree: merkle_roll.clone(),
```
danenbm, line 73: Should all instances (such as this) of `merkle_roll` as a variable name be changed to merkle_tree now?

Similarly, is `merkle_slab` still the right name where it is used?

danenbm (changes requested):

bubblegum/program/src/utils.rs
austbot: ill change to tree

bubblegum/program/Cargo.toml
austbot: ah yes thanks

bubblegum/program/Cargo.toml
austbot: good call

bubblegum/program/src/lib.rs
```diff
@@ -518,6 +530,9 @@ pub fn get_instruction_type(full_bytes: &[u8]) -> InstructionName {
         [54, 85, 76, 70, 228, 250, 164, 81] => InstructionName::DecompressV1,
         [116, 110, 29, 56, 107, 219, 42, 93] => InstructionName::Burn,
         [82, 193, 176, 117, 176, 21, 115, 253] => InstructionName::Compress,
+        [165, 83, 136, 142, 89, 202, 47, 220] => InstructionName::CreateTree,
+        [52, 17, 96, 132, 71, 4, 85, 194] => InstructionName::VerifyCreator,
+        [107, 178, 57, 39, 105, 115, 112, 152] => InstructionName::UnverifyCreator,
         _ => InstructionName::Unknown,
```
danenbm, line 536: Oh  I didn't  add `VerifyCollecton`, `UnverifyCollection`, `SetAndVerifyCollection` to this in my PR.   Could you add it?  Otherwise I can in another PR.

bubblegum/program/src/lib.rs
```diff
@@ -905,15 +920,15 @@ pub mod bubblegum {
...
         let authority_pda_signer = &[&seeds[..]];
         let cpi_ctx = CpiContext::new_with_signer(
-            ctx.accounts.gummyroll_program.to_account_info(),
-            gummyroll::cpi::accounts::Initialize {
+            ctx.accounts.compression_program.to_account_info(),
+            spl_compression::cpi::accounts::Initialize {
                 authority: ctx.accounts.authority.to_account_info(),
-                merkle_roll: merkle_slab,
-                candy_wrapper: ctx.accounts.candy_wrapper.to_account_info(),
+                merkle_tree: merkle_tree,
```
danenbm, line 926: Nit: This can probably just be `merkle_tree` once with no `:`

bubblegum/program/src/lib.rs
```diff
@@ -905,15 +920,15 @@ pub mod bubblegum {
...
+                log_wrapper: ctx.accounts.candy_wrapper.to_account_info(),
             },
             authority_pda_signer,
         );
-        gummyroll::cpi::init_empty_gummyroll(cpi_ctx, max_depth, max_buffer_size)
+        spl_compression::cpi::init_empty_merkle_tree(cpi_ctx, max_depth, max_buffer_size)
```
danenbm, line 931: Do you know what was the difference between a gummyroll and a merkle_slab?  Are they all fine being called a merkle_tree because in this PR we are changing `init_empty_gummyroll` to `init_empty_merkle_tree`, and `merkle_slab` to `merkle_tree`.  I don't want to incorrectly overload the term but I don't know the detail.

bubblegum/program/src/utils.rs
```diff
@@ -91,14 +92,14 @@ pub fn append_leaf<'info>(
...
+            log_wrapper: candy_wrapper.clone(),
```
danenbm, line 98: Nit the param is still called `candy_wrapper`, probably should be changed to `log_wrapper`?

danenbm (changes requested):
Looks good had a couple nits/questions.

bubblegum/program/src/lib.rs
austbot: they were the same

alexandruag:
If we want to pin the `spl-compression` dependency to a particular commit hash, we should add that change to the `Cargo.toml` file also, because it looks like that line remains unchanged even though the previous comment was resolved.

bubblegum/program/src/lib.rs
austbot: other pr pls

bubblegum/program/src/utils.rs
austbot: can change

danenbm (dismissed):
Looks great!

alexandruag (approved):

danenbm (approved):

PR #700, opened 2022-08-30
Feat: Candy Guard Sized Collections

stegaBOB:


candy-core/program/src/instructions/mint.rs
```diff
@@ -28,18 +28,21 @@ pub fn mint<'info>(ctx: Context<'_, '_, '_, 'info, Mint<'info>>, creator_bump: u
         return err!(CandyError::CandyMachineEmpty);
     }
 
-    if let Some(collection) = &candy_machine.collection_mint {
+    if let Some(collection_settings) = &candy_machine.collection_settings {
```
austbot, line 31: Dont we want to error if they havent set a collection? We want to make sure they set a collection.

candy-core/program/src/instructions/add_collection.rs
```diff
@@ -73,7 +78,10 @@ pub fn add_collection(ctx: Context<AddCollection>) -> Result<()> {
         );
     }
 
-    candy_machine.collection_mint = Some(mint.key());
+    candy_machine.collection_settings = Some(CollectionSettings::new(
```
austbot, line 81: collection must be required.

austbot (changes requested):

PR #702, opened 2022-08-31
Replacing asserts with explicit errors + more

danenbm:
- Also:
  - Adding new error variants.
  - Cargo fmt and clippy fixes.
  - Removing a collection TODO since we solved
a different way.
  - Adding `#[allow(unreachable_patterns)]` on one specific unsupported
schema error.
  - Updating token-metadata to be a crates.io version.
  - Regenerating API.

### DIFF
--- a/bubblegum/program/src/lib.rs
+++ b/bubblegum/program/src/lib.rs
@@ -15,29 +15,26 @@ use {
     },
     anchor_lang::{
         prelude::*,
-        system_program::System,
         solana_program::{
+            account_info::AccountInfo,
             keccak,
             program::{invoke, invoke_signed},
             program_error::ProgramError,
             program_pack::Pack,
             system_instruction,
-            account_info::AccountInfo
         },
-    },
-    spl_compression::{
-        program::SplCompression,
-        Node,
-        data_wrapper::{
-            Wrapper,
-            wrap_event
-        },
+        system_program::System,
     },
     mpl_token_metadata::{
         assertions::collection::{
             assert_collection_verify_is_valid, assert_has_collection_authority,
         },
         state::CollectionDetails,
+    },
+    spl_compression::{
+        data_wrapper::{wrap_event, Wrapper},
+        program::SplCompression,
+        Node,
     },
     spl_token::state::Mint as SplMint,
     std::collections::HashSet,
@@ -473,7 +470,7 @@ pub struct SetTreeDelegate<'info> {
     pub tree_authority: Account<'info, TreeConfig>,
 }
 
-pub fn hash_creators(creators: &Vec<Creator>) -> Result<[u8; 32]> {
+pub fn hash_creators(creators: &[Creator]) -> Result<[u8; 32]> {
     // Convert creator Vec to bytes Vec.
     let creator_data = creators
         .iter()
@@ -512,7 +509,10 @@ pub enum InstructionName {
     Burn,
     CreateTree,
     VerifyCreator,
-    UnverifyCreator
+    UnverifyCreator,
+    VerifyCollection,
+    UnverifyCollection,
+    SetAndVerifyCollection,
 }
 
 pub fn get_instruction_type(full_bytes: &[u8]) -> InstructionName {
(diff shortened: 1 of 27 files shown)


bubblegum/program/src/lib.rs
```diff
@@ -1373,9 +1383,10 @@ pub mod bubblegum {
                     version: Version::V1,
                     tree_id: ctx.accounts.voucher.merkle_slab.key(),
                     id: get_asset_id(&ctx.accounts.voucher.merkle_slab.key(), nonce),
-                    nonce: nonce,
+                    nonce,
                 })
             }
+            #[allow(unreachable_patterns)]
```
alexandruag, line 1389: Hmm, is there any reason why we don't simply remove the unreachable `_ => Err(BubblegumError::UnsupportedSchemaVersion)` match arm?

danenbm: I think there was a case in some other code - our geyser plugin - where a new version of a data type was being sent and consumed by the plugin, but the plugin was silently skipping the data because it was a V2 version.  @austbot might know about this more since I think he observed the issue.

I know in a compiled-together system the Rust compiler will ensure that the match will be exhaustive, but maybe in the world of cdylib's on a blockchain this scenario is possible?  I'm not sure.  @austbot could you comment on the need for a catch-all case vs. trusting the rust compiler take care of exhaustive match statements?

alexandruag: So I guess the main thing here is that, from Rust's perspective, there's no other variant possible than `LeafSchema::V1`, which means that optimized builds will likely not even include the unreachable pattern. Moreover, if there's some unsafe code somewhere that produces an invalid in-memory representation of a `LeafSchema` (i.e. an invalid tag) that's undefined behaviour (and something that should be performed/validated at the FFI boundary), and the catch-all arm probably wouldn't help  anyway because matches like this one are often compiled to jump tables and not necessarily `if-then-elif-else`s :(

danenbm: I think the unsafe code somewhere that produces an invalid in-memory representation would be something that results in having a C-compatible enum.  I don't really know how `build-bpf` works and how optimized our resultant build is, but I do know that we build a `cdylib` as well as a `lib`.  I'm not sure what a `cdylib` looks like inside, if/where it is used in Solana runtime, and if/where FFI boundaries exist.

Overall:
* I see that for a program Solana runtime looks up and calls the raw symbol: `pub unsafe extern "C" fn entrypoint(input: *mut u8) -> u64;` which makes me wonder how C-like the instruction code and data representations are.
* I see indications that C enums can cause issues when used in Rust previously, for example: https://github.com/rust-lang/rust/issues/36927.  Even though this issue is old it seems like it could still cause problems.
* And I did observe that @austbot  had an issue when one crate added an enum variant and it was used in some Geyser plugin code that was not updated.  This seems like it could be similar case to us because this data comes from a different account.

So even though what you said is making full sense, I'm still not fully comfortable yet that we would avoid all enum match issues by removing the catch-all arm.  Am I missing something, what do you think?  Can we just leave the catch-all match arm, does it hurt us just leaving it there?

austbot: the enums are coupled at this point so it would be okay to remove as long as we dont remove the usage of the v1 since we control it end to end the issue with geyser was

austbot (dismissed):
when conflicst resolved

bubblegum/program/src/lib.rs
danenbm: Yeah and leaf schema is defined in bubblegum so there could never be a similar case to the geyser case for this enum.  I took the catch all  match arm out in the next version of this PR.  Even in the geyser case there's a good chance a catch all arm would not have helped because the Rust compilation of the geyser plugin did not know about the V0.0.2 definition.

I talked about this with Alex more and there's some interesting problems that seem possible when gluing two pieces of Rust code together via cdylib loading.  Specifically he identified issues that can happen due to the way the geyser plugin has `#allow(improper_ctypes_definitions)]`.  Like one Rust may interpret memory in a way and the other Rust in another.

danenbm:
> looks good but there are some conflicts

Fixed merge conflicts

danenbm:
> Workflow `Verify package library` found differences when running `yarn api:gen` in the JS lib for `fixed-price-sale`. Please see the job for more details: https://github.com/metaplex-foundation/metaplex-program-library/actions/runs/2982192052.

Removed the change for fixed-price-sale because I don't want to mix this issue in with this PR.

austbot (approved):

alexandruag (approved):

PR #704, opened 2022-09-01
build: use solita api:gen to generate the client lib in the version workflow

jshiohaha:
If accepted, this change will do a few things: 

1. Consider all changed files. I previously filtered out files like `Cargo.toml` and `package.json`, but @samuelvanderwaal pointed out dependency changes might necessitate a version bump. Example workflow that failed: https://github.com/metaplex-foundation/metaplex-program-library/runs/7435095133
~2. Add `contractVersion` to each packages' `package.json` and update that value after updating a rust package version. We read the version directly from the relative `Cargo.toml`.~
3. Exit the workflow when no version changes are needed. Previously the workflow took an unnecessarily long time to run by installing Rust, Solana, and Anchor when they weren't needed.
4. Prevent using anchor/shank to regenerate IDL on version changes. Instead, we will update the version directly.

### Testing
🚧 TODO - still need to do some testing before final review + merge

### Context
Some of these changes originally included in https://github.com/metaplex-foundation/metaplex-program-library/pull/621, but aggregating them here.


jshiohaha:
Talked with @stegaBOB in chat, and he pointed out I can use each package's `.solitarc.js` in favor of the local config (yay) when updating the `version` field in an IDL. I will also play around with adding `yarn api:gen` to this workflow to ensure the library is updated, if needed.

1 thing I'm unsure of is how to cache each package's anchor version because that install is the longest part of the workflow. The challenge is that we don't know ahead of time which to try and restore and which to use reference because each package can use a different version of anchor.

jshiohaha:
@stegaBOB and I discussed a few options here - originally his suggestion was to try and use each package's `.solitarc.js` file to parse certain info. But, updating the IDL's version doesn't enforce library changes make it into the PR - this is because the contributor/reviewer must wait for the generate lib workflow to run / could miss the message saying the lib is out of sync.

so, the latest version here does not manually modify the IDL & instead generates the lib after bumping the version via `cargo release`. this only (feasibly) works if some of the dependencies used by solita are cached (e.g. anchor) - otherwise, install times will slow everything to a crawl.

attempt to cache shank/anchor here: https://github.com/metaplex-foundation/metaplex-program-library/pull/736

stegaBOB:
version patch

stegaBOB (commented):
version patch

stegaBOB (commented):
version patch

stegaBOB:
version patch

stegaBOB (commented):
version patch

stegaBOB:
version patch

stegaBOB (commented):
version bump

stegaBOB (approved):

Issue #678, opened 2022-08-15
[Bug]: Add remaining accounts to Auction House execute sale JS lib


0xPratik:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

JS SDK

### Issue description

I want to add remaining accounts to the execute sale function using the low level SDK but haven't found how can i go that.

### Priority this issue should have

Low (slightly annoying)

0xPratik:
@thlorenz This is related to Solita i think

thlorenz:
Hi thanks for the report. Could you please provide an example of what you're trying to do? I'm assuming you're referring to [this instruction](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/auction-house/js/src/generated/instructions/executeSale.ts#L100).

I looked into the [IDL genrated by anchor](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/auction-house/js/idl/auction_house.json#L931) and don't see any indicator in there of _remaining accounts_ that solita could pick up on to generate the correct code.

If I understand correctly then anchor has a notion of _implicit_ remaining accounts on the `Context` object?

```
/// Remaining accounts given but not deserialized or validated.
/// Be very careful when using this directly.
pub remaining_accounts: &'c [AccountInfo<'info>],
```

Would that mean that for each instruction their could potentially be _remaining accounts_ required even if that is not included in the IDL info? If so we should add a feature to solita to optionally provide those accounts.


If the above makes sense then let's continue discussion in the solita repo in a separate issue to work out what the generated code would look like and how it'd be used. We have to come up with a good solution if as I mentioned above potentially any instruction could have implicit _remaining accounts_ and thus those changes would affect each and every instruction rendered by solita.

0xPratik:
Yes you are correct its like optional accounts. all the deserialization of the remaining accounts takes place in the business logic of the program itself. In the case of executeSale instruction, it takes in the creator's address for royalty distribution without remaining accounts this instruction wont work

thlorenz:
Thought about how to solve this and since there is no way for solita to know which accounts need to be _writable_ and such the user would have to provide that info. Thus updating the solita generated code for (as an example) _executeSale_ could look like this:

```ts
export type ExecuteSaleInstructionAccounts = {
  buyer: web3.PublicKey
  seller: web3.PublicKey
  tokenAccount: web3.PublicKey
  tokenMint: web3.PublicKey
  [ .. many more .. ]
[...]

PR #705, opened 2022-09-01
chore: auction-house update SDK with latest solita

thlorenz:
## Summary

In order to fix the issue that there is no way to provide anchor remaining accounts I updated
solita to support this and then regenerated the SDK here. As a result the SDK is improved in
the following ways.

## Accounts

- include gpa builders
- expose `accountProviders` to be used by diagnostics/testing tools like amman
- account discriminators are exported for each account

## Instructions

- allow overriding program id
- allow overriding _builtin_ accounts
- allow providing `anchorRemaingAccounts` which when set will be pushed to the end of the
  array of accounts sent to the program
- instruction discriminators are exported for each instruction

## Deps Upgrades

Solita itself and related packages like beet and beet-solana where upgraded as well as web3.js.

FIXES: #678

### DIFF
--- a/auction-house/js/package.json
+++ b/auction-house/js/package.json
@@ -39,14 +39,14 @@
   "author": "Metaplex Maintainers <contact@metaplex.com>",
   "license": "Apache-2.0",
   "dependencies": {
-    "@metaplex-foundation/beet": "^0.1.0",
-    "@metaplex-foundation/beet-solana": "^0.1.1",
+    "@metaplex-foundation/beet": "^0.6.1",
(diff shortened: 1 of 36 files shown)


thlorenz:
The two failing checks are for the Rust part (Program code) which I did not touch.
I don't know why they are running at all + wouldn't know how to fix those (since I didn't change any code in there).

Maybe @samuelvanderwaal or @danenbm can help to figure out a solution.

blockiosaurus (approved):
The program test failures are due to known issues in our test execution (something to do with IOError timeouts). Otherwise looks good and approved!

stranzhay (approved):

PR #708, opened 2022-09-02
build: update path filters to include program/js subdirs

jshiohaha:
Update paths filters to only include rust subdirs for program workflows and js subdirs for sdk workflows. This should prevent program tests from running when only js changes are made, and vice versa.


thlorenz (approved):
Awesome! Looks like we had this setup inefficiently in a lot of places.
Thanks!

blockiosaurus (approved):

PR #710, opened 2022-09-02
fix: remaining accounts comparison

samuelvanderwaal:
Passing in too many accounts doesn't hurt anything and should still work as long as the correct ones are passed in in the correct order. Passing in _not enough_ accounts should fail because there may be cases where a bot is incorrectly spamming with not enough accounts.


samuelvanderwaal (commented):
version patch

febo (approved):

austbot (approved):

stegaBOB (approved):
Looks good to me

febo (approved):

stegaBOB (approved):

PR #711, opened 2022-09-03
Feat: remove unneeded rent sysvar from Token Metadata instructions

stegaBOB:
As with a number of other sysvars, `Rent` can be accessed using `Rent::get()` without having to pass the rent account in. This PR just removes the rents at the end of non deprecated instructions in a backwards compatible way. For the JS API, it marks most rent accounts as optional if it can without causing issues. This should also be tested with the testing lib pinned to an older token metadata version to ensure backwards compatibility. 

I tested this locally using the new .so file from this PR on the master branch and all the tests still pass. Logically it shouldn't be a breaking change and the tests seem to agree. 

PR #716 is this change without the ix builder changes to ensure backwards compatibility.


token-metadata/program/src/processor.rs
```diff
@@ -255,7 +257,7 @@ pub fn process_create_metadata_accounts_v2<'a>(
     let payer_account_info = next_account_info(account_info_iter)?;
     let update_authority_info = next_account_info(account_info_iter)?;
     let system_account_info = next_account_info(account_info_iter)?;
-    let rent_info = next_account_info(account_info_iter)?;
+    let _rent_info = next_account_info(account_info_iter)?;
```
austbot, line 260: remove

austbot (dismissed):

token-metadata/program/src/processor.rs
stegaBOB: The plan here was to only remove from the most recent handlers. Sorta as a "you should upgrade to the new handlers because it removes rent now" kinda thing. Like an incentive to upgrade I guess...

samuelvanderwaal (approved):

stegaBOB:
version bump

stegaBOB:
version patch

token-metadata/program/src/processor.rs
austbot: this is an unised var

febo: Line `260` has no effect since the variable is not used – it can be safely removed as @austbot pointed out.

stegaBOB: We have other unused vars too that were in place before this PR. Only reason I kept this one in is to add additional incentives to upgrade to the latest version of handlers. I know it can be safely removed.

austbot: How does this add incentive

stegaBOB, line 260:
```suggestion
```

stegaBOB: It doesn't. Removing.

austbot (approved):

samuelvanderwaal (approved):

PR #715, opened 2022-09-05
Bubblegum tests first iteration

alexandruag:
This a draft PR presenting an initial iteration of the code used to test Bubblegum, together with simple happy-path tests for several operations. We'll subsequently add  support for all operations, as well as more scenarios together with testing operations at higher volume to cover more corner cases. 

One significant change compared to other programs in MPL is that tests in this PR have been written as unit tests (vs integration tests for other contracts). More details about the distinction between the two categories are available [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html). Unit tests provide more flexibility when needed, as well as making it easier to work with reusable boilerplate. At the end of the day, if we decide to keep all testing done through integration tests, the conversion is straightforward.

There are some ongoing issues post transitioning to using the spl-compression lib from spl, which prevent the current tests from running properly. The PR will remain a draft while we figure those out.


alexandruag:
Closing in favor of #724.

PR #716, opened 2022-09-06
(Dont Merge!) Prove removing rent is backwards compatible

stegaBOB:
removing rent PR without the ix builder changes to make it possible to run CI on. This PR is ONLY for a proof that #711 is safe. All changes in this PR are also in 711.

### DIFF
--- a/token-metadata/program/src/processor.rs
+++ b/token-metadata/program/src/processor.rs
@@ -257,7 +257,7 @@ pub fn process_create_metadata_accounts_v2<'a>(
     let payer_account_info = next_account_info(account_info_iter)?;
     let update_authority_info = next_account_info(account_info_iter)?;
     let system_account_info = next_account_info(account_info_iter)?;
-    let rent_info = next_account_info(account_info_iter)?;
+    let _rent_info = next_account_info(account_info_iter)?;
 
     process_create_metadata_accounts_logic(
         program_id,
@@ -292,7 +292,6 @@ pub fn process_create_metadata_accounts_v3<'a>(
     let payer_account_info = next_account_info(account_info_iter)?;
     let update_authority_info = next_account_info(account_info_iter)?;
     let system_account_info = next_account_info(account_info_iter)?;
-    let rent_info = next_account_info(account_info_iter)?;
 
     process_create_metadata_accounts_logic(
         program_id,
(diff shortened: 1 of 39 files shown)


token-metadata/program/src/processor.rs
```diff
@@ -255,7 +257,7 @@ pub fn process_create_metadata_accounts_v2<'a>(
     let payer_account_info = next_account_info(account_info_iter)?;
     let update_authority_info = next_account_info(account_info_iter)?;
     let system_account_info = next_account_info(account_info_iter)?;
-    let rent_info = next_account_info(account_info_iter)?;
+    let _rent_info = next_account_info(account_info_iter)?;
```
austbot, line 260: just remove

token-metadata/program/src/utils.rs
```diff
@@ -171,13 +171,12 @@ pub fn assert_initialized<T: Pack + IsInitialized>(
...
     program_id: Pubkey,
     new_account_info: &AccountInfo<'a>,
-    rent_sysvar_info: &AccountInfo<'a>,
     system_program_info: &AccountInfo<'a>,
     payer_info: &AccountInfo<'a>,
     size: usize,
     signer_seeds: &[&[u8]],
 ) -> ProgramResult {
-    let rent = &Rent::from_account_info(rent_sysvar_info)?;
+    let rent = &Rent::get()?;
```
austbot, line 179: I did this in the mint branch lol

token-metadata/program/src/processor.rs
```diff
@@ -1243,9 +1234,19 @@ pub fn process_utilize(
     let token_program_account_info = next_account_info(account_info_iter)?;
     let _ata_program_account_info = next_account_info(account_info_iter)?;
     let _system_account_info = next_account_info(account_info_iter)?;
-    let _rent_info = next_account_info(account_info_iter)?;
+    // consume the next account only if it is Rent
+    let approved_authority_is_using = if account_info_iter
```
austbot, line 1238: was it not the last account in this casE?

austbot (changes requested):

token-metadata/program/src/processor.rs
stegaBOB: The plan here was to only remove from the most recent handlers. Sorta as a "you should upgrade to the new handlers because it removes rent now" kinda thing

token-metadata/program/src/processor.rs
stegaBOB: It was not. There are two optional accounts after this.

PR #722, opened 2022-09-06
Adding some minor refactors to TM TokenMetadataAccount trait

blockiosaurus:
Making TokenMetadataAccount trait more flexible.


samuelvanderwaal (approved):

token-metadata/program/src/state.rs
```diff
@@ -89,11 +89,16 @@ pub const USE_AUTHORITY_RECORD_SIZE: usize = 18; //8 byte padding
...
 pub const COLLECTION_AUTHORITY_RECORD_SIZE: usize = 11; //10 byte padding
 
-pub trait TokenMetadataAccount {
+pub trait TokenMetadataAccount: BorshDeserialize {
     fn key() -> Key;
 
     fn size() -> usize;
 
+    fn is_correct_account_type(data: &[u8], data_type: Key, data_size: usize) -> bool {
+        (data[0] == data_type as u8 || data[0] == Key::Uninitialized as u8)
```
alexandruag, line 98: As a general mention, I think the `as` operator should be avoided whenever possible; contrary to the core philosophy of the language, it quietly allows all sort of potentially dangerous casts, including (but not limited) to stuff like lossy narrowing of types. While a bit more verbose, I think it's better to perform conversions via specific methods (or implementations of traits like `From/Into`; or fallible stuff like `try_...` which return results if it's possible in any way to have failures).

token-metadata/program/src/state.rs
```diff
@@ -102,19 +107,19 @@ pub trait TokenMetadataAccount {
...
 
-    fn safe_deserialize<T: BorshDeserialize>(mut data: &[u8]) -> Result<T, BorshError> {
-        if !is_correct_account_type(data, Self::key(), Self::size()) {
+    fn safe_deserialize(mut data: &[u8]) -> Result<Self, BorshError> {
+        if !Self::is_correct_account_type(data, Self::key(), Self::size()) {
             return Err(BorshError::new(ErrorKind::Other, "DataTypeMismatch"));
         }
 
-        let result: T = T::deserialize(&mut data)?;
+        let result: Self = Self::deserialize(&mut data)?;
```
alexandruag, line 115: Just wondering, does `let result = Self::deserialize(&mut data)?;` also work here  (and in the similar situation below)?

danenbm (changes requested):

token-metadata/program/src/state.rs
```diff
@@ -102,19 +114,19 @@ pub trait TokenMetadataAccount {
...
+        let result = Self::deserialize(&mut data)?;
 
         Ok(result)
     }
 
-    fn from_account_info<T: BorshDeserialize>(a: &AccountInfo) -> Result<T, ProgramError>
+    fn from_account_info(a: &AccountInfo) -> Result<Self, ProgramError>
 where {
-        let ua: T = Self::safe_deserialize(&a.data.borrow_mut())
+        let ua: Self = Self::safe_deserialize(&a.data.borrow_mut())
```
danenbm, line 129: Nit: was this `ua:  Self` still needed now that safe deserialize specifically returns `Result<Self, BorshError>`?

danenbm (approved):
Looks great to me!

PR #728, opened 2022-09-09
Remove mint request mechanism in bubblegum

danenbm:
- Simplify `mint_v1` to use a payer and a tree delegate.

**UPDATE: Had to add back in code to properly track total mint amount, since i removed it when `mint_request` was removed**


austbot (approved):

alexandruag (dismissed):

austbot (commented):
version patch

austbot (dismissed):
version patch

austbot (approved):

danenbm (commented):
version patch

bubblegum/program/src/lib.rs
```diff
@@ -935,69 +827,39 @@ pub mod bubblegum {
...
-            );
-            metadata_auth.insert(mint_authority.key());
+        require!(
+            incoming_tree_delegate == tree_creator || incoming_tree_delegate == tree_delegate,
+            BubblegumError::TreeAuthorityIncorrect,
+        );
+
+        if !authority.contains_mint_capacity(1) {
+            return Err(BubblegumError::InsufficientMintCapacity.into());
         }
```
kespinola, line 853:
```suggestion
       require(authority.contains_mint_capacity(1), BubblegumError::InsufficientMintCapacity);
```

NB: Is this not equivalent?

bubblegum/program/src/lib.rs
```diff
@@ -850,6 +848,12 @@ pub mod bubblegum {
             BubblegumError::TreeAuthorityIncorrect,
         );
 
+        if !authority.contains_mint_capacity(1) {
```
alexandruag, line 851: The `spl-compression` program also does some checks during appends, and will return an error if a tree is full (for example, when eventually reaching [this](https://github.com/solana-labs/solana-program-library/blob/6bcd051d384f4ebcf8d6933291d8a3eacef412bf/libraries/concurrent-merkle-tree/src/concurrent_merkle_tree.rs#L214) line. Was just wondering, do we need to do the parallel accounting? (or, for example, are there circumstances where the max capacity would be defined as less than the theoretical max for a tree?)

bubblegum/program/src/lib.rs
```diff
@@ -850,6 +848,12 @@ pub mod bubblegum {
...
+            return Err(BubblegumError::InsufficientMintCapacity.into());
+        }
+
+        authority.increment_mint_count();
```
alexandruag, line 855: Hmm, maybe it's better to move this increment after `process_mint_v1` completes successfully, just so we don't also end up counting mint reqs that end up leading to errors.

bubblegum/program/src/lib.rs
austbot: yes but is more concise

bubblegum/program/src/lib.rs
austbot: we do so we can return a better error than spl-account-compression

bubblegum/program/src/lib.rs
danenbm: To me it seems about equivalent, I'll probably move towards require statement in later PR.  Thanks!

bubblegum/program/src/lib.rs
danenbm: Whole transaction and state change will be reverted if there's a failure.

PR #737, opened 2022-09-12
Add shank install and wrapper action for program dependencies

jshiohaha:
* We now have a cache-able action to install `shank-cli` like linux deps, rust, and solana, anchor, etc.
* We now also have a wrapper around all common program build dependencies. We can now install with something as simple as (1) instead of (2). However, we still have flexibility to exclude certain dependencies we don't want. In the future, we can also extend to specify versions, or fallback to a default.

**(1) possibly new way of installing dependencies**

```
- uses: ./.github/actions/install-common-build-deps
```

**(2) old way of installing dependencies**
```
- uses: ./.github/actions/install-linux-build-deps
- uses: ./.github/actions/install-rust
  with:
    toolchain: ${{ env.RUST_TOOLCHAIN }}
- uses: ./.github/actions/install-solana
- uses: ./.github/actions/install-anchor 
```

### DIFF
--- a/.github/actions/install-anchor/action.yml
+++ b/.github/actions/install-anchor/action.yml
@@ -1,3 +1,4 @@
+# if referencing from a workflow, ANCHOR_VERSION environment variable must be defined
 name: Setup Anchor CLI
 description: 'Setup Anchor CLI'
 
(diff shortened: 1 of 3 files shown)


stegaBOB (approved):
LGTM!

austbot (approved):

PR #749, opened 2022-09-19
Downgrades SPL ASSOC

austbot:
This change supports blockbuster, by downgrading spl associated token account to be more compatible with anchor


danenbm (approved):
lgtm

PR #752, opened 2022-09-20
Bubblegum tests: refactor aux logic and add proofs

alexandruag:
Posting a separate PR with the most recent changes to the test auxiliary logic and the use of `MerkleTree` from `spl-merkle-tree-reference` to locally generate proofs, as this influences the way incoming tests will be written/interpreted. A bit more info is present in the individual commit messages.


bubblegum/program/tests/utils/tx_builder.rs
```diff
@@ -0,0 +1,315 @@
...
+    // testing purposes).
+    pub need_proof: Option<u32>,
+    // This member holds data of a custom type that can be specific to each kind
+    // of transaction.
+    pub inner: V,
+}
+
+pub trait OnSuccessfulTxExec {
+    fn on_successful_execute(&mut self) -> Result<()>;
+}
```
danenbm, line 49: Ok cool I see how this works.

bubblegum/program/tests/utils/tree.rs
```diff
@@ -0,0 +1,631 @@
...
+            nonce: args.nonce,
+            index: args.index,
+        };
+
+        // Cloning to avoid issues with the borrow checker when passing `&mut args`
+        // to the builder below.
+        let owner = clone_keypair(&args.owner);
+        let new_owner = clone_keypair(new_leaf_owner);
+
+        let need_proof = Some(args.index);
```
danenbm, line 464: So with most if not all of these transactions looks like you are essentially hardcoding them to to needing a proof, right?  That is what the `Some()` variant indicates.

So it seems like that is not strictly true for all transfer instructions but is probably true for our default test cases now that there are more than one leaf?  Should there be a mechanism to run it with and without proofs?

bubblegum/program/tests/simple.rs
```diff
@@ -21,76 +24,83 @@ const MAX_BUF_SIZE: usize = 64;
...
+    let payer = context.payer();
+
+    let cfg = tree.read_tree_config().await.unwrap();
+    assert_eq!(cfg.tree_creator, payer.pubkey());
+    assert_eq!(cfg.tree_delegate, payer.pubkey());
+    assert_eq!(cfg.total_mint_capacity, 1 << MAX_DEPTH);
+    assert_eq!(cfg.num_minted, DEFAULT_NUM_MINTS);
 }
 
 #[tokio::test]
```
danenbm, line 54: Shouldn't we still add some `asserts` even in the happy path tests that validate the on-chain data is correct?

For example, what if someone made a code change to bubblegum that prevented the `spl-account-compression` CPI from happening but the bubblegum instruction returned `Ok()`?  To prevent that bug wouldn't we want to get the merkle root from the merkle tree, calculate the merkle root of our updated local proof tree, and assert that they match?

alexandruag: Yes that's something I want to add in another PR; for now wanted to bring these changes in since they touch a bunch of stuff that influences how tests are actually written.

bubblegum/program/tests/utils/tree.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    pub fn tree_pubkey(&self) -> Pubkey {
+        self.merkle_tree.pubkey()
+    }
+
+    pub fn authority(&self) -> Pubkey {
+        Pubkey::find_program_address(&[self.tree_pubkey().as_ref()], &mpl_bubblegum::id()).0
+    }
+
+    pub fn mint_authority_request(&self, authority: &Pubkey) -> Pubkey {
```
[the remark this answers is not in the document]
alexandruag: Good point, will remove!

bubblegum/program/tests/simple.rs
```diff
@@ -4,15 +4,18 @@ use solana_program_test::tokio;
...
-    clone_keypair,
     context::{BubblegumTestContext, DEFAULT_LAMPORTS_FUND_AMOUNT},
-    LeafArgs, Result, Tree,
+    tree::Tree,
+    LeafArgs, Result,
 };
 
 // Test for multiple combinations?
-const MAX_DEPTH: usize = 20;
+const MAX_DEPTH: usize = 14;
```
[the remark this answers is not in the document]
alexandruag: I lowered it because the default construction used in this test code for the proof `MerkleTree` here populates the entire tree, and for `1 << 20` it took quite a while.

bubblegum/program/tests/utils/context.rs
```diff
@@ -134,17 +134,28 @@ impl BubblegumTestContext {
     // The owner of the tree and leaf is `self.payer()`.
     pub async fn default_create_and_mint<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize>(
```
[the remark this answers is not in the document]
alexandruag: Well it's still the default (now parametrized :D) procedure; I can def rename it for more clarity.

bubblegum/program/tests/utils/tree.rs
alexandruag: It's not as much hard-coded as pre-configured :D, but it's meant to be an implementation detail behind the scenes; I can def add some helper method and/or parameter somewhere to make it easier to tweak the value between some and none, and that can easily be added on top at any time. Also, it seems like we need proofs for any tree that has more than just 1 leaf.

austbot (commented):
This one is big im making my way through it

alexandruag:
Rebased on top of the latest changes. Yeah that kind of refactoring is def possible; might be able to abstract Anchor away as well if/when it will make things more useful in the future.

bubblegum/program/tests/utils/tree.rs
danenbm: It could send None if there's only one leaf then, which you know from your proof tree?

bubblegum/program/tests/simple.rs
danenbm: Ok sounds good

danenbm: ok sounds good

alexandruag:
Did one more rebase; also removed the `mint_authority_request` helper fn since, as Michael mentioned, it was no longer useful. Did not end up renaming `fn default_create_and_mint` because I think this name is truer to its nature, as it both creates the tree and mints a number of leaves using a bunch of preset default value for the inner details.

austbot (approved):

danenbm (approved):

PR #754, opened 2022-09-21
chore: remove Rust workspace

samuelvanderwaal:
This PR removes Rust work spaces from the repo to decouple project dependencies from each other and make things more flexible. It also modifies and cleans up some tests in order to get CI to pass and makes some CI fixes and improvements.

The big change here regarding BPF tests is to replace the `../../target/deploy` folder with a custom `test-programs` folder in the project root and add a build script that automates installing these dependent programs.


.github/actions/install-linux-build-deps/action.yml
```diff
@@ -9,22 +9,16 @@ runs:
         sudo apt-get update
```
thlorenz, line 9: Wondering changes in this file were needed in relation to workspaces or if you just optimized the deps we need to install here.
For the record this action was originally adapted from an install script and I wasn't aware at the time it could be optimized.

.github/actions/install-linux-build-deps/action.yml
```diff
@@ -9,22 +9,16 @@ runs:
...
         sudo apt-get install -y libssl1.1 --allow-unauthenticated
       shell: bash
 
-    - name: apt-get install dev tools 
+    - name: apt-get install libudev
       run: |
-        sudo apt-get install -y libudev-dev
-        sudo apt-get install -y binutils-dev
-        sudo apt-get install -y libunwind-dev
+        sudo apt-get update
```
thlorenz, line 21: Shouldn't this run just _once_ and _before_ all other tasks?

.github/workflows/integration-hydra.yml
```diff
@@ -58,29 +60,33 @@ jobs:
...
           cache_id: sdk-hydra
           working_dir: ./hydra/js
           build_token_metadata: true
 
       # Run integration test
       - name: start-local-test-validator
         working-directory: ./hydra/js
-        run: yarn run api:gen && yarn run build && DEBUG=amman* yarn amman:start && DEBUG=mpl* yarn test && yarn run amman:stop
+        run:
+          yarn run api:gen && yarn run build && DEBUG=amman* yarn amman:start && DEBUG=mpl* yarn
```
thlorenz, line 82: This reformatting looks odd as we got one command across multiple lines. Not sure how/if this could be improved.

.github/workflows/program-auction-house.yml
```diff
@@ -123,4 +123,4 @@ jobs:
         run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../target/deploy/ --test ${{ env.TEST }} -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs/ --test ${{ env.TEST }} -- --nocapture
```
thlorenz, line 126: I'd prefer to run tests in series in CI as otherwise it gets harder to identify an issue for 2 reasons:

- test output of different tests is intermingled
- tests might affect each other in odd ways

Also given that running the tests takes a very small percentage of the time compared to the setup needed before that I believe the gain doesn't outweigh the potential issues here.

.github/workflows/program-auctioneer.yml
```diff
@@ -123,4 +123,4 @@ jobs:
         run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../target/deploy/ --test ${{ env.TEST }} -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs/ --test ${{ env.TEST }} -- --nocapture
```
thlorenz, line 126: Ditto running tests in parallel

.github/workflows/program-bubblegum.yml
```diff
@@ -117,6 +117,6 @@ jobs:
       - name: test-bubblegum-program
         id: run_bubblegum_test
         working-directory: ./bubblegum/program
-        run: |          
+        run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../target/deploy/ --test ${{ env.TEST }} -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs/ --test ${{ env.TEST }} -- --nocapture
```
thlorenz, line 122: Ditto running tests in parallel

.github/workflows/program-candy-machine.yml
```diff
@@ -74,4 +76,4 @@ jobs:
         run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../target/deploy/ -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs/ -- --nocapture
```
thlorenz, line 79: Ditto running tests in parallel

.github/workflows/program-fixed-price-sale.yml
```diff
@@ -58,13 +60,21 @@ jobs:
...
+        run: |
+          solana program dump p1exdMJcjVao65QdewkaZRUnU6VPSXhus9n2GzWfh98 test-programs/mpl_metaplex.so
+
       # Run test
       - name: test-fixed-price-sale-program
         id: run_test
         working-directory: ./fixed-price-sale/program
         run: |
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf -- --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs
```
thlorenz, line 80: Ditto running tests in parallel

.github/workflows/program-gumdrop.yml
```diff
@@ -69,4 +71,4 @@ jobs:
         run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../target/deploy/ -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs/ -- --nocapture
```
thlorenz, line 74: Ditto running tests in parallel

.github/workflows/program-nft-packs.yml
```diff
@@ -57,14 +57,22 @@ jobs:
...
+          solana program dump p1exdMJcjVao65QdewkaZRUnU6VPSXhus9n2GzWfh98 test-programs/mpl_metaplex.so
+
+      # Run tests
       - name: test-nft-packs-program
         id: run_test
         working-directory: ./nft-packs/program
         run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs -- --nocapture
```
thlorenz, line 78: Ditto running tests in parallel

.github/workflows/program-token-entangler.yml
```diff
@@ -69,6 +69,6 @@ jobs:
         id: run_test
         working-directory: ./token-entangler/program
         run: |
-          cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs -- --nocapture
```
thlorenz, line 74: Ditto running tests in parallel

.github/workflows/program-token-metadata.yml
```diff
@@ -118,4 +118,4 @@ jobs:
         run: |
           cargo +${{ env.RUST_TOOLCHAIN }} test -- --nocapture --test-threads 1
           cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --version
-          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../target/deploy/ --test ${{ env.TEST }} -- --nocapture --test-threads 1
+          cargo +${{ env.RUST_TOOLCHAIN }} test-bpf --bpf-out-dir ../../test-programs/ --test ${{ env.TEST }} -- --nocapture
```
thlorenz, line 121: Ditto running tests in parallel

auction-house/program/src/execute_sale/mod.rs
```diff
@@ -1349,8 +1349,7 @@ fn auctioneer_execute_sale_logic<'c, 'info>(
                     &token_account.key(),
                     &seller.key(),
                     &[],
-                )
-                    .unwrap(),
+                )?,
```
thlorenz, line 1352: These are unrelated changes to workspaces right?
Just double checking ... probably should have been in separate PR unless it was needed to please clippy or whatever

auction-house/program/tests/execute_sale.rs
```diff
@@ -17,7 +17,7 @@ use mpl_testing_utils::{
     solana::{airdrop, create_associated_token_account, transfer},
     utils::Metadata,
 };
-use solana_sdk::signer::Signer;
+use solana_sdk::{commitment_config::CommitmentLevel, signer::Signer};
```
thlorenz, line 20: These changes here and a lot more below seem totally unrelated to _removing Rust workspaces_ as are a lot in the files below.
So I'm a bit confused about those.

fixed-price-sale/program/Cargo.toml
```diff
@@ -29,3 +30,4 @@ solana-program = "~1.9.28"
 solana-sdk = "~1.9.28"
 spl-associated-token-account = "~1.0.3"
 mpl-testing-utils= {path="../../core/rust/testing-utils" }
+mpl-metaplex = "0.0.3"
```
thlorenz, line 33: Why is this needed? This module is super outdated .. I recently removed it from auction house.

thlorenz (changes requested):
Looks good in general, just had a question about a few things and obviously I wouldn't be aware of things that were missed (since they don't show in the diff).
My main concern is about all these code updates, but maybe they were needed due to some lib upgrades?

.github/workflows/program-auction-house.yml
samuelvanderwaal: Tests are supposed to be independent. What are the potential ways tests might affect each other?

auction-house/program/tests/execute_sale.rs
samuelvanderwaal: I thought the Banks Client defaulting to using *Finalized* might be the issue behind some of the test time out issues we've seen in CI. Changing the commitment to *Confirmed* should resolve txs faster.

fixed-price-sale/program/Cargo.toml
samuelvanderwaal: `mpl-metaplex` is used by some Fixed Price Sale tests.

.github/workflows/program-auction-house.yml
thlorenz: I am not sure, I suppose that would only be a problem if they were running against the same validator or similar.

However the first point is actually at least equally important .. running tests in parallel leads to output that makes it very hard to piece together information of a failing test

thlorenz: Actually running tests in parallel also could lead to timing issues due to the host box getting overloaded and thus tests timing out esp. considering recent block hashes and such.

fixed-price-sale/program/Cargo.toml
thlorenz: OK, actually I found the same to be true in auction house, but there it was using a struct that we now expose via mpl-token-metadata. Maybe something like this is possible here as well?

.github/workflows/program-auction-house.yml
samuelvanderwaal: Ok, I'll fix this, thanks.

auction-house/program/src/execute_sale/mod.rs
samuelvanderwaal: I neurotically fix things like this when I find them, but yeah would have been better as a separate PR.

fixed-price-sale/program/Cargo.toml
samuelvanderwaal: Rewriting Fixed Price Sale or a large refactor of the tests is out-of-scope for this PR and would be better in a separate one. This is the smallest change necessary to get these tests to pass.

.github/actions/install-linux-build-deps/action.yml
samuelvanderwaal: There were some changes required to the linux install because we were using BuildJet's Ubuntu 22.04 runner which had some different deps requirements than whatever the version of Ubuntu used previously was.

fixed-price-sale/program/Cargo.toml
thlorenz: Ah ok in that case we can just leave it in. I thought it was a simple `use` statement that'd need to be replaced. If it would be more tricky I agree that it should be in a separate PR.

.github/workflows/sec3-test.yml
```diff
@@ -1,24 +1,24 @@
...
-        uses: sec3dev/pro-action@v1
-        with:
-          sec3-token: ${{ secrets.SEC3_TOKEN }}
-          path: hydra/program
-          hide-report-link: true
-      - name: Upload Sarif Report
-        uses: github/codeql-action/upload-sarif@v1
-        with:
-          sarif_file: sec3-report.sarif
+# name: Sec3
```
thlorenz, line 1: Should probably put a comment with a reason for disabling and a note on how we could re-enable this properly.
Unless we totally gave up on this in which case we should just remove the file instead

samuelvanderwaal: It fails in CI frequently and Austin said we can remove it so I'll just delete it. We have git history if we need to add it back later.

thlorenz (dismissed):

blockiosaurus (approved):
I'm not crazy about putting so many unrelated changes in one PR but everything looks correct.

samuelvanderwaal:
> I'm not crazy about putting so many unrelated changes in one PR but everything looks correct.

Understandable, but let me refer you to this diagram so you can understand my last couple weeks.

Untitled Diagram drawio

thlorenz (approved):

Issue #755, opened 2022-09-21
[Bug]: Rent account is declared as optional, but is mandatory


tevdoradze:
### Which package is this bug report for?

token-metadata

### Which Type of Package is this bug report for?

JS SDK

### Issue description

This PR https://github.com/metaplex-foundation/metaplex-program-library/pull/711 made rent account optional. I was using the latest version of CreateMetadataAccountV3 [CreateMetadataAccountV3](https://github.com/metaplex-foundation/metaplex-program-library/blob/a5b1bb37ddba8ac682db3baa856190ea2402a09f/token-metadata/js/src/generated/instructions/CreateMetadataAccountV3.ts#L115), filled all mandatory accounts and end up with an error.

This can be really confusing for people using the method, @austbot  @stegaBOB could we reverse the PR?

It seems weird to have a mandatory account marked as optional, multiple people are reporting this on Discord.

### Relevant log output

```
'Program log: Instruction: Create Metadata Accounts v3',
    'Program log: Error: NotEnoughAccountKeys',
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s consumed 3890 of 797033 compute units',
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s failed: insufficient account keys for instruction'
   ````
### Priority this issue should have

Medium (should be fixed soon)

samuelvanderwaal:
It's optional in v1.4 which is on `devnet` but not `mainnet`. The problem here is we released a new version of the JS libs before going to mainnet (which normally isn't an issue, but in this case it is). Once 1.4 is on mainnet this shouldn't be an issue.

tevdoradze:
> It's optional in v1.4 which is on `devnet` but not `mainnet`. The problem here is we released a new version of the JS libs before going to mainnet (which normally isn't an issue, but in this case it is). Once 1.4 is on mainnet this shouldn't be an issue.
> 
> Downgrade metaplex-foundation/js version to unblock yourself, or pass in the rent account even though it's optional.

Could we have a way to track the commit that's used on mainnet? Either with a tag or by only merging on master when we deploy on mainnet?

samuelvanderwaal:
[...]

PR #756, opened 2022-09-21
[#755] reverting optional on CreateMasterEditionV3

ohaddahan:
#755  Didn't see solita specific tests, if got some pointers would love to follow and add as needed.


samuelvanderwaal:
Thanks for the PR but this is actually an optional account in v1.4, see my comments on the original issue.

PR #757, opened 2022-09-21
Adds Eq types to aid in testing of blockbuster parsers

austbot:
the blockbuster parser tests need to check equality of types, without haveing to deeply nest equality checks we can do what is in this pr.


alexandruag (approved):
fmt strikes again

alexandruag (approved):

blockiosaurus (approved):

PR #759, opened 2022-09-21
disallow switching FROM hiddensettings to normal CM

MarkSackerberg:
Switching back and forth between having config lines to no config lines is already disallowed.
Nevertheless trying to remove hiddensettings also causes issues and should not be allowed. 

If you have a candy machine and remove the hiddensettings trying to mint will cause:
> Program log: panicked at 'index out of bounds: the len is 713 but the index is 144842', src/processor/mint.rs:768:12
Switching back is not possible and therefore any candymachine where someone tries to remove hiddensettings is basically bricked.

This change should disallow removing hiddensettings, therefore those candy machines can not be bricked anymore.


samuelvanderwaal (approved):

candy-machine/program/src/processor/update.rs
```diff
@@ -61,6 +60,13 @@ pub fn handle_update_candy_machine(
         return err!(CandyError::CannotSwitchToHiddenSettings);
     }
 
+    if candy_machine.data.items_available > 0
```
febo, line 63: I think we can remove this condition `candy_machine.data.items_available > 0`; otherwise you could technically set the `items_available` to `0` and then switch from hidden settings. There is no scenario where you can switch from hidden settings, given that the account size does not have space for config lines.

febo (changes requested):

candy-machine/program/src/processor/update.rs
samuelvanderwaal: I agree

MarkSackerberg: True. Thanks for the input.
Done.

candy-machine/program/src/errors.rs
```diff
@@ -108,4 +108,6 @@ pub enum CandyError {
     NoUnlockWithNFTsStillFrozen,
     #[msg("Setting a sized collection requires the collection metadata to be mutable.")]
     SizedCollectionMetadataMustBeMutable,
+    #[msg("Cannot Switch back from Hidden Settings after items available is greater than 0")]
```
febo, line 111: We need to update the error message to something like `"Cannot switch from hidden settings"`.

febo (changes requested):

candy-machine/program/src/errors.rs
samuelvanderwaal: @MarkSackerberg one more small change

MarkSackerberg: Done.

febo (dismissed):

samuelvanderwaal:
I'll merge and deploy this later this week.

samuelvanderwaal (changes requested):
The CI revealed a couple more things that need to be fixed:

* run `cargo fmt` on the code base as there's a formatting issue in `processor.rs`
* cd in the `candy-machine/js` directory, run `yarn install` and `yarn gen:api` to regenerate the JS lib 
* commit and push the above changes

samuelvanderwaal (approved):

PR #762, opened 2022-09-22
fix candidate

konrads:


samuelvanderwaal:
The system program rent account was removed from the instructions in v1.4, which hasn't landed on mainnet yet. It will likely go to mainnet today, but in the meantime just use 1.3.6 for instructions that go to mainnet.

PR #764, opened 2022-09-22
Enforce rent pubkey for createMetaDataAccountV3 and createMasterEditi…

arvinaik:
…onV3


samuelvanderwaal:
Rent pubkey is an optional account in Token Metadata v1.4. If your JS isn't working downgrade your library until this feature hits `mainnet` (probably this week). 

Also note that we don't accept PRs to the autogenerated JS code because any changes would just get erased next time the code gets generated.

PR #765, opened 2022-09-22
Trifle implementation

nilz3ro:
WIP


token-metadata/program/src/escrow/create_escrow_account.rs
```diff
@@ -67,15 +69,20 @@ pub fn process_create_escrow_account(
...
+    msg!("trying authority");
     let maybe_authority_info: Option<&AccountInfo> = if is_using_authority {
+        msg!("creator authority");
         Some(next_account_info(account_info_iter)?)
     } else {
+        msg!("token authority");
         None
     };
 
+    msg!("asserting owners");
```
nilz3ro, line 85: We can take these comments out, this was for debugging

PR #771, opened 2022-09-26
Bubblegum tests: baseline redeem, cancel, and decompress

alexandruag:
This PR is created (for now) on top of https://github.com/metaplex-foundation/metaplex-program-library/pull/752, so the first four commits are duplicated. It introduces the baseline tests for (cancel) redeem, together with decompress. A few checks are omitted for now due to some weird behaviour that needs to be sorted out (will update the PR afterwards).


alexandruag:
#752 got merged a few days ago, and this finally got rebased and in better shape for review.

austbot (approved):
This looks great to me pending a few items as nits or comments

bubblegum/program/build.rs
```diff
@@ -11,20 +11,27 @@ fn main() {
         Path::new("../../target/deploy/GRoLLzvxpxxu2PGNJMMeZPyMxjAUH9pKqxGXV9DGiceU.so");
     let spl_wrapper_so_path =
         Path::new("../../target/deploy/WRAPYChf58WFCnyjXKJHtrPgzKXgHp6MD9aVDqJBbGh.so");
+    let mpl_tm_so_path = Path::new("../../target/deploy/mpl_token_metadata.so");
```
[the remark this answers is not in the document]
alexandruag: Currently the `build.rs` scripts in Rust have no (straightforward) way of knowing whether it's a test build or not, so this runs on every build, but nothing really happens if that path already exists. If we ever need/want to only run stuff like this for tests, there are ways to orchestrate that, but they're more convoluted (and generally involve breaking things into multiple crates, with test dependencies).

bubblegum/program/build-token-metadata.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Simple script to build token-metdata and move it to the root
+# `target/deploy` folder to load for banks client tests. This is
+# mean to be invoked via `build.rs` where the current working dir
+# is already set properly.
+
+set -e
+
+cd ../../token-metadata
+cargo build-bpf
```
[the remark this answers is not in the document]
alexandruag: oh that's totally right ... will incorporate

alexandruag:
Rebased on top of latest upstream.

bubblegum/program/tests/simple.rs
```diff
@@ -54,54 +65,211 @@ async fn test_create_tree_and_mint_passes() {
...
+
+        assert!(meta.collection.is_none());
+        assert!(meta.uses.is_none());
+        assert!(meta.collection_details.is_none());
+
+        // Test master edition account.
+        let me_key = find_master_edition_account(&mint_key).0;
+        let me_account = tree.read_account(me_key).await.unwrap();
+        let me: MasterEditionV2 =
+            MasterEditionV2::safe_deserialize(me_account.data.as_slice()).unwrap();
```
danenbm, line 265: Similar to the Metadata I believe you could use `MasterEditionV2::from_account_info()` and get one extra check.

bubblegum/program/tests/simple.rs
```diff
@@ -54,54 +65,211 @@ async fn test_create_tree_and_mint_passes() {
...
+        //     assert_eq!(c1.address, c2.address);
+        //     assert_eq!(c1.verified, c2.verified);
+        //     assert_eq!(c1.share, c2.share);
+        // }
+
+        assert!(!meta.primary_sale_happened);
+        assert!(!meta.is_mutable);
+
+        // TODO: Is the edition nonce field predictable so we can/should test for a specific
+        // value there as well?
```
danenbm, line 253: I can look at this tomorrow or you can ask @samuelvanderwaal he would probably know off the top of his head.

bubblegum/program/tests/simple.rs
```diff
@@ -54,54 +65,211 @@ async fn test_create_tree_and_mint_passes() {
...
+        // assert_eq!(meta_creators.len(), leaf.metadata.creators.len());
+        // // Can't compare directly as they are different types for some reason.
+        // for (i, c1) in meta_creators.iter().enumerate() {
+        //     let c2 = leaf.metadata.creators[i];
+        //     assert_eq!(c1.address, c2.address);
+        //     assert_eq!(c1.verified, c2.verified);
+        //     assert_eq!(c1.share, c2.share);
+        // }
+
+        assert!(!meta.primary_sale_happened);
```
danenbm, line 249: I am assuming you are testing pretty much every field for these account data types.  I don't think I need to double check your coverage but lmk if I should.

danenbm (changes requested):

bubblegum/program/tests/simple.rs
alexandruag: I've switched to an approach where I instantiate the entire expected struct such that we are sure no fields etc. are missing.

alexandruag:
Pushed an extra commit to address comments; I get some weird errors when rebasing on top of latest master so that will need a bit more work.

alexandruag:
Pushed an extra commit to address the breakages post compression and noop bump.

bubblegum/program/tests/simple.rs
```diff
@@ -54,54 +65,211 @@ async fn test_create_tree_and_mint_passes() {
...
-    let leaf = leaves.last_mut().unwrap();
 
-    tree.verify_creator(leaf, &context.default_creators[0])
-        .await
-        .unwrap();
+    for leaf in leaves.iter_mut() {
+        tree.verify_creator(leaf, &context.default_creators[0])
+            .await
+            .unwrap();
+    }
```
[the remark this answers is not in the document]
danenbm: Yes I think comments would be great solution.

bubblegum/program/tests/simple.rs
danenbm: Good idea

alexandruag:
Rebased again and also added the comments regarding the on-chain tree root and expected value being checked implicitly by the operations based on `TxBuilder::execute` to individual tests in `simple.rs`. As a thought, when breaking tests into multiple files later, we can probably keep the reminder as module/file level documentation.

bubblegum/program/build.rs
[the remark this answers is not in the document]
danenbm: I think we just need a linear task (not needing to be resolved in this PR) to fix the issue where the build.rs script prevents our automation from publishing to crates.io.

bubblegum/program/download-compression-programs.sh
```diff
@@ -4,19 +4,17 @@
 set -e
 
 MPL_ROOT=$(git rev-parse --show-toplevel)
-mkdir -p $MPL_ROOT/target/deploy
+mkdir -p $MPL_ROOT/test-programs
 
 mkdir -p solana_program_library
-curl -LkSs https://api.github.com/repos/solana-labs/solana-program-library/tarball/5f189d1fc0722a619c9b850f2769f85fccb5baea | tar -xz --strip-components=1 -C ./solana_program_library
+curl -LkSs https://api.github.com/repos/solana-labs/solana-program-library/tarball/b0d84fc4 | tar -xz --strip-components=1 -C ./solana_program_library
```
danenbm, line 10: This probably should be updated to 6e8179484413cd8a50ee794c0a49cfef91e1fcb66e8179484413cd8a50ee794c0a49cfef91e1fcb6 or newer

bubblegum/program/tests/simple.rs
```diff
@@ -1,7 +1,18 @@
 pub mod utils;
```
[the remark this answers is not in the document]
danenbm: It can wait.  Please make sure it is captured in a linear task if you plan to do it later.  Thanks!

danenbm (approved):
Couple minimal comments.  LGTM!

danenbm (dismissed):
Couple minimal comments. LGTM!

danenbm (approved):

bubblegum/program/download-compression-programs.sh
austbot: we can actuially take the newest master its a good signal if they break something

austbot (approved):

PR #779, opened 2022-09-29
fix editions accounting bug

samuelvanderwaal:


samuelvanderwaal (commented):
version patch

token-metadata/program/src/utils.rs
```diff
@@ -550,26 +550,35 @@ pub fn calculate_supply_change<'a>(
     // instead of properly tracking the supply.
     // Now, we increment this by one if the edition number is less than the max supply.
     // This allows users to mint out missing edition numbers that are less than the supply, but
-    // tracks the supply correctly once all missing editions are minted.
+    // tracks the supply correctly for new Master Editions.
     let new_supply = if let Some(max_supply) = max_supply {
+        // We should never be able to mint an edition number that is greater than the max supply.
+        if edition > max_supply {
```
danenbm, line 556: So I guess this was needed because the instructions to mint print editions look like they would have just allocated a new editionMarker account if the edition was out of bounds from existing ones?  So no other checks would have caught that.

token-metadata/program/src/utils.rs
```diff
@@ -550,26 +550,35 @@ pub fn calculate_supply_change<'a>(
...
+    // With no max supply we can increment each time.
+    else {
         current_supply
             .checked_add(1)
             .ok_or(MetadataError::NumericalOverflowError)?
     };
 
-    if let Some(max) = get_max_supply_off_master_edition(master_edition_account_info)? {
-        if new_supply > max {
-            return Err(MetadataError::MaxEditionsMintedAlready.into());
```
danenbm, line 570: I assume you are leaving `MaxEditionsMintedAlready` in `MetadataError` enum since its in the middle and then you don't want the enum to change even though its not being used any more?

token-metadata/program/tests/mint_new_edition_from_master_edition_via_token.rs
```diff
@@ -307,10 +305,39 @@ mod mint_new_edition_from_master_edition_via_token {
         assert_custom_error!(result, MetadataError::EditionOverrideCannotBeZero);
     }
 
+    #[tokio::test]
+    async fn fail_to_mint_edition_num_zero() {
```
danenbm, line 309: Looking at commit history and this PR, you did not change any logic for detecting `MetadataError::EditionOverrideCannotBeZero` right?  Just adding a test case here?

token-metadata/program/tests/mint_new_edition_from_master_edition_via_token.rs
```diff
@@ -423,35 +420,184 @@ mod mint_new_edition_from_master_edition_via_token {
...
+
+        let master_edition_struct = master_edition.get_data(&mut context).await;
+        assert!(master_edition_struct.supply == 2);
+        assert!(master_edition_struct.max_supply == Some(999));
+
+        // Try to remint edition numbers 1 and 99, this should fail.
+        let print_edition = EditionMarker::new(&original_nft, &master_edition, 1);
+        let err = print_edition.create(&mut context).await.unwrap_err();
+
+        assert_custom_error!(err, MetadataError::AlreadyInitialized);
```
danenbm, line 529: It is kind of an obscure sounding error message for this case, reused for a couple different things.  A nitpick would be to ask you to add an error code called `EditionTaken` but if you don't want to at the moment that's fine with me.

token-metadata/program/src/utils.rs
samuelvanderwaal: Correct.

token-metadata/program/tests/mint_new_edition_from_master_edition_via_token.rs
samuelvanderwaal: Correct.

danenbm (approved):
Also discussed test cases in detail and code changes offline, LGTM!

PR #792, opened 2022-10-05
Remove JavaScript Workspaces

samuelvanderwaal:


.eslintrc.js
```diff
@@ -1,16 +1,16 @@
...
-  },
-  extends: ['plugin:@typescript-eslint/recommended', 'prettier', 'plugin:prettier/recommended'],
-  rules: {
-    '@typescript-eslint/explicit-module-boundary-types': 'off',
-    '@typescript-eslint/no-empty-function': 'off',
-    '@typescript-eslint/ban-types': ['error', { extendDefaults: true, types: { '{}': false } }],
-    '@typescript-eslint/ban-ts-comment': 'off',
-  },
-};
+// module.exports = {
```
danenbm, line 1: Can this be removed instead of commented out in main?

package.json
```diff
@@ -1,46 +1,19 @@
 {
```
danenbm, line 1: Can you explain why we still have a top level package.json and a top level yarn.lock?

auction-house/js/.eslintrc.js
```diff
@@ -1 +1,17 @@
-module.exports = require('../../.eslintrc');
+// module.exports = require('../../.eslintrc');
```
danenbm, line 1: Nit: cleanup

.eslintrc.js
samuelvanderwaal: Yeah, will do.

auction-house/js/.eslintrc.js
samuelvanderwaal: Will do.

package.json
samuelvanderwaal: We shouldn't need them anymore. That's leftover from when Thorsten and I were trying to keep some common dependencies without the workspace. I'll remove and we'll see if CI still passes.

samuelvanderwaal: Actually, we still need the root `package.json` because the base amman config relies on it. We can create an amman config for each package and remove the root amman config and `package.json` files later in a separate PR if we decide to go that route.

febo (approved):

danenbm (commented):
approved

PR #794, opened 2022-10-06
Upgrade spl-account-compression and spl-noop 0.1.2

tracy-codes:
Upgrades `spl-account-compression` and `spl-noop` to the latest versions (`0.1.2` at the time of this PR).

`spl-account-compression v0.1.2` contains a breaking change where a previously public module is now private. That update is reflected in `src/lib.rs` changes here.

### DIFF
--- a/bubblegum/program/Cargo.lock
+++ b/bubblegum/program/Cargo.lock
@@ -67,9 +67,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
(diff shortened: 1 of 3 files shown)


danenbm:
Hi Tracy I believe we actually got this change covered as someone from Solana made this change yesterday and we merged it recently as part of this: https://github.com/metaplex-foundation/metaplex-program-library/pull/793/files#diff-887afc68d4b425e306186003fcac2e9d0d278472052c7c5221737ef13e477335

tracy-codes:
Excellent! Looks like work was being done in tandem haha. I'll go ahead and close this one then 🫡

PR #801, opened 2022-10-10
Update generated events in Bubblegum and fix collection verification

danenbm:
### Notes
- Remove deprecated calls to `emit!()` for instruction-based parsing
- Add `AccountType` enum to differentiate different Bubblegum events sent to `spl-noop`
- Instruction IDL changes: add collection authority record PDA as a required account for collection verification functions, and add the log_wrapper account to `decompress_v1`.
- Regenerate API.
- Clippy fixes.

### TODO
Fixing tests


bubblegum/program/src/state/mod.rs
```diff
@@ -50,17 +50,42 @@ impl Voucher {
...
+    pub account_type: AccountType,
     pub version: Version,
     pub id: Pubkey,
     pub tree_id: Pubkey,
     pub nonce: u64,
 }
+
+#[derive(BorshDeserialize, BorshSerialize, PartialEq, Eq, Debug, Clone)]
+#[repr(u8)]
+pub enum AccountType {
```
ngundotra, line 80: Should probably be called `BubblegumEventType`

ngundotra:
Please use the const here instead of `size_of` https://github.com/solana-labs/solana-program-library/blob/1e2579f8e84e62bbc70904d2d44d95679f567514/account-compression/programs/account-compression/src/state/concurrent_merkle_tree_header.rs#L6

bubblegum/program/src/lib.rs
```diff
@@ -500,15 +505,14 @@ fn process_mint_v1<'info>(
         creator_hash.to_bytes(),
     );
     let new_nft = NewNFTEvent {
+        account_type: AccountType::NewNFTEvent,
```
austbot, line 508: We dont need this event

bubblegum/program/src/lib.rs
```diff
@@ -1142,9 +1157,17 @@ pub mod bubblegum {
             data_hash,
             creator_hash,
         );
-        emit!(previous_leaf.to_event());
+
         let new_leaf = Node::default();
-        wrap_application_data_v1(new_leaf.try_to_vec()?, &ctx.accounts.log_wrapper)?;
+
+        let new_leaf_event = NodeEvent {
```
austbot, line 1163: We dont need this event

bubblegum/program/src/state/leaf_schema.rs
```diff
@@ -1,15 +1,16 @@
+use crate::state::AccountType;
 use anchor_lang::{prelude::*, solana_program::keccak};
 use spl_account_compression::Node;
 
-#[event]
+#[derive(AnchorSerialize, AnchorDeserialize, PartialEq, Eq, Debug, Clone)]
 pub struct LeafSchemaEvent {
```
austbot, line 6: We may need a discriminator method.

bubblegum/program/src/state/leaf_schema.rs
```diff
@@ -98,8 +99,9 @@ impl LeafSchema {
 
     pub fn to_event(&self) -> LeafSchemaEvent {
         LeafSchemaEvent {
+            account_type: AccountType::LeafSchemaEvent,
             version: self.version(),
-            schema: *self,
+            schema: self.clone(),
```
austbot, line 104: How was moving working before if so why change?

bubblegum/program/src/state/leaf_schema.rs
austbot: Nevermind you have the one bytes

bubblegum/program/src/state/metaplex_adapter.rs
```diff
@@ -1,12 +1,12 @@
 use anchor_lang::prelude::*;
 
-#[derive(AnchorSerialize, AnchorDeserialize, PartialEq, Eq, Copy, Clone)]
+#[derive(AnchorSerialize, AnchorDeserialize, PartialEq, Eq, Debug, Clone)]
```
austbot, line 3: Doesnt anchor macros add a discrimnator? If so you dont need it as you do it below

austbot (changes requested):

bubblegum/program/src/state/leaf_schema.rs
alexandruag: It wasn't moved before because the struct used to implement `Copy`.

bubblegum/program/src/state/mod.rs
```diff
@@ -50,17 +49,41 @@ impl Voucher {
...
 pub struct NFTDecompressionEvent {
+    pub account_type: AccountType,
     pub version: Version,
     pub id: Pubkey,
     pub tree_id: Pubkey,
     pub nonce: u64,
 }
+
+#[derive(AnchorSerialize, AnchorDeserialize, PartialEq, Eq, Debug, Clone)]
+pub enum AccountType {
```
alexandruag, line 80: At first, this `enum` seemed a bit redundant because the structures we're adding the new field to already encode the kind of the even in their identity (i.e. `struct NodeEvent`, `struct NFTDecompressionEvent`, etc). Is the addition meant to effectively generate a discriminant when they are serialized?

If so, WDYT about (like Noah meantioned) renaming this to something like `BubblegumEvent`,and then having it wrap the even data, i.e.:

```rust
pub enum BubblegumEvent {
    ...
    Node(NodeEvent),
    Decompress(NftDecompressionEvent)   
    ...
}
```
As things are right now, there's nothing stopping on particular event to be initialized with a different event type, and all the initializations are explicit. The may be a bit clearer and less error prone.

bubblegum/program/src/lib.rs
```diff
@@ -638,12 +644,16 @@ fn process_collection_verification<'info>(
...
     let edition_account = ctx.accounts.edition_account.to_account_info();
     let collection_authority = ctx.accounts.collection_authority.to_account_info();
+    let collection_authority_record_pda = ctx
+        .accounts
+        .collection_authority_record_pda
+        .to_account_info();
     let bubblegum_signer = ctx.accounts.bubblegum_signer.to_account_info();
     let token_metadata_program = ctx.accounts.token_metadata_program.to_account_info();
 
     // Look for collection authority record PDA as a remaining account.
```
alexandruag, line 654: This comment looks obsolete now.

bubblegum/program/src/lib.rs
```diff
@@ -638,12 +644,16 @@ fn process_collection_verification<'info>(
...
-    let collection_authority_record = if !ctx.remaining_accounts.is_empty() {
-        Some(&ctx.remaining_accounts[0])
+    let collection_authority_record = if collection_authority_record_pda.key() == crate::id() {
```
alexandruag, line 655: The json docs for `collectionAuthorityRecordPda` mention "If there is no collecton authority record PDA then this must be the Bubblegum program address". Is setting `collection_authority_record` here to `None` when `collection_authority_record_pda.key()` is different from `crate::id()` in line with that? (seems a bit like the other way around at first, name it being `None` where that account is the program id)

bubblegum/program/src/state/leaf_schema.rs
danenbm: yes it had copy, I thought it was better to be more explicit.

bubblegum/program/src/state/metaplex_adapter.rs
danenbm: Probably I should switch to Borsh

bubblegum/program/src/lib.rs
danenbm: In the program parser PR you also mentioned: "We dont need any event except leaf schema".

I'm wondering should I just get rid of all the other event types then, and get rid of the AccountType 1-byte discriminator?

Then on the program parser we can just always assume its a `LeafSchemaEvent`.  This would be much simpler.

The choice is between:
A) Simpler
B) Potential future proofing if we need more info than what LeafSchema can provide.

bubblegum/program/src/lib.rs
danenbm: Yes you are correct I reversed the logic.

bubblegum/program/src/lib.rs
danenbm: Fixed

danenbm:
> Please use the const here instead of `size_of` https://github.com/solana-labs/solana-program-library/blob/1e2579f8e84e62bbc70904d2d44d95679f567514/account-compression/programs/account-compression/src/state/concurrent_merkle_tree_header.rs#L6

Done

bubblegum/program/src/lib.rs
danenbm: Had brief offline conversation with @austbot while he's travelling.  From my impression, I will leave the 1-byte discriminator account type, but remove the currently unneeded events.  This will future proof us in a couple different ways but also remove currently extraneous code.

bubblegum/program/src/lib.rs
danenbm: Removed in next version.

bubblegum/program/src/state/leaf_schema.rs
danenbm: Keeping the one byte per offline conversation, see other comment for details.

bubblegum/program/src/state/mod.rs
danenbm: Yeah a constructor would be less error prone.  I don't know if we need a wrapper struct, the `LeafSchemaEvent` struct is already wrapping the `LeafSchema`, seems like adding a lot of wrapping, and I think the constructor solves the issue of manually setting the field.  LMK what you think of my next draft, and yes to confirm its a bit strange but meant to add a one-byte discriminant to the struct.

bubblegum/program/src/state/mod.rs
danenbm: Done in next version.

bubblegum/program/src/lib.rs
danenbm: Updated in latest code to have just `LeafSchemaEvent`

bubblegum/program/src/state/mod.rs
danenbm: Added constructor in next version.

bubblegum/program/src/state/metaplex_adapter.rs
danenbm: I switched all the LeafSchema stuff to use Borsh, but haven't changed metaplex_adapter at all.  I got some JS API generation failures when I tried to switch metaplex_adapter to Borsh.

austbot (approved):

alexandruag (approved):

PR #810, opened 2022-10-12
chore: bump token-metadata deps

samuelvanderwaal:
This may not be strictly necessary since I removed the tilde pinning allowing more flexibility with versions.


austbot (approved):
version minor

danenbm (approved):

PR #814, opened 2022-10-12
fix: Close the Listing Config for Cancel

blockiosaurus:
The Listing Config was not being closed when canceling a listing, resulting in issues when attempting to relist the NFT. This PR fixes that so the Listing Config will also be closed in the cancel program flow.


samuelvanderwaal (approved):

stranzhay (approved):

blockiosaurus:
version patch

danenbm (approved):

PR #818, opened 2022-10-13
Transfer effects

nilz3ro:


trifle/cli/src/trifle-cli.ts
```diff
@@ -38,6 +39,45 @@ import { PublicKeyMismatchError } from "@metaplex-foundation/mpl-auction-house";
 //   chalk.green(figlet.textSync("Trifle CLI", { horizontalLayout: "full" })),
 // );
 
+const addTransferEffectsOptions = (cmd: Command) => {
```
blockiosaurus, line 42: This is really clever and nicely done!

blockiosaurus (approved):
Looks awesome! A couple random notes.
- Once we update to the newer Solita with Optional accounts in fixed positions we'll probably need to change the Instruction builders.
- I removed the try_len in favor of the serialize & len() calls when I was in a frenzy this week and didn't feel like doing math. I'm hoping we can jump into a Token Metadata refactor soon and wrap this in a trait anyway.
- I removed constraint_model from the PDA seeds so hopefully that doesn't cause too much merge pain.

Otherwise good to go!

trifle/cli/src/trifle-cli.ts
nilz3ro: Thank you @blockiosaurus! Compliments from a talented engineer are always appreciated.

blockiosaurus (approved):

PR #821, opened 2022-10-13
refactor(token-metadata): move constants to separate file

samuelvanderwaal:
This is cleanup work preparing for some other changes I need to make to Token Metadata. It moves constants out of the `state.rs` file into `constants.rs` to limit the bloat of the state file. 

Closing this because needs and requirements have changed. Token Metadata still needs some refactoring but it will be done in a different PR.

### DIFF
--- a/token-metadata/program/src/assertions/collection.rs
+++ b/token-metadata/program/src/assertions/collection.rs
@@ -1,11 +1,12 @@
 use solana_program::{account_info::AccountInfo, program_error::ProgramError, pubkey::Pubkey};
 
 use crate::{
+    constants::{EDITION, PREFIX},
     error::MetadataError,
     pda::find_collection_authority_account,
     state::{
         Collection, CollectionAuthorityRecord, MasterEditionV2, Metadata, TokenMetadataAccount,
-        TokenStandard, EDITION, PREFIX,
+        TokenStandard,
     },
     utils::assert_derivation,
 };
--- a/token-metadata/program/src/constants.rs
+++ b/token-metadata/program/src/constants.rs
@@ -1,0 +1,69 @@
+/// prefix used for PDAs to avoid certain collision attacks (<https://en.wikipedia.org/wiki/Collision_attack#Chosen-prefix_collision_attack>)
+pub const PREFIX: &str = "metadata";
+
+/// Used in seeds to make Edition model pda address
+pub const EDITION: &str = "edition";
+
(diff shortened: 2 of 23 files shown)


token-metadata/program/src/assertions/collection.rs
```diff
@@ -1,11 +1,12 @@
 use solana_program::{account_info::AccountInfo, program_error::ProgramError, pubkey::Pubkey};
 
 use crate::{
+    constants::{EDITION, PREFIX},
```
alexandruag, line 4: As a mention, I've generally seen `defs` used as the module that holds various constants and other definitions in a bunch of projects.

token-metadata/program/src/constants.rs
```diff
@@ -0,0 +1,69 @@
...
+
+pub const COLLECTION_AUTHORITY: &str = "collection_authority";
+
+pub const MAX_NAME_LENGTH: usize = 32;
+
+pub const MAX_SYMBOL_LENGTH: usize = 10;
+
+pub const MAX_URI_LENGTH: usize = 200;
+
+pub const MAX_METADATA_LEN: usize = 1 //key
```
alexandruag, line 21: The detailed breakdown of what each literal value is very useful; would be great if we can do the same for the other similarly defined constants.

token-metadata/program/src/constants.rs
```diff
@@ -0,0 +1,69 @@
...
+
+// can hold up to 200 keys per reservation, note: the extra 8 is for number of elements in the vec
+pub const MAX_RESERVATION_LIST_V1_SIZE: usize = 1 + 32 + 8 + 8 + MAX_RESERVATIONS * 34 + 100;
+
+// can hold up to 200 keys per reservation, note: the extra 8 is for number of elements in the vec
+pub const MAX_RESERVATION_LIST_SIZE: usize = 1 + 32 + 8 + 8 + MAX_RESERVATIONS * 48 + 8 + 8 + 84;
+
+pub const MAX_EDITION_MARKER_SIZE: usize = 32;
+
+pub const EDITION_MARKER_BIT_SIZE: u64 = 248;
```
alexandruag, line 65: Just wondering, what does this value stand for?

samuelvanderwaal: Edition Marker is an account that tracks which edition numbers have been printed. It has a 31 byte ledger that represents up to 248 prints. A new edition market is created for prints 249 - 496, and so on.

PR #832, opened 2022-10-18
Add hidden settings validation

febo:
This PR adds an additional check when using hidden settings: if config line settings are present at the same time as hidden settings, the `initialize` and `update` instructions will report an error. Currently, hidden settings take precedence over config line settings, so the behviour of the candy machine does not change. The extra validation will make the on-chain configuration of the candy machine "cleaner".


samuelvanderwaal:
There's a clippy problem, otherwise lgtm.

samuelvanderwaal (approved):

samuelvanderwaal (approved):

stegaBOB (approved):

PR #846, opened 2022-10-24
Public trees and Future Bot Taxability

austbot:
Bubblegum trees were previously permissioned, meaning only the authority or some delegate could mint to the tree, but what if metaplex wanted to make some BIG OLE Junk in the TRUNK(its tree get it) trees and have them open for people to mint to without them needing to make a tree. This pr allows this, adds tests to ensure private trees are still private and also adds the needed accounts to be able to bot tax spammers.

NOTE this does add an addition 32 bytes on an already squeeezing by set of proof transactions. NOTE we send proof elements in as accounts.


ngundotra:
Why is `systemProgram` needed on each account + why the additional 15 bytes of padding on the `TreeAuthority`?

bubblegum/program/tests/simple.rs
```diff
@@ -291,3 +295,43 @@ async fn test_decompress_passes() {
...
+    let tree = context
+        .create_public_tree::<MAX_DEPTH, MAX_BUF_SIZE>()
+        .await.unwrap();
+    let tree_private = context
+        .default_create_tree::<MAX_DEPTH, MAX_BUF_SIZE>()
+        .await.unwrap();
+    let payer = context.payer();
+    let minter = Keypair::new(); // NON tree authority payer, nor delegate
+
+    context.fund_account(minter.pubkey(), 10000000000).await.unwrap();
```
danenbm, line 313: Nit: can you use `DEFAULT_LAMPORTS_FUND_AMOUNT`?

bubblegum/program/README.md
```diff
@@ -436,3 +436,9 @@ Decompress an NFT into an uncompressed Metaplex NFT.  This will cost rent for th
 | `data`                            | 0      | ~    | [`MetadataArgs`](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/bubblegum/program/src/state/metaplex_adapter.rs#L81) object.
 
 </details>
+
+## Testing
+```bash
+cargo build
+cargo test-bpf --bpf-out-dir ../../test-programs
+```
```
danenbm, line 444: Wdyt about moving this more towards the top, i.e. after the Basic operation section but before the detailed Accounts section? Otherwise seems like it is too far down to immediately see and this is good info to have at top level.

bubblegum/program/tests/simple.rs
```diff
@@ -291,3 +295,43 @@ async fn test_decompress_passes() {
...
+    assert_eq!(cfg.tree_creator, payer.pubkey());
+    assert_eq!(cfg.tree_delegate, payer.pubkey());
+    assert_eq!(cfg.total_mint_capacity, 1 << MAX_DEPTH);
+    assert_eq!(cfg.public, true);
+
+    tree.mint_v1_non_owner(&minter, &mut args).await.unwrap();
+    let cfg = tree.read_tree_config().await.unwrap();
+    assert_eq!(cfg.num_minted, 1);
+
+    if let BanksClient(BanksClientError::TransactionError(e)) = tree_private.mint_v1_non_owner(&minter, &mut args).await.unwrap_err() {
```
danenbm, line 329: Nit: this test tests a couple different things.  Would it be incrementally better to move this part be into a separate test case called something like `test_private_tree_mint_v1_non_owner_fails`?

bubblegum/program/tests/simple.rs
austbot: needed more

bubblegum/program/README.md
austbot: fine with me

danenbm (approved):
LGTM!

blockiosaurus (approved):

PR #850, opened 2022-10-26
CI Workflow Refactor

stegaBOB:
Refactors all the CI to make it a lot faster and more efficient.


.github/actions/integration/candy-machine-core/action.yml
```diff
@@ -0,0 +1,27 @@
...
+      with:
+        name: token-metadata
+
+    # Install JS SDK deps
+    - uses: ./.github/actions/yarn-install-and-build
+      with:
+        cache_id: sdk-candy-machine-core
+        working_dir: ./candy-machine-core/js
+        build_token_metadata: true
+        build_auction_house: true
```
samuelvanderwaal, line 26: Why does this require building Auction House?

.github/actions/integration/candy-machine-core/action.yml
```diff
@@ -0,0 +1,27 @@
...
+        build_candy_machine: true
```
samuelvanderwaal, line 27: Does it need this? Is this candy machine v2?

.github/actions/yarn-install-and-build/action.yml
```diff
@@ -11,19 +11,21 @@ inputs:
   build_core:
     description: If not false it will build the core package
     required: false
-    default: true
+    default: 'true'
```
samuelvanderwaal, line 14: Why do these need quotes?

.github/actions/integration/candy-machine-core/action.yml
stegaBOB: I think I just duplicated how we had the behavior already? I think it might be a consequence of how we had workspaces on in the past and it may not be needed anymore?

.github/actions/integration/candy-machine-core/action.yml
stegaBOB: I think this was an artifact from yarn workspaces.

.github/actions/yarn-install-and-build/action.yml
stegaBOB: YAML stuff. I think github will cast it to a string anyways, but the inputs have to be strings I think

samuelvanderwaal: Why only these inputs and not all of them?

stegaBOB: it is all of them. for actions at least I think? tbh I'm just doing what my IDE linter tells me to do.

samuelvanderwaal (dismissed):

samuelvanderwaal (approved):

blockiosaurus (approved):

PR #853, opened 2022-10-26
fix: Allow the seller to cancel its bids from its own listing

akkoucai:
I added the condition buyer_price == AUCTIONEER_BUYER_PRICE in the last "if" statement of cancel instruction.

Indeed, when a seller wants to cancel its bids from a listing they created, the listing config account is getting closed because of the last commit in auctioneer.

Adding this condition make sure the seller wants to cancel its listing not its bids.


akkoucai:
@austbot Thought you could allow that by default as I don't see why a seller shouldn't bid on its own item.
I experienced some issues with this case when I was doing some tests with the same wallet to create the auction and to bit on it.
I agree that it's uncommon though.

PR #854, opened 2022-10-26
Remove constraints from constraint model

nilz3ro:


trifle/program/src/processor.rs
```diff
@@ -832,3 +837,49 @@ fn add_tokens_constraint_to_escrow_constraint_model(
...
+    if escrow_constraint_model.update_authority != *update_authority_info.key {
+        return Err(TrifleError::InvalidUpdateAuthority.into());
+    }
+
+    // remove the constraint by key.
+    escrow_constraint_model
+        .constraints
+        .remove(&args.constraint_name);
+
+    let serialized_data = escrow_constraint_model.try_to_vec().unwrap();
```
blockiosaurus, line 868: I have learned Sam doesn't approve anything with unwraps so this should be changed to
```
let serialized_data = escrow_constraint_model
        .try_to_vec()
        .map_err(|_| MetadataError::BorshSerializationError)?;
```

blockiosaurus: Or similar error.

blockiosaurus (changes requested):
Awesome, this looks great! Just a few minor comments

trifle/program/src/instruction.rs
```diff
@@ -150,6 +157,13 @@ pub enum TrifleInstruction {
     #[account(3, name = "system_program", desc = "System program")]
     AddTokensConstraintToEscrowConstraintModel(AddTokensConstraintToEscrowConstraintModelArgs),
 
+    #[default_optional_accounts]
+    #[account(0, writable, name = "constraint_model", desc = "Constraint model account")]
+    #[account(1, writable, signer, name = "payer", desc = "Wallet paying for the transaction")]
+    #[account(2, optional, signer, name = "update_authority", desc = "Update authority of the constraint model")]
```
samuelvanderwaal, line 163: Why is this account optional in the JS given the on-chain handler expects it and explicitly checks it?

trifle/program/src/state/transfer_effects.rs
```diff
@@ -22,6 +22,7 @@ impl From<u16> for TransferEffects {
     }
 }
 
+#[allow(clippy::from_over_into)]
```
samuelvanderwaal, line 25: Is this a deliberate choice to implement `Into`, instead of `From`, which gives you the `Into` implementation automatically?

trifle/program/src/instruction.rs
nilz3ro: This is in response to Keith's feedback but I think it actually should be explicitly required.

samuelvanderwaal (approved):

PR #857, opened 2022-10-26
fix: set freeze auth in DecompressV1

samuelvanderwaal:
Token Metadata v1.6 disallows setting a freeze authority to `None`. This fixes the DecompressV1 instruction to use the mint authority as the freeze authority. (This is overridden with the edition when create_master_edition is called.)

There are also some `cargo fmt` lints.


danenbm (approved):

febo (approved):

samuelvanderwaal (commented):
version patch

PR #862, opened 2022-10-27
Removing enum variant

blockiosaurus:


token-metadata/program/src/instruction.rs
```diff
@@ -544,38 +544,42 @@ pub enum MetadataInstruction {
...
+    #[account(2, writable, signer, name="payer", desc="Wallet paying for the transaction and new account")]
+    #[account(3, name="attribute_mint", desc="Mint account for the new attribute")]
+    #[account(4, writable, name="attribute_src", desc="Token account source for the new attribute")]
+    #[account(5, writable, name="attribute_dst", desc="Token account, owned by TM, destination for the new attribute")]
+    #[account(6, name="escrow_mint", desc="Mint account that the escrow is attached")]
+    #[account(7, name="escrow_account", desc="Token account that holds the token the escrow is attached to")]
+    #[account(8, name="system_program", desc="System program")]
+    #[account(9, name="ata_program", desc="Associated Token program")]
+    #[account(10, name="token_program", desc="Token program")]
+    #[account(11, name="rent", desc="Rent info")]
```
samuelvanderwaal, line 580: We do not need the rent account, we can delete it and use the `instruction::create_associated_account` which does not require rent passed in. The version of that instruction currently used is deprecated.

token-metadata/program/src/escrow/transfer_out.rs
```diff
@@ -85,6 +90,7 @@ pub fn process_transfer_out_of_escrow(
     let ata_program_info = next_account_info(account_info_iter)?;
     let token_program_info = next_account_info(account_info_iter)?;
     let rent_info = next_account_info(account_info_iter)?;
+    let _sysvar_ix_account_info = next_account_info(account_info_iter)?;
```
samuelvanderwaal, line 93: Remove the rent account.

samuelvanderwaal (commented):
Looks good except for one minor improvement we should make which I missed in the initial TOE review.

token-metadata/program/src/instruction.rs
blockiosaurus: Good call. I believe this was made before the TM dependency update.

samuelvanderwaal (dismissed):

samuelvanderwaal (commented):
version patch

samuelvanderwaal (approved):

austbot (approved):

Issue #822, opened 2022-10-13
[Bug]: Execute sale - sum of account balances before and after instruction do not match


viyer28:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

Running into a weird error with **executeSale** - I'm trying to execute a sale as the seller of a token on an existing buy offer.
Basically, I'm packing **createSellOffer**, **printListingReceipt**, **executeSale**, and **printPurchaseReceipt** in one transaction.
In the **executeSale** instruction, I get this error:
 
`sum of account balances before and after instruction do not match`

It looks like something to do with ImmutableOwner and Token Program 2022. I've attached a sample failed devnet transaction [here](https://explorer.solana.com/tx/2Tnh1L9DSDCBAFsKiTT1Q3VYb9SdzjFiLProfGqafNrLXABxjK24bhD1NgRf99YnHMdcSitean1VSmyoTCYSu4k8?cluster=devnet).

Any idea what could be going on?

### Relevant log output

```shell
> Program log: Instruction: ExecuteSale
> Invoking System Program
  > Program returned success
> Invoking System Program
  > Program returned success
> Invoking System Program
  > Program returned success
> Invoking System Program
  > Program returned success
> Invoking Associated Token Account Program
  > Program log: Create
  > Invoking Token Program
    > Program log: Instruction: GetAccountDataSize
    > Program Token Program  consumed 1622 of 637426 compute units
    > Program return: TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA pQAAAAAAAAA=
    > Program returned success
  > Invoking System Program
    > Program returned success
  > Program log: Initialize the associated token account
  > Invoking Token Program
    > Program log: Instruction: InitializeImmutableOwner
    > Program log: Please upgrade to SPL Token 2022 for immutable owner support
    > Program Token Program  consumed 1405 of 630936 compute units
    > Program returned success
  > Invoking Token Program
    > Program log: Instruction: InitializeAccount3
    > Program Token Program  consumed 4241 of 627052 compute units
    > Program returned success
  > Program Associated Token Account Program  consumed 23545 of 646018 compute units
  > Program returned success
> Invoking Token Program
  > Program log: Instruction: Transfer
  > Program Token Program  consumed 4735 of 613710 compute units
  > Program returned success
> Program Auction House  consumed 104003 of 709593 compute units
[...]

PR #868, opened 2022-11-02
add seller flow test

samuelvanderwaal:
Attempting to duplicate [this bug](https://github.com/metaplex-foundation/metaplex-program-library/issues/822).


samuelvanderwaal:
> @samuelvanderwaal want this in ?

I couldn't get this test to fail for the specific scenario it was supposed to test so I think we can just close this PR.

PR #875, opened 2022-11-03
[Hydra] move extra SOL in fanout account to native account to distribute

cryptorrivem:
Purpose: Enable the ability to use Hydra fanout account address for creator royalties across SOL and SPL markets.

Background:
When we create a fanout wallet, it'll allow to distribute SOL and SPL funds to the members with the shares setup. But, the tests I've made, and then looking at the code, doesn't offer a uniform address to use so whether the wallet is receiving SOL or SPL tokens, we have only one address to send to. SOL is distributed from the Native account, and SPL tokens from the PDA for each mint out of the Fanout account.

My goal was to use the fanout account as the uniform for accepting both SOL and SPL, and then, when trying to distribute SOL, transferring the extra (besides rent) from the fanout account to the native account, to continue with the regular shares distribution.

This will also require that to compute the SOL balance for the wallet, we should consider both Native and Fanout accounts as available lamports.

Disclaimer:
I haven't been able to deploy this on localnet or devnet (I'm still a novice dealing with Solana programs and Rust), so I'm bringing this code to help someone else look at this and let me know if the approach is incorrect.

### DIFF
--- a/hydra/program/src/state.rs
+++ b/hydra/program/src/state.rs
@@ -16,6 +16,7 @@ impl Default for MembershipModel {
     }
 }
 
+pub const FANOUT_ACCOUNT_SIZE: usize = 300;
 #[account]
 #[derive(Default, Debug)]
 pub struct Fanout {
--- a/hydra/program/src/utils/logic/distribution.rs
+++ b/hydra/program/src/utils/logic/distribution.rs
@@ -1,4 +1,4 @@
-use crate::state::{Fanout, FanoutMembershipVoucher, HOLDING_ACCOUNT_SIZE};
+use crate::state::{Fanout, FanoutMembershipVoucher, FANOUT_ACCOUNT_SIZE, HOLDING_ACCOUNT_SIZE};
 use crate::utils::logic::calculation::*;
 use crate::utils::logic::transfer::{transfer_from_mint_holding, transfer_native};
 use crate::utils::parse_fanout_mint;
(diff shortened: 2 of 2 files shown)


cryptorrivem:
yes @austbot, that's why the rent (300 bytes) is deducted before doing the transfer, just like the `holding_account` does lines after my code piece

cryptorrivem:
well, in this case, it's not accidental. I want to have a SINGLE Solana address to use on an NFT creators array. If I use the native account, then I cannot send SPL tokens, cause it cannot have a token account derived (I need this for the custom market I'm building with LIQNFT), and if I use the fanout account, it'll work for SPL when distributing, but doesn't for SOL, as SOL is expected to be in the native account.

So my "hack" is using the fanout account as the unique address, and then, on distribute, move the SOL that's in there to the native account, so then the regular SOL distribution happens.

cryptorrivem:
ok @austbot, I finally managed to deploy my version of Hydra and successfully did the distribution with the fanout account as the unique deposit address. Here's a list of transactions I did to prove this code is working:

Fanout account (B8DeGb6ySo2z8iF5Fbg3J5JFbpiWtVwYFQ6iCKswsJpR):
https://explorer.solana.com/address/B8DeGb6ySo2z8iF5Fbg3J5JFbpiWtVwYFQ6iCKswsJpR?cluster=devnet

Deposit SOL in the Fanout account:
https://solana.fm/tx/4aGHdbJh9bTJ2N31KZE59xv1DcV9KmP6ibA9assRaMQd5oufQVJaxw4oJ8vw1KspqzUw7mDv8ijgfYim6qYRitzj?cluster=devnet-solana

Distribute SOL to the members (moved from the Fanout account to the Native account, then to the members)
https://solana.fm/tx/2Xkt3tb6bzcd4YGFefJfHPZQDaQQhTWRHkCwUrSoysP9RKZppjpN1KAqd6NKUNMpCtULTG5Q6ikcKRmgS3mXvW2L?cluster=devnet-solana

Deposit SPL to the Fanout account derived token address:
https://solana.fm/tx/65PkqQJFLumXu8vjd4oZ8CsmQnsEmn3Rn5QLBeQHf6J8uSPdLDBwiA4wZDgaqxjFvpLHcqpxhsWCVAd4yvXFPqgY?cluster=devnet-solana

Distribute SPL to the members (as usual)
https://solana.fm/tx/3hidGSbELs2ctNHMDanQwJGZeb9WSauC2YEDvXDEkSeAyb9fs4rGnL3XA2AUBnt1zyfzJKVnCc5KW2KzDbCSxFxD?cluster=devnet-solana

cryptorrivem:
@austbot then how will I be able to withdraw the SPL from those?

My idea was to use the default distribute instructions, so the integration was seamless

cryptorrivem:
exactly, the distribution for SPL will continue to work as usual, but the change will allow to use the same address to accept SOL and SPL deposits for later distribution

Issue #897, opened 2022-11-25
[Bug]: Auctioneer Sell Differences in Computed ATA from Actual ATA


kespinola:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

There is a difference in the actual ata and computed ata for an NFT which is causing [`assert_is_ata`](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/auction-house/program/src/sell/mod.rs#L391) to fail preventing the auctioneer_sell from succeeding.

```
// computed
// await Token.getAssociatedTokenAddress(ASSOCIATED_TOKEN_PROGRAM_ID, TOKEN_PROGRAM_ID, tokenMint, publicKey);
ata.toBase58() 
'DTpjAMm9UREMqN68rwAYBHdsC9F1gaSH7DWv9cXuoeHk'

// Actual ATA
// new PublicKey(nft.owner.associatedTokenAccountAddress);
associatedTokenAccount.toBase58()
'7AtjFtiCySYsEhc7sHykJwEiKjYe914Gd8SgWKsRaViH'
```

In doing some debugging on different marketplaces the NFT was listed on Hyperspace in which extra instructions were prepended to the sell instruction to close out the current ata of `7Ai...RaViH` and sent the NFT to `DTP...eHK` the one computed by `getAssociatedTokenAddress` before creating the listing.

### Questions

Why is the computed ATA different from the actual?
Should our instructions be updated to perform the same prep when differences in the ATA are computed?

### Hyperspace Listing Transaction 
https://solscan.io/tx/TFkUfchbyrqrxVKvGm5kHyMciqNkrmgkwmmrJTGmLDUeS2my1pacJkv7KFgAvDUNZettotQY3Q2xcHBCiGSqW4N

### Priority this issue should have

Medium (should be fixed soon)

PR #898, opened 2022-11-29
fix: remove the seed assertion on the assert_is_ata because there are…

kespinola:
### Changes
- Remove the ATA seed check from `assert_is_ata` as their are multiple valid variants. See attached issue for example of valid ATA being rejected by the program.

resolves https://github.com/metaplex-foundation/metaplex-program-library/issues/897


austbot (approved):

austbot (approved):
version patch

PR #902, opened 2022-12-06
update spl account compression

austbot:
This updates account compression to the newest which exposes needed values. 
Also this pr adds the indexing foundation for mintinto collection


stegaBOB:
Looks like the tests are failing. How come the IDL changed so much?

PR #907, opened 2022-12-10
Bump Token Metadata to 1.6.3

austbot:
1.6.3 which is alreay on dev and mainnet because it was a security fix needs to be represented in the cargo.


febo (dismissed):
LGTM

samuelvanderwaal (approved):

blockiosaurus (approved):

PR #916, opened 2022-12-15
Fix: re-export assertions and mpl-utils methods from TM to undo breaking change

stegaBOB:
replaces #914


samuelvanderwaal (approved):

stegaBOB (commented):
version patch

PR #918, opened 2022-12-17
Upgrade compression to anchor 26

ChewingGlass:


samuelvanderwaal:
Does #991 address this?

PR #919, opened 2022-12-19
Add accounts context

febo:
This PR creates a new struct `Context` to get access to the accounts passed to a transaction. The creation of each instructions' accounts struct is automated by a macro using the Shank annotation and an addition `#[args]` attribute.

The advantage of this approach is that once Shank annotation are updated, all related accounts struts are automatically updated. The macro uses a builder pattern, which avoids introducing breaking changes to the client even if a new accounts is added to the instruction as long it is not a required account. Optional accounts are supported using the `program_id` pattern, where if an optional account is not set, the instruction builder can set the `program_id` as the pubkey to maintain the position of accounts.

## Example

Consider a `create` instruction with Shank annotation:
```Rust
#[account(0, writable, name="metadata", desc="Metadata account key (pda of ['metadata', program id, mint id])")]
#[account(1, name="mint", desc="Mint of token asset")]
#[account(2, signer, name="mint_authority", desc="Mint authority")]
#[account(3, signer, writable, name="payer", desc="Payer")]
#[account(4, name="update_authority", desc="update authority info")]
#[account(5, name="system_program", desc="System program")]
#[account(6, name="sysvar_instructions", desc="Instructions sysvar account")]
#[account(7, name="spl_token_program", desc="SPL Token program")]
#[account(8, optional, name="master_edition", desc="Unallocated edition account with address as pda of ['metadata', program id, mint, 'edition']")]
#[account(9, optional, name="authorization_rules", desc="Token Authorization Rules account")]
#[args(initialize_mint: bool)]
#[args(update_authority_as_signer: bool)]
Create(CreateArgs)
```

The macro generates a struct to be used by the program, where each account is a `AccountInfo<'a>`; it also generates an struct where each account is a `Pubkey`, together with any `#[args]` attributes. This last struct has a builder struct to facilitate the creation of the instruction:
```Rust
let create_ix = CreateBuilder::new()
    .metadata(metadata)
    .master_edition(master_edition)
    .mint(mint.pubkey())
    .mint_authority(payer_pubkey)
    .payer(payer_pubkey)
    .update_authority(payer_pubkey)
    .initialize_mint(false)
    .update_authority_as_signer(true)
    .build(CreateArgs::V1 {
        asset_data: asset,
        decimals: Some(0),
        max_supply: Some(0),
    })?
    .instruction();
```
The logic to create the instruction is defined using a trait `InstructionBuilder`, which needs to be implemented for each builder struct:
```Rust
impl InstructionBuilder for super::builders::Create {
    fn instruction(&self) -> solana_program::instruction::Instruction {
        let mut accounts = vec![
            AccountMeta::new(self.metadata, false),
            AccountMeta::new_readonly(self.mint_authority, true),
            AccountMeta::new(self.payer, true),
            AccountMeta::new_readonly(self.update_authority, self.update_authority_as_signer),
            AccountMeta::new_readonly(self.system_program, false),
            AccountMeta::new_readonly(self.sysvar_instructions, false),
            AccountMeta::new_readonly(self.spl_token_program, false),
        ];
        ...
        Instruction {
            program_id: crate::ID,
            accounts,
            data: MetadataInstruction::Create(self.args.clone())
                .try_to_vec()
                .unwrap(),
        }
    }
}
```
On the program, accounts can be accessed using a `Context` struct, where access to accounts can be done by name:
```Rust
pub struct Context<'a, T> {
    pub accounts: T,
    pub remaining_accounts: Vec<&'a AccountInfo<'a>>,
}
...

pub fn create<'a>(
    program_id: &Pubkey,
    accounts: &'a [AccountInfo<'a>],
    args: CreateArgs,
) -> ProgramResult {
    // unpacks all accounts into the autogenerated struct
    let cxt = Create::as_context(accounts)?;
    ...
    cxt.accounts.update_authority
}
```

In summary, code that needs manual input:
* Shank annotation on the instruction enum variant
* Implementation of the `InstructionBuilder` trait


samuelvanderwaal (approved):

token-metadata/program/src/instruction/mod.rs
```diff
@@ -515,6 +516,8 @@ pub enum MetadataInstruction {
     #[account(7, name="spl_token_program", desc="SPL Token program")]
     #[account(8, optional, name="master_edition", desc="Unallocated edition account with address as pda of ['metadata', program id, mint, 'edition']")]
     #[account(9, optional, name="authorization_rules", desc="Token Authorization Rules account")]
+    #[args(initialize_mint: bool)]
+    #[args(update_authority_as_signer: bool)]
     Create(CreateArgs),
```
blockiosaurus, line 521: This is going to need a Shank and Solita update for optional signers so the mint can be dynamically marked as signer or not.

blockiosaurus (changes requested):

token-metadata/program/src/instruction/mod.rs
febo: I am not sure if this is possible since Shank/Solita do not know anything about the logic that would make the account a signer or not. One way to address this is to replicate the idea of the instruction builder to the typescript SDK and add a way to register a plug-in to define additional logic.

Our currents test use a higher-level API built on top of the autogenerated one that includes this logic.

febo: @blockiosaurus Loris is working on a new version of solita (a.k.a. kinobi) which will support optional signers.

PR #924, opened 2022-12-24
serialize pubkeys with As DisplayFromStr; clippy fixes

samuelvanderwaal:
Uses `As::DisplayFromStr` on data struct `Pubkey`s to have them serialized as a Base58 strings instead of a `Vec<u8>`. Fix es some clippy lints and unused test items.


samuelvanderwaal:
This is in v1.7 on the feature branch `feat/programmable-assets`.

PR #925, opened 2022-12-27
Fix escrow delegate and closure

blockiosaurus:
Transfer Out on the Escrow feature didn't check if the escrow was empty before closing the account. This resulted in failures when transferring out SFTs in quantities less than the full amount in escrow. This PR adds a check for Escrow ATA amount.

This PR also adds a delegate check for Transfer Out. This resolves a social exploit where a user could list a token with valuable assets in escrow, wait for a bid, then remove the escrowed tokens before executing the sale.


samuelvanderwaal (approved):

austbot (approved):
This change fixed a gap in the prior TOE code that basically assumed that all Token owned tokens were non-fungibles meaning the token account would only have 1 always. This assumption was proven by the bug here, the code assumed that when transferring it needed to close the token account, which should only be the case if the TOE token account has 0. which can take multiple transfer-out operations depending on the setup. I. the case of fungible tokens. 
This code also performs a delegate check on the base token of the TOE, or the actual TOE token to ensure if this TOE is listed in an escrow-less sale, the seller cannot rug the potential buyer via removing the tokens owned by the escrow.

PR #928, opened 2022-12-28
change ProgrammableConfig to triple opt; fix migrate config setting

samuelvanderwaal:


lorisleiva (approved):
## Reviewer summary

This PR provides an alternative way of indicating a change in data.

### Context

The current way of indicating a change is via an `Option<T>` where:
- `Some<T>` tells the program we'd like to change the data to the provided `T`
- `None` tells the program we do not wish to update the data.

For instance, the following code...

```rust
UpdateArgs {
  name: None,
  age: Some(42),
}
```

...indicates that we want to update the `age` attribute to `42` without affecting the `name` attribute.

However, some attributes are themselves `Option` types. This means we end up with a double `Option` type where the first `Option` indicates whether we want to change the attribute or not, and the second `Option` indicates the new value of that attribute.

Here's an example.

```rust
UpdateArgs {
  maxSupply: None, // <- Do not update the max supply. 
  maxSupply: Some(None), // <- Update the max supply to `None`. 
  maxSupply: Some(Some(42)), // <- Update the max supply to `Some(42)`.
}
```

### Problem

The issue with that is Solita, our client generator for JavaScript, interprets `Option<T>` as `T | null` since the `Option` monad does not exist natively in JavaScript.

```ts
type Option<T> = T | null;
```

This means nested `Option` types don't keep enough information as they are unwrapped into "all or nothing". Here's an example of how TypeScript interprets nested options. 

```ts
type NestedOption = Option<Option<T>>
type NestedOption = Option<T | null>
type NestedOption = (T | null) | null
type NestedOption = T | null
```

Therefore, due to the current lack of support for nested options in our clients, we need an alternative way to convey that ternary of information:
- A type for: we don't want to change the attribute.
- A type for: we want to change the attribute to `None`.
- A type for: we want to change the attribute to something.

### Solution

To go around the nested option problem, this PR creates a special ternary type for the `ProgrammableConfig` attribute.

```ts
pub enum ProgrammableConfigOpt {
    Some(ProgrammableConfig),
    None,
    Unchanged,
}
```

This fixes our problem as Solita will properly generate this as a data enum like so:

```ts
type ProgrammableConfigOpt =
  | { __kind: 'Some', fields: [ProgrammableConfig] }
  | { __kind: 'None' }
  | { __kind: 'Unchanged' }
```

### Remarks

Note that this only fixes our problem for the `ProgrammableConfig` attribute. Going forward, we would need to create a matching `*Opt` enum for each attribute that can be changed to `None`.

This could be achieved via macros in the future to streamline the process.

It is also worth noting that a new client generator is on its way and will provide a better way of handling `Option` types in JavaScript by representing options as data enums like so:

```ts
type Option<T> =
  | { __option: 'Some', value: T }
  | { __option: 'None' }
```

That way, `Some(None)` will be represented in TypeScript as `{ __option: 'Some', value: { __option: 'None' } }` without losing any information. Some helper methods will also be offered to improve the DX around dealing with options in TypeScript.

lorisleiva (approved):

PR #909, opened 2022-12-10
Agressive Dep Upgrade to reach 1.14

austbot:

PR #931, opened 2023-01-03
token-metadata: Update to Solana v1.14.10

joncinque:
#### Problem

In order to eventually support token-2022 in token-metadata, the underlying dependencies should match the token-2022 crate, which is currently at 1.14.10.

#### Solution

In order to keep changes separated, this started with updating just token-metadata to 1.14, along with other associated crates, spl-token and spl-associated-token-account.  For now, this doesn't update anything else, since the repo and CI use many different versions.

While making the small changes, I ran `cargo +nightly fmt` to respect the `rustfmt.toml` at the top-level, which brought on some other changes. It seems like CI isn't respecting `rustfmt.toml` since it isn't using a nightly version, which is required to set things like `imports_granularity = Crate`.

### DIFF
--- a/token-metadata/Cargo.lock
+++ b/token-metadata/Cargo.lock
@@ -1698,9 +1698,9 @@ dependencies = [
 
 [[package]]
 name = "lz4"
-version = "1.23.3"
+version = "1.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


joncinque:
Just realizing that #909 takes care of this. Happy to help out over there if that's the preferred approach!

joncinque:
Closing this in favor of #949

PR #932, opened 2023-01-03
ci: Update Solana to 1.14.10

joncinque:
#### Problem

#931 upgrades token-metadata to use Solana version 1.14.10, but this brings the CI version out of sync with the program.

#### Solution

Upgrade CI to use the Solana version 1.14.10. Note that this still keeps things out of sync, since many individual programs use older versions, but this better replicates how the real network behaves.


joncinque:
Just realizing that #909 addresses this, so I'm happy to close this and help out on the other one if that's better!

joncinque:
Closing this in favor of #949

PR #935, opened 2023-01-06
Split instruction match in processor

febo:
This PR divides the instruction match in the processor to add a constraint that `ProgrammableNonFungible` assets can only be manipulated by the new instruction handlers.


febo:
Wrong target branch.

PR #940, opened 2023-01-08
handle burned collection parents in unverify_collection

samuelvanderwaal:
The `unverify` handler requires the collection metadata account in order to allow unverifying. This means if a collection parent is burned, any verified items are locked in that collection and cannot be migrated out or into a new collection. 

This PR modifies the `unverify` handler to check if a collection parent is burned and allows unverifying in that case as long as appropriate checks are met.


token-metadata/program/src/processor/collection/unverify_collection.rs
```diff
@@ -1,71 +1,122 @@
...
     let using_delegated_collection_authority = accounts.len() == 6;
 
-    assert_signer(collection_authority_info)?;
+    // Account validation.
     assert_owned_by(metadata_info, program_id)?;
-    assert_owned_by(collection_info, program_id)?;
-    assert_owned_by(collection_mint, &spl_token::id())?;
-    assert_owned_by(edition_account_info, program_id)?;
+    assert_signer(collection_authority_info)?;
+    assert_owned_by(collection_mint_info, &spl_token::id())?;
```
blockiosaurus, line 32: Does the mint account persist even after a full burn? Wouldn't it be returned to the system program?

samuelvanderwaal: Mints from the original SPL token program cannot be closed. If we add support for SPL_Token_2022 then we would need to change this logic.

blockiosaurus (dismissed):

token-metadata/program/src/processor/collection/unverify_collection.rs
```diff
@@ -1,71 +1,122 @@
...
+        // respected as there's currently no way to revoke them.
+
+        if metadata.update_authority != *collection_authority_info.key {
+            return Err(MetadataError::UpdateAuthorityIncorrect.into());
+        }
+
+        metadata.collection.as_mut().unwrap().verified = false;
+        clean_write_metadata(&mut metadata, metadata_info)?;
+        return Ok(());
+    }
```
danenbm, line 85: Style nit:  Instead of setting `collection.verified = false` and `clean write metadata` in this block and then returning early, I would actually put the rest of this code in an else statement, except have the `collection.verified = false` and `clean write metadata` at the end, outside of the if-else block.  To me that is slightly clearer control flow.  But this is a style nit based on my opinion so feel free to leave it as-is.

token-metadata/program/tests/collection_migrations.rs
```diff
@@ -774,4 +779,360 @@ mod collection_migrations {
...
+            collection_a_nft.token.pubkey(),
+            collection_a_me.pubkey,
+            None,
+        )
+        .await
+        .unwrap();
+
+        // This should succeed.
+        nft_item_one
+            .unverify_collection(
```
danenbm, line 945: It may be slightly confusing that this handler works with a burned sized parent, because yeah I know the collection is gone, but it seems like if I was trying to use this even if I knew it was burned, I might reach for `unverify_sized_collection` first.  It might be worth considering making that handler able to handle burned parents as well, what do you think?

token-metadata/program/src/processor/collection/unverify_collection.rs
```diff
@@ -1,71 +1,122 @@
...
-    )?;
+
+    // First, if there's no collection set, we can just short-circuit
+    // since there's nothing to unverify.
+    if metadata.collection.is_none() {
+        return Ok(());
+    }
+
+    // Unwrap ok because of check above.
+    let collection = metadata.collection.clone().unwrap();
```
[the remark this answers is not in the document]
samuelvanderwaal: Good call.

danenbm (changes requested):

token-metadata/program/src/processor/collection/unverify_collection.rs
```diff
@@ -1,71 +1,115 @@
...
+        && collection_metadata_info.owner == &system_program::ID;
+
+    if parent_burned {
+        // If the parent is burned, we need to check that the authority
+        // is the update authority on the item metadata.
+        //
+        // Collection Delegates for burned collection parents should not be
+        // respected as there's currently no way to revoke them.
+
+        if metadata.update_authority != *collection_authority_info.key {
```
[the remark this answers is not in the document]
samuelvanderwaal: Changing Shank annotation names is a breaking API change, though, which I'd like to avoid. I can change the name internally to be more clear but don't want to change the name in the instruction API in either Rust or JS.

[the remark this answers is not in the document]
danenbm: If its internal then maybe its the scenario where you are looking at the code anyways, so maybe it doesn't add much of a benefit.  I understand about not breaking API change with shank annotation

token-metadata/program/tests/collection_migrations.rs
danenbm: wdyt about this comment above?

samuelvanderwaal: added in the latest commit

token-metadata/program/src/processor/collection/unverify_collection.rs
```diff
@@ -1,71 +1,115 @@
...
     }
 
-    // If the NFT has collection data, we set it to be unverified and then update the collection
-    // size on the Collection Parent.
-    if let Some(collection) = &mut metadata.collection {
-        collection.verified = false;
+    // We need to ensure the metadata is derived from the mint so
+    // someone cannot pass in a burned metadata account associated with
+    // a different mint.
+    assert_metadata_derivation(program_id, collection_metadata_info, collection_mint_info)?;
```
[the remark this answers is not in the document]
samuelvanderwaal: Edition account is not validated against the mint but it's also unused in both these handlers.

danenbm (approved):
LGTM left a comment for you to read making sure I understand the account verification changes

blockiosaurus (dismissed):

token-metadata/program/src/processor/collection/unverify_collection.rs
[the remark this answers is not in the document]
danenbm: >assert collection verify is valid Does two things:

`assert collection verify is valid` has a third check which it ensures `collection_metadata.mint == collection_mint.key`.

 Since you are no longer calling `assert_collection_verify_is_valid` and now do one of the checks on line 55,  the third check is no longer done, nor is there the edition derivation.

My point in the comment was that both of those checks are not required because they are validations of the collection NFT itself.

And the checks you do have make sense to cover the NFT belongs to the collection and matches the collection mint.

danenbm (dismissed):

danenbm (approved):

PR #943, opened 2023-01-09
Add asset state to metadata

febo:
This PR adds an `asset_state` to the Metadata struct so that pNFTs can be locked/unlocked – this behaviour is supported for NFTs as well, in combination with the spl-token state.


token-metadata/program/src/instruction/mod.rs
```diff
@@ -578,19 +578,20 @@ pub enum MetadataInstruction {
...
     #[account(7, name="system_program", desc="System program")]
-    #[account(8, optional, writable, name="use_authority_record", desc="Use Authority Record PDA (if present the program assumes a delegated use authority)")]
-    #[account(9, optional, name="authorization_rules", desc="Token Authorization Rules account")]
+    #[account(8, name="sysvar_instructions", desc="System program")]
+    #[account(9, optional, name="spl_token_program", desc="SPL Token Program")]
     #[account(10, optional, name="authorization_rules_program", desc="Token Authorization Rules Program")]
+    #[account(11, optional, name="authorization_rules", desc="Token Authorization Rules account")]
     #[default_optional_accounts]
-    UseAsset(UseAssetArgs),
+    Utility(UtilityArgs),
```
austbot, line 594: what prompts the name change here, why "utilize"? or for you EUROs "utilisssssssse"

token-metadata/program/src/instruction/utility.rs
```diff
@@ -180,10 +182,71 @@ pub fn utilize(
 #[repr(C)]
 #[cfg_attr(feature = "serde-feature", derive(Serialize, Deserialize))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Debug, Clone)]
-pub enum UseAssetArgs {
-    V1 {
-        use_count: u64,
+pub enum UtilityArgs {
+    LockV1 {
```
austbot, line 186: I think is missing the old utilize stuff and is overloading the term.

do we just add Use here too.

Lock, Unlock, Use, ?

austbot (changes requested):
im confuised on how this impacts the actual use methods from long ago

token-metadata/program/src/instruction/utility.rs
febo: That is a good question – I think all "use" relating things go here, so that is the motivation for the rename to `Utility`. We haven't ported the "use" things yet.

token-metadata/program/src/instruction/mod.rs
febo: `Utilize` is already taken by the "legacy" processor.

token-metadata/program/src/instruction/utility.rs
febo: These have been renamed to `lock` and `unlock`. Both were "promoted" to top-level instructions.

token-metadata/program/src/instruction/mod.rs
febo: Some as below: "These have been renamed to lock and unlock. Both were "promoted" to top-level instructions."

PR #944, opened 2023-01-10
refactor transfer

samuelvanderwaal:


token-metadata/program/src/processor/metadata/transfer.rs
```diff
@@ -142,49 +167,36 @@ fn transfer_v1(program_id: &Pubkey, ctx: Context<Transfer>, args: TransferArgs)
...
         false
     };
 
-    // Wallet-to-wallet transfer are always allowed except if a sale
-    // delegate is set.
-    let is_wallet_to_wallet = ctx.accounts.token_owner_info.owner == &system_program::ID
-        && ctx.accounts.destination_owner_info.owner == &system_program::ID;
-
     // Determine transfer type.
     let transfer_authority = if let Some(delegate_role) = metadata.persistent_delegate {
```
febo, line 187: There is now a "generic" function to determine the authority of a request `AuthorityType::get_authority_type`. This will check whether it is the `Holder` (token owner), `Delegate` or `Metadata` (update authority). Wonder if this could/should be used here so we have a consistent way to determine the authority.

token-metadata/program/tests/utils/programmable.rs
```diff
@@ -17,62 +18,117 @@ use solana_sdk::{
...
+        dest_owned_by_sys_program,
+        nft_amount,
+    );
 
-    // Rule for Transfers: Allow transfers to a Token Owned Escrow account.
-    let owned_by_token_metadata = Rule::ProgramOwned {
-        program: mpl_token_metadata::ID,
-        field: PayloadKey::Destination.to_string(),
+    // (source is owned by system program && dest is on allow list && destination is a PDA) ||
+    // (source is on allow list && source is a PDA && dest is owned by system program)
```
danenbm, line 91: This does not seem like the RuleSet you want for default as it limits dest to PDA only.   That was like a draft, but I have since updated it.  Do you want the new one that is?:
```
// amount is 1 &&
// (source is owned by system program || (source is on allow list && source is a PDA) &&
// (dest is owned by system program || (dest is on allow list && dest is a PDA)
```

See this update:
https://github.com/metaplex-foundation/mpl-token-auth-rules/blob/main/program/tests/basic_royalty_enforcement.rs#L88-L115

token-metadata/program/tests/utils/programmable.rs
```diff
@@ -96,8 +152,92 @@ pub async fn create_test_ruleset(
 
     // Client can add additional rules to the Payload but does not need to in this case.
     let payload = Payload::new();
-
     let auth_data = AuthorizationData { payload };
 
     (ruleset_addr, auth_data)
 }
+
```
danenbm, line 159: Nitpick: I normally nitpick and ask my team to not have commented out code checked in, but I'm not the authority so feel free to leave it here if you want it for reference.  I just personally prefer to add it back in when its ready to be uncommented/active.

danenbm (changes requested):

token-metadata/program/src/state/programmable.rs
```diff
@@ -70,35 +70,33 @@ impl AuthorityType {
...
     Amount,
     Authority,
+    Delegate,
     Destination,
+    DestinationSeeds,
     Holder,
-    Delegate,
+    Source,
+    SourceSeeds,
     Target,
```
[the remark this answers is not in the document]
samuelvanderwaal: Good call.

token-metadata/program/tests/utils/programmable.rs
samuelvanderwaal: I think my text just doesn't match what the actual rule is. If you look at the code the rule should be what you describe in your text above.

token-metadata/program/src/state/programmable.rs
```diff
@@ -70,35 +70,33 @@ impl AuthorityType {
...
-            Operation::Update => "Update",
+            Self::Transfer { scenario } => format!("TRA:{}", scenario),
+            Self::Update { scenario } => format!("UPD:{}", scenario),
         }
-        .to_string()
     }
 }
 
 #[derive(Debug, Clone, ToPrimitive)]
 pub enum PayloadKey {
```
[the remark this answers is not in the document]
samuelvanderwaal: Seems worth future-proofing. Added.

PR #945, opened 2023-01-10
Instructions breakdown and optional account details

febo:
This PR adds a list of operations for each instruction with an indication of whether they are completed or not. It also includes details of how optional positional accounts are used, which is important for users creating instructions without using the provided SDK/Rust crate.


token-metadata/program/ProgrammableNFTGuide.md
```diff
@@ -69,22 +73,112 @@ pub enum MetadataInstruction {
...
+    * [X] Revoke of `Update` delegates
+    * [ ] Revoke of `Use` delegates
+    * [X] Revoke of `Utility` delegates
+- [ ] `Transfer`
+    * [X] wallet-to-wallet transfers
+    * [X] wallet-to-program transfers
+    * [X] program-to-wallet transfers
+- [X] `Update`
+    * [X] Update metadata details for Programmable Non-Fungibles
+    * [X] Update metadata details for Non-Fungibles
```
lorisleiva, line 121: I'd probably add `FungibleAssets` and `Fungibles` to stay consistent with the other instructions. Otherwise, it looks like the `Update` instruction is only for non-fungibles.

lorisleiva (approved):

token-metadata/program/ProgrammableNFTGuide.md
febo: Thanks - added both `FungibleAssets` and `Fungibles`.

token-metadata/program/ProgrammableNFTGuide.md
```diff
@@ -69,22 +73,115 @@ pub enum MetadataInstruction {
...
+// Accounts relative position:
+0. token
+1. tokenOwner
+2. metadata
+3. masterEdition
+4. mint
+5. payer
+...
+```
+When you are minting from a semi-fungible token, there is no need to pass a `masterEdition` account (semi-fungibles do not have a master edition account associated). If we simple ommit the `masterEdition` account, the relative position of the remaining accounts (the accounts after the master edition) would change, resulting in the program logic to be inconsistent. One way to address this is to set another `PublicKey` value to represent a "not-set-value" to maintain the position but at the same time indicate that the master edition account was not set. This is accomplished by setting the Token Metadata program key as the `PublicKey` for any account that should be ommited. This is an efficient approach since:
```
danenbm, line 157: /sp "simple ommit" should be "simply omit"

danenbm: Also I suppose you could also mention that this is somewhat of a standard approach?  I believe it is as Solana has suggested it to me on a different program.  Feel free to not say that though, it was just a thought.

danenbm (approved):

lorisleiva (approved):

PR #948, opened 2023-01-12
aggressive update: Fix test failures

joncinque:
#909 runs into some errors on tests because of some incompatibilities between 1.14 and 1.10.

These extra commits should take care of it


joncinque:
Closing this in favor of #949

PR #949, opened 2023-01-12
deps: Update to 1.14

joncinque:
#### Problem

As its label clearly states, #909 is an absolute pain, but it's better to be on 1.14.

#### Solution

This branch rebases #909 onto master and adds #948. The only rebase conflict was the token-metadata crate version update to 1.7. Gonna cross my fingers that this works!


joncinque:
Welp, so much for that. Just a sec...

joncinque:
This is a bit messy, since some of the tests are running against the newest programs with old instruction creators. For example, token-entangler is using the deprecated `create_metadata_accounts` instruction, which is removed in this PR, so of course it fails. If I try to update the crates, anchor falls out of sync. I'll look for the "happy medium" version that satisfies everything.

joncinque:
I think I found it! Let's see how CI reacts...

joncinque:
Ok, we're making progress! The "program tests" errors are all due to calling the now-removed `create_metadata_accounts` or `create_metadata_accounts_v2` instruction, and the failing integration test should be pretty ok to clean up.

joncinque:
Gonna leave it here for now, since I'm not sure what's going on with that last failure.

@austbot apologies for the ping, but this continues your work and is just about ready to be merged!

You can follow through all of the commits pretty much in order to see what was done. Most of it is updating everything to use `create_metadata_accounts_v3` everywhere since you removed the deprecated instruction processors, so the changes are very straightforward.

The only part that seemed weird were the `set_token_standard` tests. With the deprecated instructions gone, it seems kind of obviated, so the tests are a bit less useful in this state.

Anyway, let me know what you think!

joncinque:
We did it, this is good to go! absolute-pain, meet head-banged-against-wall-until-ci-green

joncinque:
Closing since the upgrade was done with #947

PR #950, opened 2023-01-13
refactor wallet-to-wallet to be more robust

samuelvanderwaal:


febo (approved):
LGTM

danenbm (commented):
LGTM but do we also need to update your RuleSet to allow PDA to PDA?

PR #953, opened 2023-01-16
Check token status on processor

febo:
This PR adds a check on the processor on the `TokenRecord` account to determine whether a token is locked or not. If the token is locked, the only operation available involving a token account is `Unlock`. It also disables the ability of a token owner to lock/unlock to avoid conflicts with delegates.


token-metadata/program/src/processor/mod.rs
```diff
@@ -435,3 +450,24 @@ fn has_programmable_metadata<'a>(
 
     Ok(false)
 }
+
+/// Checks if the instruction's accounts contain a pNFT metadata.
+///
+/// We need to determine if we are dealing with a pNFT metadata or not
```
samuelvanderwaal, line 456: These comments need slight updating to reference checking for Token Record instead of Metadata account.

token-metadata/program/src/processor/mod.rs
```diff
@@ -435,3 +450,24 @@ fn has_programmable_metadata<'a>(
...
+/// so we can restrict the available instructions.
+fn is_locked<'a>(program_id: &Pubkey, accounts: &'a [AccountInfo]) -> Result<bool, ProgramError> {
```
samuelvanderwaal, line 458: This does not need to return a `Result` as there's no possible error condition. Just return a `bool` and that will also make the logic in the processor cleaner.

samuelvanderwaal (changes requested):

token-metadata/program/src/processor/mod.rs
febo: Done.

token-metadata/program/src/processor/mod.rs
febo: Good point, returning a `bool` now.

samuelvanderwaal (approved):

PR #956, opened 2023-01-17
implement nft -> pnft migration case

samuelvanderwaal:
This PR adds a migration method to allow NFTs to be upgraded to pNFTs via a CPI from the `mpl-migration-validator` program. Other migration methods are not supported at this time.


token-metadata/program/src/processor/metadata/migrate.rs
```diff
@@ -127,42 +157,78 @@ pub fn migrate_v1(program_id: &Pubkey, ctx: Context<Migrate>, args: MigrateArgs)
...
             }
 
-            // IDK how this would be possible, but if we have the freeze
-            // authority and it's the right kind of token we can just
-            // unfreeze it before migration.
-            if token.is_frozen() {
-                thaw(
+            // NFT --> PNFT migration must maintain the current level of functionality
+            // that the token has, but all pNFTs must be frozen. To accomplish this,
+            // we assign Utility delegate to any pNFTs that have a SPL token delegate
```
austbot, line 162: migration

token-metadata/program/src/processor/metadata/migrate.rs
```diff
@@ -127,42 +157,78 @@ pub fn migrate_v1(program_id: &Pubkey, ctx: Context<Migrate>, args: MigrateArgs)
...
-            if metadata.collection.is_none() {
-                return Err(MetadataError::NotAMemberOfCollection.into());
-            }
-            let collection = metadata.collection.as_ref().unwrap();
-            if collection.key != collection_metadata.mint || !collection.verified {
-                return Err(MetadataError::NotVerifiedMemberOfCollection.into());
+            // Set Utility delegate if SPL delegate is set.
+            if let COption::Some(current_delegate) = token.delegate {
+                token_record.delegate = Some(current_delegate);
+                token_record.delegate_role = Some(TokenDelegateRole::Utility);
```
austbot, line 222: Migration

austbot (changes requested):
need to replace utility delegate with migration,

do need to check here or in the other migrator to make sure the NFT does not have any printable supply ?

samuelvanderwaal:
> need to replace utility delegate with migration,
> 
> do need to check here or in the other migrator to make sure the NFT does not have any printable supply ?

I replaced the Utility check with Migration. I don't think we need to restrict migration based on printable supply as long as it's a `NonFungible` it should be allowed to migrate. There may be creators who want to migrate their Master Editions in advance of being able to make pNFT prints from them in the future.

PR #959, opened 2023-01-18
Add optional update authority to revoke

febo:
PR to add an optional `previous_authority` to allow metadata delegates self-revoke. It also renames the transaction signer from `approver` to `authority` to match the name used in other instructions.


febo:
This change is not relevant anymore.

cc: #963

PR #960, opened 2023-01-18
Add length check on edition byte test

febo:
Add a check on the length of the edition account before indexing for the programmable byte test.


austbot (approved):
Lgtm this pr limits programmable checks to master editions

PR #961, opened 2023-01-18
fix bug in transfer when no rule set present on NFT

samuelvanderwaal:
Previously, `assert_valid_authorization` would fail because the auth rules validation function was only checking if a programmable config present and not if an actual rule set was set. This PR fixes this and writes a test for the behavior.


austbot (approved):
This pr checks if the programmable config has a ruleset. Previously this transfer function was assuming the ruleset was present if the programma config was present

PR #962, opened 2023-01-18
Add program delegate lock and unlock tests

febo:


samuelvanderwaal:
@febo can this be closed?

febo:
These tests are not needed as the functionality is already tested in other test cases.

PR #965, opened 2023-01-22
Token address on token record seed

febo:
PR to use the token account address as part of the token record seed, given that there is a 1:1 relationship between token account and token record.


samuelvanderwaal:
This PR changes the `token_record` derivation to use the token account pubkey instead of the token owner pubkey in order to properly tie the `token_record` directly to the token account. It updates the token record derivation in all the places it needs to be updated and updates tests as well.

samuelvanderwaal (approved):

PR #968, opened 2023-01-23
add ser/deser implementations for pubkeys

samuelvanderwaal:
Allow conditional compilation serde implementations for `Pubkey`s to serialize into and from strings instead of `u8` arrays.

This adds support for most public `Pubkey` and `Option<Pubkey>` types, but not all of them. Escrow, notably, is left out as it has a wrapper around its Pubkey. We can add support for this later as it won't be a breaking change.


febo (approved):
This PR adds a serde-with feature flag on pubkey fields to allow clients to deserialize/serialize pubkeys from/to strings.

PR #969, opened 2023-01-23
Add enum to represent the print supply

febo:
This PR adds an enum to represent the print supply when creating a metadata account of non-fungible assets:
```rust
/// Represents the print supply of a non-fungible asset.
pub enum PrintSupply {
    /// The asset does not have any prints.
    Zero,
    /// The asset has a limited amount of prints.
    Limited(u64),
    /// The asset has an unlimited amount of prints.
    Unlimited,
}
```


samuelvanderwaal (approved):
This PR changes the `max_edition` field of CreateArgs to be an `Option<PrintSupply>` to more clearly represent the options for this field on-chain. The `PrintSupply` enum allows specifying the maximum number of prints in user friendly variants where as the `None` variant of the `Option` ignores the `PrintSupply` field.

PR #971, opened 2023-01-24
No default for optional accounts

febo:
This PR changes the way the macro handles default values for optional accounts. Currently, a default value will be added when the optional account is not set. This creates an undesirable behaviour since optional accounts will be present in the transaction: (1) they will take more space on the transaction; and (2) when used on a CPI call, they will need to be added to the outer instruction.

The solution is to only use the default value for required accounts, when accounts are not set.

For required accounts, the code generated is:
```rust
system_program: self.system_program.unwrap_or(solana_program::system_program::ID)
```

For optional accounts, the code generated is (even when the name of the account matches one of the default ones):
```rust
authorization_rules_program: self.authorization_rules_program
```


samuelvanderwaal (approved):
This PR changes the macro functionality to no longer automatically set the value for optional accounts. This will save transaction bytes in calls to handlers when optional accounts are not required.

PR #972, opened 2023-01-24
Improve errors on instructions

febo:
This PR adds more JS tests and improves errors from the program.

Two important errors were added:
1. Trying to mint `0` (zero) tokens now generates an error;
2. Trying to create a metadata with `TokenStandard::NonFungibleEdition` generates an error at the start of the instruction.


samuelvanderwaal (approved):
This PR adds a few new errors to improve accuracy of error messages in the `create` and `mint` handlers, refactors a few code patterns, and checks that the users is not trying to mint an amount of "0" tokens, and fails with an error in that case.

PR #973, opened 2023-01-24
Refactor/remove deprecated fn

samuelvanderwaal:


febo (approved):
This PR removes a deprecated instruction and related dependency – this provides a small reduction on the size of the binary. As part of the changes, imports have been refactored.

PR #975, opened 2023-01-25
Add transfer test with lookup table

febo:
This PR adds a test showing how an address lookup table (LUT) can be used to reduce the transaction size of the transfer instruction. The test adds 8 accounts to the LUT:
1. `owner`
2. `owner token record`
3. `token`
4. `mint`
5. `metadata`
6. `master edition`
7.  `rule set PDA`
8. `SYSVAR_INSTRUCTIONS_PUBKEY`

Bigger transaction size reductions can be achieved if more accounts are added to the LUT.


samuelvanderwaal (approved):
Looks good!

PR #980, opened 2023-01-27
Features/allow payouts even if hydra shares not exhausted

staccDOTsol:
so what's a pnft


staccDOTsol:
hehehe @stegaBOB oops I should not merge this here? this is mostly your work on fix/readme

staccDOTsol:
https://github.com/metaplex-foundation/metaplex-program-library/pull/1054

PR #982, opened 2023-01-28
Add precedence field to authority request

febo:
Currently the method to determine the authority type of a request has a fixed order. This does not work in all cases, since an address (authority) can match multiple authority types.

This PR adds a `precedence` field to the `AuthorityRequest` so the caller code can specify the order for the evaluation of the type. For example, passing the precedence `[AuthorityType::Metadata, AuthorityType::Delegate]` will make the evaluation to try to match the authority to the update authority; then it would try to match to a delegate.

### DIFF
--- a/token-metadata/program/src/state/programmable.rs
+++ b/token-metadata/program/src/state/programmable.rs
@@ -145,6 +145,8 @@ pub enum AuthorityType {
 }
 
 pub struct AuthorityRequest<'a, 'b> {
+    /// Determines the precedence of authority types.
+    pub precedence: &'a [AuthorityType],
     /// Pubkey of the authority.
     pub authority: &'a Pubkey,
     /// Metadata's update authority pubkey of the asset.
@@ -168,6 +170,11 @@ pub struct AuthorityRequest<'a, 'b> {
 impl<'a, 'b> Default for AuthorityRequest<'a, 'b> {
     fn default() -> Self {
         Self {
+            precedence: &[
+                AuthorityType::Delegate,
+                AuthorityType::Holder,
+                AuthorityType::Metadata,
+            ],
             authority: &DEFAULT_PUBKEY,
             update_authority: &DEFAULT_PUBKEY,
             mint: &DEFAULT_PUBKEY,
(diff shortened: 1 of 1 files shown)


samuelvanderwaal (approved):
This PR modifies the `get_authority_type` method on the `AuthorityRequest` struct to take in an array of `AuthorityType`s as a precedence list. It then goes over each item in the list and searches for a match returning immediately, if found, to avoid running checks more than once. 

This change allows callers of the method to specify their individual precedence lists while using the same method. Specifically, this lets a handler like `Update` specify that `MetadataDelegate` should be of higher priority than a `Delegate` type.

lorisleiva:
Why not just ask the caller to explicitly pass the authority type it's trying to use?

In an previous version we were doing this and I thought it was much better as the client could explicitly tell the program what type of authority it is providing instead of using "magical" heuristics to automatically recognise it.

febo:
> Why not just ask the caller to explicitly pass the authority type it's trying to use?
> 
> In an previous version we were doing this and I thought it was much better as the client could explicitly tell the program what type of authority it is providing instead of using "magical" heuristics to automatically recognise it.

Regardless whether the client explicitly tells the type of the authority or not, this check is needed in the program to validate the authority type. The program cannot "trust" that the client has the requested authority. 

Since the check is required, the information from the client is somewhat redundant. We could ask the client to pass it as a parameter, but this would be only to validate that the "heuristically" determined matches the client one. 

The cases that there are ambiguity are very specific - e.g., the update authority is also the owner of the NFT; or the owner is also a delegate. These tend to happen in our tests.

samuelvanderwaal:
> Why not just ask the caller to explicitly pass the authority type it's trying to use?
> 
> In an previous version we were doing this and I thought it was much better as the client could explicitly tell the program what type of authority it is providing instead of using "magical" heuristics to automatically recognise it.

You still have to check what the client passes and then it's an extra data field taking up precious transaction space.

lorisleiva (approved):
My bad, I thought this `precedence` attribute was going to creep into instruction arguments. I realise now this is only for internal use.

PR #986, opened 2023-01-31
Removing inline attributes from mpl-utils

blockiosaurus:


samuelvanderwaal (approved):
This PR removes the forced inlining of `create_or_allocate_account_raw`, `resize_or_reallocate_account_raw`, and `close_account_raw` functions to reduce binary sizes of programs that include this library.

febo (approved):
LGTM

PR #990, opened 2023-02-01
MPL Testing Utils, AuctionHouse, Auctioneer: Upgrade Deps

samuelvanderwaal:
This updates `mpl-testing-utils`, `fixed-price-sale`, `auction-house` and `auctioneer` dependencies to Solana 1.14 and Anchor 0.26. `mpl-testing-utils` will need a new version published to crates and then `auction-house` and `auctioneer` will need to bump their versions to the new one.

### DIFF
--- a/auction-house/program/Cargo.toml
+++ b/auction-house/program/Cargo.toml
@@ -21,24 +21,24 @@ cpi = ["no-entrypoint"]
 default = []
 
 [dependencies]
-solana-program = "1.10"
-anchor-lang = "0.25.0"
-anchor-spl = "0.25.0"
-spl-token = { version = "3.2",  features = ["no-entrypoint"] }
-spl-associated-token-account = {version = "1.0.5", features = ["no-entrypoint"]}
-mpl-token-metadata = { version="1.4", features = [ "no-entrypoint" ] }
+solana-program = "1.14"
+anchor-lang = "0.26.0"
+anchor-spl = "0.26.0"
+spl-token = { version = "3.5",  features = ["no-entrypoint"] }
+spl-associated-token-account = {version = "1.1.1", features = ["no-entrypoint"]}
+mpl-token-metadata = { version="1.7", features = [ "no-entrypoint" ] }
 thiserror = "1.0"
 arrayref = "0.3.6"
 
 [dev-dependencies]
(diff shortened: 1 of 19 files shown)


blockiosaurus (approved):
This PR is updating the Auction House, Auctioneer, and FPS dependencies to the latest Anchor and Solana minor versions.

auction-house/program/Cargo.toml
```diff
@@ -21,24 +21,24 @@ cpi = ["no-entrypoint"]
...
-anchor-spl = "0.25.0"
-spl-token = { version = "3.2",  features = ["no-entrypoint"] }
-spl-associated-token-account = {version = "1.0.5", features = ["no-entrypoint"]}
-mpl-token-metadata = { version="1.4", features = [ "no-entrypoint" ] }
+solana-program = "1.14"
+anchor-lang = "0.26.0"
+anchor-spl = "0.26.0"
+spl-token = { version = "3.5",  features = ["no-entrypoint"] }
+spl-associated-token-account = {version = "1.1.1", features = ["no-entrypoint"]}
+mpl-token-metadata = { version="1.7", features = [ "no-entrypoint" ] }
```
danenbm, line 29: Should this be token-metadata be 1.8 now?

For that matter, should token-metadata also be updated to use auth rules 1.2.0?  I don't think auth rules changes are significant for this,

But I'm just wondering if you are going through the trouble of updating everything should we put them all at the latest now?

danenbm (approved):
LGTM but had one comment

auction-house/program/Cargo.toml
samuelvanderwaal: It's a moving target tho, if we wait another week or so to merge this PR it can be updated to 1.9 when `Burn` and `Verify` are released. The goal of this PR was to get everything up to Anchor 0.26 , Solana 1.14 and a version of Token Metadata that supports pNFTs. The 1.7 dependency will pull newer minor versions if necessary and any future updates that specifically require it can bump.

samuelvanderwaal: For example, AH still needs to be updated to support pNFTs so the dependency can be bumped to whatever the current version of Token Metadata is at that point.

PR #993, opened 2023-02-01
Add LockedTransfer delegate

febo:
## Context

In specific scenarios it is necessary to create a delegate to both lock the token and transfer it if certain conditions are met. Currently, there is no delegate that can perform both operations.

## Proposed Solution

This PR create a new delegate role `LockedTransfer`, which is a delegate that can perform `Lock` and `Unlock` operations, and transfer the token to a predetermined address. The address is specified at the time the delegate is approved:
```rust
LockedTransferV1 {
    amount: u64,
    locked_address: Pubkey,
    authorization_data: Option<AuthorizationData>,
},
```
The `locked_address` is stored on the `TokenRecord` account. The transfer instruction then checks if the authority requesting the transfer is a `LockedTransfer` delegate and if it is, the transfer is only allowed if the destination address matches the `locked_address`.
```rust
pub struct TokenRecord {
    ...
    /// Stores the destination pubkey when a transfer is lock to an allowed address. This
    /// pubkey gets set when a 'LockTransfer' delegate is approved.
    pub locked_transfer: Option<Pubkey>,
}
```

To maintain compatibility with existing `TokenRecord` accounts, given that a new field was added to the struct, the PR includes a new `Resizable` trait. Implementing this trait automatically provides a `save` method implementation that resizes the account (if needed) before saving the struct data. There is also a "manual" deserialization method that handles variable account sizes.

## Other Considerations

* Client changes: the PR includes a custom deserializer to support "old" TokenAccount deserialization; this should be transparent to clients.
* Security: the `Transfer` instruction includes an explicit check of the destination address when the transfer is initiated by a `LockedTransfer` delegate.


token-metadata/program/src/state/mod.rs
```diff
@@ -156,3 +163,83 @@ where
...
+        let required_size = Self::size();
+
+        if account_info.data_len() != required_size {
+            // no risk of overflow here since max account size is 10_000_000
+            let difference = required_size as i64 - account_info.data_len() as i64;
+            let snapshot = account_info.lamports();
+
+            if difference > 0 {
+                if difference as usize > MAX_PERMITTED_DATA_INCREASE {
+                    //return err!(CandyGuardError::DataIncrementLimitExceeded);
```
samuelvanderwaal, line 195: Looks like you missed adding a `MetadataError` here.

token-metadata/program/src/state/mod.rs
```diff
@@ -156,3 +163,83 @@ where
...
+                msg!("Funding {} lamports for account realloc", lamports_diff);
+
+                invoke(
+                    &system_instruction::transfer(payer.key, account_info.key, lamports_diff),
+                    &[payer.clone(), account_info.clone(), system_program.clone()],
+                )?;
+            } else {
+                let lamports_diff = snapshot
+                    .checked_sub(Rent::get()?.minimum_balance(required_size))
+                    .ok_or(MetadataError::AddressNotInReservation)?;
```
samuelvanderwaal, line 212: This should be `MetadataError::NumericalOverflowError`.

token-metadata/program/src/state/mod.rs
```diff
@@ -156,3 +163,83 @@ where
...
+                msg!(
+                    "Withdrawing {} lamports from account realloc",
+                    lamports_diff
+                );
+
+                **account_info.lamports.borrow_mut() = snapshot - lamports_diff;
+                **payer.lamports.borrow_mut() = payer
+                    .lamports()
+                    .checked_add(lamports_diff)
+                    .ok_or(MetadataError::AddressNotInReservation)?;
```
samuelvanderwaal, line 223: This should be `MetadataError::NumericalOverflowError`.

samuelvanderwaal (changes requested):

token-metadata/program/src/state/mod.rs
febo: Fixed.

token-metadata/program/src/state/mod.rs
febo: Fixed.

token-metadata/program/src/state/mod.rs
febo: Fixed.

token-metadata/program/src/state/mod.rs
```diff
@@ -156,3 +162,74 @@ where
...
+    fn save<'a>(
+        &self,
+        account_info: &'a AccountInfo<'a>,
+        payer: &'a AccountInfo<'a>,
+        system_program: &'a AccountInfo<'a>,
+    ) -> Result<(), ProgramError> {
+        // the required account size
+        let required_size = Self::size();
+
+        if account_info.data_len() != required_size {
```
blockiosaurus, line 187: Can we just use `resize_or_reallocate_account_raw` here?

token-metadata/program/src/state/programmable.rs
```diff
@@ -27,28 +28,45 @@ pub const TOKEN_RECORD_SEED: &str = "token_record";
...
-+ 2; // delegate role
+pub const LOCKED_TRANSFER_SIZE: usize = 33; // Optional Pubkey
 
+pub const TOKEN_RECORD_SIZE: usize = 1 // Key
++ 1   // bump
++ 1   // state
++ 9   // rule set revision
++ 33  // delegate
++ 2   // delegate role
++ 33; // locked transfer
```
blockiosaurus, line 39: We might want to allocate extra padding to prevent the need for additional migration functions in the future.

token-metadata/program/src/state/mod.rs
febo: Good suggestion - done!

token-metadata/program/src/state/programmable.rs
febo: In this case padding is not going to avoid having the manual deserialization since there are already accounts that are shorter and might not get updated in the near future.

samuelvanderwaal (approved):
This PR adds a new delegate type: `LockedTransfer`, which allows both locking an NFT and transferring to a specific pubkey which must be specified when the delegate is created. The pubkey is stored on the TokenRecord and is checked in transfer for this particular delegate.

A new scenario is used for this delegate in the `delegate` handler to allow creators to specify which programs should be allowed to receive the delegate type, but in `transfer` it uses the existing `DelegateTransfer` scenario.

blockiosaurus (approved):

PR #995, opened 2023-02-01
Fix/update auth rule

samuelvanderwaal:
This PR removes the redundant authorization rule account program owned check as it's already validated in the CPI to `mpl-token-auth-rules` and relaxing this check in Token Metadata allows MetadataAuthority to fix invalid rule sets by updating to a new one.

In addition, it checks if a SPL token delegate is set and fails if the authority tries to update a rule set while a SPL token delegate is set.


token-metadata/program/src/processor/metadata/update.rs
```diff
@@ -137,11 +138,19 @@ fn update_v1(program_id: &Pubkey, ctx: Context<Update>, args: UpdateArgs) -> Pro
         token_account: token.as_ref(),
         metadata_delegate_record_info: ctx.accounts.delegate_record_info,
         metadata_delegate_role: Some(MetadataDelegateRole::Update),
+        precedence: &[
+            AuthorityType::Metadata,
+            AuthorityType::Holder,
```
febo, line 143: Does it make sense to have the `Holder` here? What if the holder happens to have a delegate, then this will always evaluate to `Holder`, which might not be desirable.

token-metadata/program/src/state/metadata.rs
```diff
@@ -168,12 +169,24 @@ impl Metadata {
...
         // if the rule_set data is either 'Set' or 'Clear', only allow updating if the
-        // token standard is equal to `ProgrammableNonFungible`
+        // token standard is equal to `ProgrammableNonFungible` and no SPL delegate is set.
         if matches!(rule_set, RuleSetToggle::Clear | RuleSetToggle::Set(_)) {
             if token_standard != TokenStandard::ProgrammableNonFungible {
                 return Err(MetadataError::InvalidTokenStandard.into());
             }
 
+            // Require the token so we can check if it has a delegate.
+            if token.is_none() {
```
febo, line 179: I think here we can use the pattern:
```rust
let token = token.ok_or(MetadataError::MissingTokenAccount)?;
```

token-metadata/program/src/processor/metadata/update.rs
samuelvanderwaal: Good point. Also, only `Metadata` authority type is supported right now so I could just pass that in but for future proofing it I think `Metadata` and `Delegate` are probably sufficient?

token-metadata/program/src/state/metadata.rs
samuelvanderwaal: Yes, nice.

token-metadata/program/src/processor/metadata/update.rs
febo: Yes, agreed.

token-metadata/program/tests/update.rs
```diff
@@ -161,4 +171,286 @@ mod update {
...
+            1,
+        )
+        .await
+        .unwrap();
+
+        let metadata = da.get_metadata(context).await;
+
+        if let Some(ProgrammableConfig::V1 { rule_set }) = metadata.programmable_config {
+            assert_eq!(rule_set, None);
+        }
```
danenbm, line 210: Nit: for the test shouldn't this be more like:
```
let config = metadata.programmable_config.unwrap();
assert_eq!(config.rule_set, None)
```
Rather than ignoring non matching cases?

Or at least panic on else like you do below when it's set?

token-metadata/program/tests/update.rs
```diff
@@ -161,4 +171,286 @@ mod update {
...
+        assert_custom_error!(err, MetadataError::CannotUpdateAssetWithDelegate);
+
+        // Try to update the rule set.
+        let mut update_args = UpdateArgs::default();
+        let UpdateArgs::V1 {
+            rule_set,
+            authorization_data,
+            ..
+        } = &mut update_args;
+        // remove the rule set
```
danenbm, line 422: Nit: outdated comment should read update the ruleset

febo (approved):
This PR relaxes the verification of the auth rules account by allowing the update authority to replace an "invalid" rule set. It also prevents updating the rule set if there is a delegate set to avoid affecting the delegate ability to work in case the new rule set is more restrictive.

danenbm (approved):
I just had a couple nits in the test but overall LGTM

token-metadata/program/tests/update.rs
samuelvanderwaal: Thanks, fixed.

token-metadata/program/tests/update.rs
samuelvanderwaal: Added the `else` statement. It's an enum not a struct so can't read the field directly like you do above.

febo (approved):

danenbm (approved):

PR #996, opened 2023-02-02
Gumdrop: Upgrade Deps

samuelvanderwaal:


gumdrop/program/Cargo.toml
```diff
@@ -19,12 +19,12 @@ cpi = ["no-entrypoint"]
...
-solana-program = "1.9.18"
-spl-token = "3.2.0"
-spl-associated-token-account = { version = "=1.0.3", features = [ "no-entrypoint" ] }
-mpl-token-metadata = { version="=1.2.7", features = [ "no-entrypoint" ] }
+anchor-lang = "0.26.0"
+anchor-spl = "0.26.0"
+solana-program = "1.14.13"
+spl-token = "3.5.0"
+spl-associated-token-account = { version = "1.1.1", features = [ "no-entrypoint" ] }
+mpl-token-metadata = { version="=1.7.0", features = [ "no-entrypoint" ] }
```
danenbm, line 27: Should this be token-metadata be 1.8 now?

For that matter, should token-metadata also be updated to use auth rules 1.2.0? I know auth rules doesn't have anything to do with this change but I'm just wondering if you are going through the trouble of updating everything should we put them all at the latest now?

danenbm (approved):
LGTM but left one question comment

austbot (approved):

PR #998, opened 2023-02-02
Chore/upgrade hydra deps

samuelvanderwaal:


hydra/program/Cargo.toml
```diff
@@ -19,10 +19,10 @@ cpi = ["no-entrypoint"]
...
 
 [dependencies]
-anchor-lang = "0.24.2"
-anchor-spl = "0.24.2"
-spl-token = {version="3.2.0", features = [ "no-entrypoint" ]}
-mpl-token-metadata = {version="1.2.7", features = [ "no-entrypoint" ]}
+anchor-lang = "0.26.0"
+anchor-spl = "0.26.0"
+spl-token = {version="3.5.0", features = [ "no-entrypoint" ]}
+mpl-token-metadata = {version="1.7.0", features = [ "no-entrypoint" ]}
```
danenbm, line 25: Should this be token-metadata be 1.8 now?

For that matter, should token-metadata also be updated to use auth rules 1.2.0? I know auth rules doesn't have anything to do with this change but I'm just wondering if you are going through the trouble of updating everything should we put them all at the latest now?

danenbm (approved):
LGTM but left one question comment

hydra/program/Cargo.toml
austbot: yeah we should get up to the latest

austbot (approved):

PR #1002, opened 2023-02-03
Fix instruction comments

febo:
This PR fix missing accounts in instructions comments. It also fixes the Shank annotation on the migration instruction, where the `token_record` account is writable.


samuelvanderwaal (approved):
This PR fixes instruction comments and Shank annotations to properly reflect what's actually in the code.

blockiosaurus (approved):

PR #1003, opened 2023-02-03
Refactor/token auth payload

samuelvanderwaal:
* Adds precedence back to Update command because we have a test that is broken if Holder isn't present, but Holder is behind Delegate in the precedence list.
* Makes `token_auth_payload` values private to avoid confusion. These exist solely to give Shank values it can extract JS types from.
* Adds a PDA Match JS test and fixes some JS tests that were missing the token account being passed into Update

We need to make sure JS integration tests for Token Metadata are running in our CI but I'll save that for a separate PR since it's not a simple fix, apparently.

### DIFF
--- a/token-metadata/js/test/update.test.ts
+++ b/token-metadata/js/test/update.test.ts
@@ -992,7 +992,7 @@ test('Update: Update pNFT Config', async (t) => {
   const API = new InitTransactions();
   const { fstTxHandler: handler, payerPair: payer, connection } = await API.payer();
 
-  const { mint, metadata, masterEdition } = await createAndMintDefaultAsset(
+  const { mint, metadata, masterEdition, token } = await createAndMintDefaultAsset(
     t,
     connection,
     API,
@@ -1021,6 +1021,7 @@ test('Update: Update pNFT Config', async (t) => {
     updateData,
     null,
     masterEdition,
+    token,
   );
   await updateTx.assertSuccess(t);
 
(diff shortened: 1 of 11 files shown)


token-metadata/js/test/update.test.ts
```diff
@@ -1327,7 +1329,7 @@ test('Update: Holder Authority Type Not Supported', async (t) => {
   await updateTx.assertError(t);
 });
 
-test('Update: Update pNFT Config with locked token', async (t) => {
+test('Update: Cannot Update pNFT Config with locked token', async (t) => {
```
danenbm, line 1332: Maybe this is just too much of a nitpick detail, but this test delegates and locks and then it fails because it won't update pNFT config due to delgate.

Should this or another test exist that only delegates and then fails because it is delegated?  Since that is kind of the new requirement specifically.

danenbm (dismissed):
Approved but with one comment for maybe different test

blockiosaurus (dismissed):

token-metadata/js/test/update.test.ts
samuelvanderwaal: Yeah, probably. I just modified this to fix it given the updates in v1.8.0.

febo (approved):
This PR adds a `PDAMatch` JS test for auth rules, fix the payload key type to match the expected type by the program.

It also makes the types copied from token auth rules private since they are there only to be included in the autogenerated JS code.

blockiosaurus (approved):

PR #1005, opened 2023-02-03
Add validation of authorization rules on delegate

febo:
## Context

Currently, the `Delegate` instruction does not validate authorization rules even when dealing with `pNFT` – it only validates that the correct rules set has been passed.

## Changes

The PR defines a set of operations for each delegate type and validates whether the delegate is authorized to get the specified delegate type.

The different scenarios are defined by the different delegate types:
```rust
pub enum DelegateScenario {
    Metadata(MetadataDelegateRole),
    Token(TokenDelegateRole),
}
```

The scenario is used to define the operation, which can then have a rule set associated:
```rust
let delegate_sale_operation = Operation::Delegate {
    scenario: DelegateScenario::Token(TokenDelegateRole::Sale),
};
```

Below is example of a rule set that validates that the delegate address is in an allow list:
```rust
Rule::ProgramOwnedList {
    programs: PROGRAM_ALLOW_LIST.to_vec(),
    field: PayloadKey::Delegate.to_string(),
}
```

## Considerations

* This change does not impact clients nor security (in a way it increases the security since the delegate address is now validated against the `pNFT` rule set).
* When this change is deployed, the rule sets will need to include an entry for each delegate type.

### DIFF
--- a/token-metadata/program/tests/delegate.rs
+++ b/token-metadata/program/tests/delegate.rs
@@ -327,7 +327,8 @@ mod delegate {
         // creates the auth rule set
 
         let payer = context.payer.dirty_clone();
-        let (rule_set, auth_data) = create_default_metaplex_rule_set(&mut context, payer).await;
+        let (rule_set, auth_data) =
+            create_default_metaplex_rule_set(&mut context, payer, false).await;
 
         // asset
 
@@ -472,7 +473,8 @@ mod delegate {
         // creates the auth rule set
 
         let payer = context.payer.dirty_clone();
-        let (rule_set, auth_data) = create_default_metaplex_rule_set(&mut context, payer).await;
+        let (rule_set, auth_data) =
+            create_default_metaplex_rule_set(&mut context, payer, true).await;
 
         // asset
 
@@ -510,7 +512,7 @@ mod delegate {
                 payer,
                 // delegate not authorized
                 user_pubkey,
-                DelegateArgs::SaleV1 {
+                DelegateArgs::TransferV1 {
                     amount: 1,
                     authorization_data: None,
                 },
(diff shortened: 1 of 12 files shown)


token-metadata/program/tests/delegate.rs
```diff
@@ -353,15 +356,14 @@ mod delegate {
...
-        let user = Keypair::new();
-        let user_pubkey = user.pubkey();
         let payer = Keypair::from_bytes(&context.payer.to_bytes()).unwrap();
 
         asset
             .delegate(
                 &mut context,
                 payer,
-                user_pubkey,
+                // delegate must the from Token Auth Rules or Rooster
```
danenbm, line 365: Nitpick: typo

token-metadata/program/tests/utils/programmable.rs
```diff
@@ -136,6 +163,34 @@ pub async fn create_default_metaplex_rule_set(
         .add(sale_delegate_operation.to_string(), transfer_rule.clone())
         .unwrap();
 
+    if use_delegate_allow_list {
```
danenbm, line 166: was wondering why this boolean is here

danenbm (dismissed):

austbot (approved):

token-metadata/program/tests/utils/programmable.rs
febo: The boolean is used for a test to choose to whether add a program allow list for the delegate scenarios or not.

danenbm (approved):
LGTM

token-metadata/program/tests/delegate.rs
febo: Done.

PR #1007, opened 2023-02-04
Update Token Auth Rules dependency

febo:
This PR bumps Token Auth Rules to include the non-zero byte PDA account check, which simplifies the rule sets.


blockiosaurus (approved):
This PR is updating to the new Token Auth Rules crate version and removing the PDAMatch rule from tests because it's no longer needed.

samuelvanderwaal (approved):
This PR updates the MPL token auth rules package to 1.1 and updates the test rule sets for the new program owned rule that does not require PDA match.

PR #1011, opened 2023-02-05
Relax token standard check

febo:
This PR relaxes the token standard check on transfer.


blockiosaurus (approved):
This PR is allowing tokens without a Token standard be transferred.

samuelvanderwaal (approved):

PR #1013, opened 2023-02-07
Relax token standard check

febo:
This PR relaxes the token standard check on `Delegate` and `Revoke`.


austbot (approved):

samuelvanderwaal (approved):
This PR relaxes the Token Check in `delegate` and `revoke` to allow NFTs with a Token Standard of `None` to receive delegates. Previously, it enforced that NFTs had to have  Token Standard value as we want to encourage the ecosystem to update NFTs to that standard. However, the appropriate place to do this is in the `Create` and `Update` instructions where the caller already has to have the appropriate authority to make this change.

PR #1015, opened 2023-02-10
Allow update to be performed by a delegate

febo:
This PR adds the `ProgrammableConfig` metadata delegate. The `ProgrammableConfig` is a "collection-level" delegate that can update the programmable config on the metadata account of NFTs of the same collection as long as the collection `key` of the NFT matches the collection `mint` and the update authorities of both the collection and the NFT item match.

### DIFF
--- a/token-metadata/js/test/update.test.ts
+++ b/token-metadata/js/test/update.test.ts
@@ -1174,7 +1174,7 @@ test('Update: Invalid Update Authority Fails', async (t) => {
     null,
     masterEdition,
   );
-  await updateTx.assertError(t, /Update Authority given does not match/);
+  await updateTx.assertError(t, /Invalid authority type/);
 });
 
 test('Update: Delegate Authority Type Not Supported', async (t) => {
@@ -1246,7 +1246,7 @@ test('Update: Delegate Authority Type Not Supported', async (t) => {
     daManager.masterEdition,
   );
   updateTx.then((x) =>
-    x.assertLogs(t, [/Update Authority given does not match/i], {
+    x.assertLogs(t, [/Invalid authority type/i], {
       txLabel: 'tx: Update',
     }),
   );
--- a/token-metadata/js/test/update.test.ts
+++ b/token-metadata/js/test/update.test.ts
@@ -1422,7 +1422,7 @@ test('Update: Cannot Update pNFT Config with locked token', async (t) => {
   );
 });
 
-test('Update: programmable config delegate update collection item', async (t) => {
+test('Update: rule set update with programmable config delegate', async (t) => {
   const API = new InitTransactions();
   const { fstTxHandler: handler, payerPair: payer, connection } = await API.payer();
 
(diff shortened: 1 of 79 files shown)


token-metadata/program/src/state/metadata.rs
```diff
@@ -98,104 +102,114 @@ impl Metadata {
             ..
         } = args;
 
-        if let Some(data) = data {
-            if !self.is_mutable {
-                return Err(MetadataError::DataIsImmutable.into());
-            }
+        // only allow update if the token standard is known
+        let token_standard = self
```
samuelvanderwaal, line 106: Since this is an update and the update authority or delegate is signing I think we could just automatically update their token standard for them if it's not set, but that can be a separate PR.

samuelvanderwaal (approved):

token-metadata/program/src/state/metadata.rs
febo: Done.

samuelvanderwaal (dismissed):

token-metadata/program/src/state/metadata.rs
```diff
@@ -98,104 +103,122 @@ impl Metadata {
...
             }
-            CollectionToggle::Clear => {
-                if let Some(current_collection) = self.collection.as_ref() {
-                    // Can't change a verified collection in this command.
-                    if current_collection.verified {
-                        return Err(MetadataError::CannotUpdateVerifiedCollection.into());
+
+            // if the Collection data is 'Set', only allow updating if it's unverified
+            // or if it exactly matches the existing collection info; if the Collection data
+            // is 'Clear', then only set to 'None' it if it's unverified.
```
danenbm, line 137: ultra nit: typo in this comment

danenbm (approved):
LGTM

token-metadata/js/test/update.test.ts
```diff
@@ -1422,3 +1421,109 @@ test('Update: Cannot Update pNFT Config with locked token', async (t) => {
     /Cannot update the rule set of a programmable asset that has a delegate/i,
   );
 });
+
+test('Update: programmable config delegate update collection item', async (t) => {
```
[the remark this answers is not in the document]
febo: This is a good point, I will add this. The only caveat is whether we should fail the transaction if you try to update a field you are not allowed or the update should succeed but don't change anything.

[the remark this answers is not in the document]
danenbm: To me it seems better to fail, otherwise maybe someone tries to use the wrong delegate and thinks they were successful but there was a silent failure.  Although most people probably should and would choose to inspect the metadata after doing something to see if it really updated.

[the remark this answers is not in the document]
febo: Done – now there is a validation of the update args.

token-metadata/program/src/state/metadata.rs
febo: Done.

danenbm (approved):

samuelvanderwaal (approved):

PR #1017, opened 2023-02-11
Refactor lock and unlock logic

febo:
The current logic for lock/unlock of fungibles gives the impression that a delegate can be used to freeze/thaw a token account holding fungibles assets. This PR refactors the code to clearly separate the 3 different authorities that can lock/unlock depending on the token standard:
1. `token delegate` for pNFTs
2. `spl-token delegate` for non-fungibles
3. `freeze authority` for fungibles

### DIFF
--- a/token-metadata/js/idl/mpl_token_metadata.json
+++ b/token-metadata/js/idl/mpl_token_metadata.json
@@ -2829,7 +2829,7 @@
           "name": "authority",
           "isMut": false,
           "isSigner": true,
-          "desc": "Delegate account"
+          "desc": "Delegate or freeze authority"
         },
         {
           "name": "tokenOwner",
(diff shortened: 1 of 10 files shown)


samuelvanderwaal (approved):
This PR refactors the lock/unlock logic to clearly distinguish between three different authority types:

* TokenDelegate for `ProgrammableNonFungible` type
* SPL Delegate for `NonFungible` and `NonFungibleEdition` types
* Freeze Authority for `Fungible` and `FungibleAsset` types

Each of these three authority types can "lock" the token using the appropriate method:

* TokenDelegate sets the actual TokenState
* SPL delegate is a pass through to freeze and thawing via SPL Token
* Freeze Authority is also a pass through to SPL token but directly invokes it from the handler instead of using our `thaw` and `freeze` convenience functions since those assert edition derivations are only for nonfungibles

danenbm (dismissed):

samuelvanderwaal (approved):

PR #1020, opened 2023-02-13
Add unified `Verify` and `Unverify` handlers to support pNFT and existing types

danenbm:
#### Verify
Add unified `Verify` handler to support pNFT and existing types 
* Adding instruction builder to token-metadata for `Verify`.
* Add version to `Verify`.
* Port creator sign.
* Port sized collection verification.
* Port support for unsized collection.
* Replace creator loop with iterator and filter.
* Remove payer from `Verify` since no account creation/realloc.
* Split creator vs. collection handling based on `VerifyArgs`.
* Updating collection verify to use new delegate.
  * Change which accounts are needed for handler.
  * Update account ownership/signature verifcations accordingly.
* Moved verification out of `collection` and to its own files under `verification`.
* Added tests for both NFTs and pNFTs.

#### Unverify
Adding `Unverify` code, ported from existing handlers, including unverifying
with burned collection parent.
* Created instruction builder and handler for `Unverify`.
* Currently returns `FeatureNotSupported` since the unverify tests not
in place.
* Added test cases for creator and collection unverify of NFTs and pNFTs.
* Added test cases for collection unverify of burned collection parent NFTs and pNFTs.


token-metadata/program/src/instruction/metadata.rs
```diff
@@ -759,3 +759,75 @@ impl InstructionBuilder for super::builders::Update {
...
+                AccountMeta::new_readonly(creator, true)
+            } else {
+                AccountMeta::new_readonly(crate::ID, false)
+            },
+            if let Some(collection_authority) = self.collection_authority {
+                AccountMeta::new_readonly(collection_authority, true)
+            } else {
+                AccountMeta::new_readonly(crate::ID, false)
+            },
+            if let Some(collection_mint) = self.collection_mint {
```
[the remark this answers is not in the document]
danenbm: yes will do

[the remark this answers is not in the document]
danenbm: Done

token-metadata/js/idl/mpl_token_metadata.json
```diff
@@ -3449,37 +3449,120 @@
     {
       "name": "Verify",
       "accounts": [
+        {
+          "name": "authority",
```
austbot, line 3453: maybe we call this target or subject

austbot: the reasoning is that its a more general term

token-metadata/program/tests/utils/digital_asset.rs
```diff
@@ -64,26 +65,146 @@ impl DigitalAsset {
...
+        collection_mint: Option<Pubkey>,
+        collection_metadata: Option<Pubkey>,
+    ) -> Result<(), BanksClientError> {
+        let mut builder = UnverifyBuilder::new();
+        builder
+            .authority(authority.pubkey())
+            .metadata(metadata.unwrap_or(self.metadata));
+
+        match args {
+            VerifyArgs::CreatorV1 => (),
```
austbot, line 135: this is possible right? they must unverify before they can be removed.

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+    utils::{clean_write_metadata, decrement_collection_size, increment_collection_size},
+};
+use mpl_utils::assert_signer;
+use solana_program::{entrypoint::ProgramResult, pubkey::Pubkey};
+
+pub(crate) fn verify_collection_v1(program_id: &Pubkey, ctx: Context<Verify>) -> ProgramResult {
+    // Assert program ownership/signers.
+
+    // Authority account is the collection authority and must be a signer.
+    assert_signer(ctx.accounts.authority_info)?;
```
austbot, line 18: the comment doesnt match this statement this only checks signer

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    if let Some(delegate_record_info) = ctx.accounts.delegate_record_info {
+        assert_owned_by(delegate_record_info, program_id)?;
```
austbot, line 21: im pretty sure this check is already happening in getAuthorityType

austbot (dismissed):
I have either missed it or its not here but where are you ensuring that the update authority of the NFT to be verified matches the current authority as the collection NFT?

token-metadata/program/src/instruction/metadata.rs
```diff
@@ -759,3 +759,87 @@ impl InstructionBuilder for super::builders::Update {
         }
     }
 }
+
+#[repr(C)]
```
samuelvanderwaal, line 763: Since verification has its own section under processor it should have it's own section in instruction IMO.

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+            // In the case of a sized collection, update the size on the collection parent.
+            if collection_metadata.collection_details.is_some() {
+                increment_collection_size(&mut collection_metadata, collection_metadata_info)?;
+            }
+        }
+        None => return Err(MetadataError::CollectionNotFound.into()),
+    };
+
+    // Reserialize metadata.
+    clean_write_metadata(&mut metadata, ctx.accounts.metadata_info)?;
```
samuelvanderwaal, line 100: Once stylistic suggestion here is you can elide the `Ok(())` and return the result from the `clean_write_metadata` call directly:

```rust
     . . .
    clean_write_metadata(&mut metadata, ctx.accounts.metadata_info)
}
```

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+    let collection_mint_info = ctx
+        .accounts
+        .collection_mint_info
+        .ok_or(MetadataError::MissingCollectionMint)?;
+    assert_owned_by(collection_mint_info, &spl_token::ID)?;
+
+    let collection_metadata_info = ctx
+        .accounts
+        .collection_metadata_info
+        .ok_or(MetadataError::MissingCollectionMetadata)?;
```
samuelvanderwaal, line 130: Missing owner check for `collection_metadata`.

samuelvanderwaal (commented):
This is good, clean code. I did a first pass and had some minor suggestions. I'll do a more thorough review after you finish all the tests.

token-metadata/program/src/instruction/metadata.rs
febo: ^ Agreed

token-metadata/program/tests/verify.rs
```diff
@@ -0,0 +1,1929 @@
...
+    transaction::TransactionError,
+};
+use utils::*;
+
+// Note: at the time these tests were created, the only Metadata delegates that have been
+// implemented are `Collection`, `Update`, and `ProgrammableConfig`.  We have tested each of these
+// cases.
+
+// Also at this time, a collection parent NFT cannot have a token standard of
+// `ProgrammableNonFungible`.  This means that using the new delegate handler, the only Token
```
febo, line 27: Where is this restriction enforced?

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+    // size data.
+    if let Some(collection) = &metadata.collection {
+        if collection.verified {
+            return Err(MetadataError::AlreadyVerified.into());
+        }
+    }
+
+    // Verify the collection in the item's metadata matches the collection mint.  Also verify
+    // the collection metadata matches the collection mint, and the collection edition derivation.
+    assert_collection_verify_is_valid(
```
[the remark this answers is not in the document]
samuelvanderwaal: I think there's a larger refactor you can do here to only do the match on the metadata.collection once. [I made a PR to your branch for it](https://github.com/metaplex-foundation/metaplex-program-library/pull/1042/files) to make it easier to see the diff. If you like it, you can just merge it into your branch, otherwise you can close it after you've taken a look.

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+        return Ok(());
+    }
+
+    // The collection parent must be the actual parent of the collection item.
+    if collection.key != *collection_mint_info.key {
+        return Err(MetadataError::NotAMemberOfCollection.into());
+    }
+
+    // Ensure the metadata is derived from the mint.
+    assert_metadata_derivation(program_id, collection_metadata_info, collection_mint_info)?;
```
samuelvanderwaal, line 148: I wrote this original code and function, and I like having clean assertion functions like this that do one thing, but thinking about it more, a much cheaper check here than doing the derivation is:

```rust
if metadata.mint != mint_info.key {
    return Err(MetadataError::MintMismatch.into());
}
```

We have already validated that the Metadata account is legitimate and belongs to Token Metadata, so we know that the mint field on it is the mint it was derived from and so we can just do a pubkey comparison instead of a derivation check.

If you like the function assertion approach this could be:

```rust
assert_metadata_matches_mint(&collection_metadata, collection_mint_info)?;
```

Where `assertions/metadata.rs` has it defined:

```rust
pub fn assert_metadata_matches_mint(metadata: &Metadata, mint_info: &AccountInfo) -> ProgramResult {
    if metadata.mint != *mint_info.key {
        return Err(MetadataError::MintMismatch.into());
    }
    Ok(())
}
```

token-metadata/program/src/processor/verification/collection_verification.rs
samuelvanderwaal, line 125: Note to myself and everyone else: owner check for this happens later after we ensure it's not a burned collection parent, otherwise the check might fail if the burned NFT has been reassigned to the system program.

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+        assert_owned_by(collection_metadata_info, program_id)?;
+
+        // Now we can deserialize the collection metadata account.
+        let mut collection_metadata = Metadata::from_account_info(collection_metadata_info)?;
+
+        // Determines if we have a valid authority to perform the collection verification.  The
+        // required authority is either the collection parent's metadata update authority, or a
+        // collection delegate for the collection parent.  This call fails if no valid authority is
+        // present.
+        let authority_response = AuthorityType::get_authority_type(AuthorityRequest {
```
samuelvanderwaal, line 185: [I think there's a refactor here that reduces code redundancy](https://github.com/metaplex-foundation/metaplex-program-library/pull/1043), but you may not prefer it or see something better so feel free to not use the PR.

samuelvanderwaal (dismissed):
Overall the code is good and well organized. I have a number of refactor suggestions and one major requested change: dropping the unused master_edition account. 

I'll review tests next.

token-metadata/program/tests/verify.rs
```diff
@@ -0,0 +1,1928 @@
...
+            .await;
+    }
+}
+
+mod verify_collection {
+    use super::*;
+
+    #[tokio::test]
+    async fn delegate_record_wrong_owner() {
+        // See `standard_delegate_fails_collection_created_new_handlers`.
```
samuelvanderwaal, line 224: Is `standard_delegate_fails_collection_created_new_handlers` an unwritten test? I don't see it anywhere.

danenbm: Yeah I renamed it to the shorter `standard_delegate_fails`

token-metadata/program/src/processor/verification/collection_verification.rs
danenbm: I can add a comment here saying owner check happens later

token-metadata/program/src/error.rs
```diff
@@ -687,16 +685,30 @@ pub enum MetadataError {
...
     #[error("Invalid InstructionsSysvar")]
     InvalidInstructionsSysvar,
 
-    /// 175
+    /// 176
     #[error("Authority cannot apply all update args")]
     InvalidUpdateArgs,
+
+    /// 177
+    //#[error("Missing delegate record PDA account")]
```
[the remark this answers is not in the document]
danenbm: yeah I 'm going to merge `feat/burn` into this branch now

token-metadata/program/tests/verify.rs
danenbm: https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/assertions/collection.rs#L139

I think its just outdated and will fix as part of this PR.

danenbm:
> I have either missed it or its not here but where are you ensuring that the update authority of the NFT to be verified matches the current authority as the collection NFT?

We are not doing that in this handler because this new `Verify` handler will only do verification functionality.  So the only person allowed to do verification is the collection parent's update authority or a collection delegate - both are just authority for the collection parent.

If the member NFT needs the collection set in its metadata, that will need to be done by an update metadata process, and that will need to be the NFT's metadata update authority or a delegate. 

Now there is one caveat, which is in the case of unverifying a collection with a burned parent NFT.  In this case we require the metadata's update authority or delegate to do the unverify.

We have a future task to allow for "set and verify" functionality to occur in the new `Update` handler if and only if the condition you state above is true (update authorities match between NFT and collection NFT).

token-metadata/js/idl/mpl_token_metadata.json
danenbm: Yes I agree with this but I think we should leave it as `authority` for now only because all the other new handlers are calling it `authority`.

token-metadata/program/src/error.rs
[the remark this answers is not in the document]
danenbm: Merged in burn features from `master` branch.

token-metadata/program/tests/verify.rs
danenbm: fixed

token-metadata/program/src/processor/verification/collection_verification.rs
danenbm: Added clarifying comment.

token-metadata/program/src/processor/verification/collection_verification.rs
danenbm: Fixed

token-metadata/program/tests/utils/digital_asset.rs
danenbm: After https://github.com/metaplex-foundation/metaplex-program-library/commit/ae882ee7fe85e8e3ba566f93f2968df7ea39eee9 we don't let them remove a verified creator either by setting creators array to `None` or setting creators array to `Some(Vec<Creator>)`.  So they need to unverify first before removal.

I'm not 100% sure what you mean by this comment though, because this unverify function is just building the instruction, and in the case of unverifying creators, it is specified using `VerifyArgs::CreatorV1` and then there's nothing else in the instruction to add (no delegate record, collection mint, collection metadata needed when unverifying a creator).

One thing I'm doing in the new version is renaming `VerifyArgs` to `VerificationArgs` to make it clear that the same args struct is used by both `Verify` and `Unverify`.

token-metadata/program/src/processor/verification/collection_verification.rs
danenbm: yes you are right it is redundant

danenbm: Removed and replaced with a comment indicating it.

token-metadata/program/src/processor/verification/collection_verification.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    // Deserialize item metadata and collection parent metadata.
+    let mut metadata = Metadata::from_account_info(ctx.accounts.metadata_info)?;
+    let mut collection_metadata = Metadata::from_account_info(collection_metadata_info)?;
+
+    // Don't verify already verified items, otherwise for sized collections we end up with invalid
+    // size data.
+    if let Some(collection) = &metadata.collection {
+        if collection.verified {
+            return Err(MetadataError::AlreadyVerified.into());
```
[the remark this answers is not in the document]
danenbm: Yeah it makes sense to me to let it pass if already verified for this new handler.

token-metadata/program/src/instruction/mod.rs
```diff
@@ -726,20 +726,34 @@ pub enum MetadataInstruction {
...
-    #[account(1, signer, writable, name="collection_authority", desc="Collection Update authority")]
-    #[account(2, signer, writable, name="payer", desc="payer")]
-    #[account(3, optional, name="authorization_rules", desc="Token Authorization Rules account")]
-    #[account(4, optional, name="authorization_rules_program", desc="Token Authorization Rules Program")]
+    #[account(0, signer, name="authority", desc="Creator to verify, collection update authority or delegate")]
+    #[account(1, optional, name="delegate_record", desc="Delegate record PDA")]
+    #[account(2, writable, name="metadata", desc="Metadata account")]
+    #[account(3, optional, name="collection_mint", desc="Mint of the Collection")]
+    #[account(4, optional, writable, name="collection_metadata", desc="Metadata Account of the Collection")]
+    #[account(5, optional, name="collection_master_edition", desc="Master Edition Account of the Collection Token")]
```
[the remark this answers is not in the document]
danenbm: Pending other comment resolution

token-metadata/program/src/processor/verification/collection_verification.rs
danenbm: Merged it but currently getting an issue with: error[E0597]: `collection_metadata.update_authority` does not live long enough`.  Also I need to specify different `metadata_delegate_roles` for the two cases.  Let me take a look at it.

danenbm: Ok fixed the issues and incorporated this.

token-metadata/program/src/processor/verification/collection_verification.rs
danenbm: > We have already validated that the Metadata account is legitimate and belongs to Token Metadata
Not the collection metadata though.  We only do that in the branch when parent is not burned, i.e. data is not empty.

If the collection parent was burned, then the owner of `collection_metadata` will be System Program and its data will be empty.  So we cannot check that `collection_metadata.mint == collection_mint_info.key`.

token-metadata/program/src/instruction/metadata.rs
danenbm: Done

token-metadata/program/src/instruction/metadata.rs
```diff
@@ -759,3 +759,87 @@ impl InstructionBuilder for super::builders::Update {
...
+#[cfg_attr(feature = "serde-feature", derive(Serialize, Deserialize))]
+#[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Debug, Clone)]
+pub enum VerifyArgs {
```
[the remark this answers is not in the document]
danenbm: Done

token-metadata/program/tests/verify.rs
danenbm: Fixed `assert_master_edition` but still going to add some tests.

token-metadata/program/src/processor/verification/collection_verification.rs
samuelvanderwaal: Ah you're right, probably best to leave that `assert_metadata_matches_mint` in place then.

token-metadata/program/src/processor/verification/collection_verification.rs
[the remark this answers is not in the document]
samuelvanderwaal: I missed that `assert_master_edition` is doing the supply check. Yeah, we want to avoid weirdness by ensuring that Collection Parents are not print editions and are master editions that cannot print so I think you're right that this account needs to stay.

token-metadata/program/tests/verify.rs
danenbm: Added tests for NFT/pNFT collections

And a test that makes sure in the case of pNFT, that Utility Delegate cannot verify or unverify.

febo (approved):
This PR adds the `Verify` instruction to verify a collection item or creator. Verification can be authorized by the update authority or collection delegate in the case of collection; for creators, only the creator can verify themselves.

blockiosaurus (approved):

token-metadata/js/idl/mpl_token_metadata.json
febo: Sounds sensible.

token-metadata/program/tests/utils/digital_asset.rs
febo: Looks good.

PR #1022, opened 2023-02-14
Adding minor size optimizations

blockiosaurus:
Adding size reductions to token metadata in the form of abbreviated debug messages or removals where unnecessary.

### DIFF
--- a/token-metadata/program/src/processor/mod.rs
+++ b/token-metadata/program/src/processor/mod.rs
@@ -93,47 +93,47 @@ pub fn process_instruction<'a>(
     // match on the new instruction set
     match instruction {
         MetadataInstruction::Burn(args) => {
-            msg!("Instruction: Token Metadata Burn");
+            msg!("IX: Burn");
             burn::burn(program_id, accounts, args)
         }
         MetadataInstruction::Create(args) => {
-            msg!("Instruction: Token Metadata Create");
+            msg!("IX: Create");
             metadata::create(program_id, accounts, args)
         }
         MetadataInstruction::Mint(args) => {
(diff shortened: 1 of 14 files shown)


samuelvanderwaal (approved):

samuelvanderwaal (approved):

token-metadata/program/src/processor/mod.rs
```diff
@@ -201,39 +201,39 @@ fn process_legacy_instruction<'a>(
             )
         }
         MetadataInstruction::DeprecatedCreateMasterEdition(_args) => {
-            msg!("Instruction: Deprecated Create Master Edition, Removed in 1.1.0");
+            msg!("IX: Deprecated Create Master Edition, Removed in 1.1.0");
             Err(MetadataError::Removed.into())
         }
         MetadataInstruction::DeprecatedMintNewEditionFromMasterEditionViaPrintingToken => {
-            msg!("Instruction: Deprecated Mint New Edition from Master Edition Via Token, Removed in 1.1.0");
+            msg!("IX: Deprecated Mint New Edition from Master Edition Via Token, Removed in 1.1.0");
```
febo, line 208: For the ones that have been removed, perhaps we can remove the whole message – not necessarily now.

blockiosaurus: Yeah that's a good call. I didn't notice how verbose those messages are until today.

febo (approved):
This PR removes and shorten debug messages to reduce the binary size.

token-metadata/program/src/processor/mod.rs
samuelvanderwaal: Good point

PR #1024, opened 2023-02-14
Improve nft-packs error enum to show better code error

Peixer:
### Description
This change is intended to save newbies a lot of time. NFT-packs is returning an error code which is easily confused with other native errors like this example `failed to send transaction: Transaction simulation failed: Error processing Instruction 0: custom program error: 0x1` for those not used with this error message can easily mixed up with `0x1 = InsufficientFunds`(the first thing you will find on google).

### Solution
I made a change in the error enum to start from 6000, the same code error used in the other programs.

### Testing
Call the `createInitPackInstruction` function with `redeemStartDate` parameter below today, you must get 0x1 code error.


samuelvanderwaal:
I like this this change but need to think about if this could break anyone. In the mean time, can you run `cargo fmt` to fix the failing CI?

Peixer:
@samuelvanderwaal I did it 👍 Let me know how I can be of more help. I hope this change can reach and help more people 🤩

PR #1026, opened 2023-02-15
remove redundant token standard check

samuelvanderwaal:
The token standard checks are too strict since there are many legacy `NonFungible` assets without a Token Standard set. The validation of the token standard is done in `mpl-migration-validator` [here](https://github.com/metaplex-foundation/mpl-migration-validator/blob/e8dd04252b3a24548c80f6e285209a62d95c43d4/program/src/processor/migrate/validate.rs#L146), so can be removed from this handler.


danenbm (approved):
This PR removes the token standard check from the token-metadata migrate handler because it blocks legacy NFTs that don't have a token standard set.

We could instead add a check such that if the standard is not present, it checks that it's a master edition with supply of 1 and decimals of 0.

This is in fact what `mpl-migration-validator` does.  We would rather have the whole check there because 1) uses less token-metadata binary size and 2) easier to change `mpl-migration-validator` and that check is more in line with the purpose of `mpl-migration-validator`.

febo (approved):
This PR removes a redundant check – a more comprehensive check is already done in `mpl-migration-validator`.

PR #1027, opened 2023-02-17
Removing removed ix warnings and error messages

blockiosaurus:


samuelvanderwaal:
Some of the JS tests rely on specific log or error messages so you need to update those to fix CI. It looks like only two are failing right now so should be an easy fix.

samuelvanderwaal:
> Some of the JS tests rely on specific log or error messages so you need to update those to fix CI. It looks like only two are failing right now so should be an easy fix.

Actually, this is a bigger issue than I thought because it completely erases the JS error messages. I think it would probably require a Solita change to use the name as the message instead of the predefined string. It might be worth chatting with @lorisleiva about the best way to go about this since he's switching Token Metadata to use Kinobi soon.

PR #1028, opened 2023-02-18
Add revoke for migration delegate

febo:
This PR adds the ability to revoke a `TokenDelegateRole::Migration` delegate.

### DIFF
--- a/token-metadata/program/src/state/programmable.rs
+++ b/token-metadata/program/src/state/programmable.rs
@@ -131,7 +131,7 @@ impl TokenRecord {
 
 impl Resizable for TokenRecord {
     fn from_bytes<'a>(account_data: &[u8]) -> Result<TokenRecord, ProgramError> {
-        // we perform a manual deserializagtion since we are potentially dealing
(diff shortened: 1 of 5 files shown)


samuelvanderwaal (dismissed):
This allows revoking of the one-time `MigrationDelegate` which is given to all pNFTs created by migration. Without this, certain contracts would have to perform annoying work-arounds by transferring internally. `MigrationDelegate`s shouldn't be able to be created via `delegate` but they do need to be revokable.

token-metadata/program/src/state/programmable.rs
```diff
@@ -131,7 +131,7 @@ impl TokenRecord {
 
 impl Resizable for TokenRecord {
     fn from_bytes<'a>(account_data: &[u8]) -> Result<TokenRecord, ProgramError> {
-        // we perform a manual deserializagtion since we are potentially dealing
```
danenbm, line 134: maybe deserializagtion uses less compute

danenbm (approved):

samuelvanderwaal (approved):

PR #1037, opened 2023-02-24
Rework the metadata args into a wrapper around actual Metadata

ngundotra:


ngundotra:
If this looks good, I'll refactor the `process_collection_verification_mpl_only` and `process_creator_verification`

Issue #922, opened 2022-12-22
Not able to run Auctioneer Execute Sale instruction on purchased NFT[Bug]: 


samia-invozone:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

Following are the activity details on this NFT, when the highest bidder is trying to execute the sale then I'm facing an error. Last & highest bidder is trying to execute sale but it throws error.
<img width="781" alt="image" src="https://user-images.githubusercontent.com/94119384/209084094-fb1ccad2-356c-4ecc-846f-d1aeddbef6dc.png">

Following are all transactions of this NFT. You guys can check accounts.
[Mint](https://solscan.io/token/9j4QUBnxGtjHKnWpUPzseuB1SLdUm9FunzUBC5fTWVr4?cluster=devnet#holders)

[List](https://solscan.io/tx/2Wr2AGHzKwWajKQBTen1CKUMFqtJyFhaAimTezxgoBboAesjzeqEr1m2YZTbygspCRuvowCsKBc58fLuqGao5MLF?cluster=devnet)

[Bid 1](https://solscan.io/tx/m7MsS98hwHTTRF4XKEjvgiNC2sHeJw93iZkf8nAuX2xYtf3WKC8QfLS5WKVHSBhRQtf3MvgDqp9tJRsA5mCQ5Yh?cluster=devnet)

[Bid 2](https://solscan.io/tx/3JRZtxk9p5UgjHHf9TYU9yEasoZhyje94Lcd5poRPWrLug9whFn8duvJxuwqHeqzeF1pS9TVhVNN8nrndWccrLLL?cluster=devnet)

[Execute Sale](https://solscan.io/tx/5nx4114SqEWNACUEHMBP1K8D8ueWgBRcZ7nZ9Z4Q9UZTJxDsoTHWvYvUTKRuawnd4oA6jygoZob2ooSE6uhLeyWX?cluster=devnet)

[List from new owner](https://solscan.io/tx/33QYbtnhogJC76pEyLFo1Nzi64huqHb3DNZW7zxn1AmRrRkcqkikyxUUS8wkNXvCsmmrRwCtHhBDhthpZXvvSt9u?cluster=devnet)

[...]

Issue #852, opened 2022-10-26
Execute_Sale(remaining_accounts) issue on NFT that is minted through CM/Sugar


samia-invozone:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

Hi, I'm trying to run execute instruction using Auctioneer with NFT that has been minted through candy machine or sugar etc;

But actually, it looks like the auction house smart contract seems not to be compatible with these types of collections because creator on index 0 is candy machine address and when I try to pass these addresses into remaining accounts it returns an error of "An account required by the instruction is missing”.  Here is the NFT for which i'm trying to run execute rpc call with remaining accounts. https://solscan.io/token/78sJUf27U45coCt1HzPMAdQKRLWWee46CfRTaRRZgASk?cluster=devnet#metadata

Following are the successful transactions for listing & bidding:

https://solscan.io/tx/5xkp1DFpKqLq2fkLM9MYFCLQdqDf9TMEhrXkuyGkzH84Fyjf92XMztfyi9VDpxJEj4ZDxPvgMPqKSCiuJhvtjvYj?cluster=devnet

https://solscan.io/tx/2niP4SNtB5xWSkaMoBaqi7e2y1i8P2kkq1DaBHh6gpMBdwk84FkAuSWRVD8h3Vgfez8Chtz9P87fVnomnYUQo8Um?cluster=devnet

Using following code for sending remaining accounts:

` .remainingAccounts([
        {
          pubkey: new PublicKey("5k2EPeEp4ZZaRKdW7uDvTxwoRgYkatk15LEhRPYXhaxD"),
          isWritable: false,
          isSigner: false,
        },
        {
          pubkey: new PublicKey("4douvideaZJtZLztHJK8H5FDrr1AkVYZUWGDgyYxCwL8"),
          isWritable: true,
          isSigner: false,
        },
      ])`

### Relevant log output

```shell
logs
: 
Array(5)
0
: 
"Program neer8g6yJq2mQM6KbnViEDAD4gr3gRZyMMf4F2p3MEh invoke [1]"
1
: 
"Program log: Instruction: ExecuteSale"
2
: 
"Instruction references an unknown account 5k2EPeEp4ZZaRKdW7uDvTxwoRgYkatk15LEhRPYXhaxD"
3
: 
[...]

Issue #786, opened 2022-10-03
[Bug]: New auctioneer js sdk execute_sale function still asks for remaining accounts even after pasiing them?


Maverick9081:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

JS SDK

### Issue description

I'm passing these accounts 
![image](https://user-images.githubusercontent.com/57398733/193570892-39de6cf7-896c-4e96-9535-196c47e85100.png)
I tried to pass the remaining accounts but the same error, Tried to add remaining accounts in instruction.keys still not working.

### Relevant log output

```shell
logs: [
    'Program neer8g6yJq2mQM6KbnViEDAD4gr3gRZyMMf4F2p3MEh invoke [1]',
    'Program log: Instruction: ExecuteSale',
    'Program hausS13jsjafwWwGqZTUQRmWyvyxn9EQpqMwV1PBBmk invoke [2]',
    'Program log: Instruction: AuctioneerExecuteSale',
    'Program log: ProgramError occurred. Error Code: NotEnoughAccountKeys. Error Number: 47244640256. Error Message: The instruction expected additional account keys.',
    'Program hausS13jsjafwWwGqZTUQRmWyvyxn9EQpqMwV1PBBmk consumed 54166 of 135807 compute units',
    'Program hausS13jsjafwWwGqZTUQRmWyvyxn9EQpqMwV1PBBmk failed: insufficient account keys for instruction',
    'Program neer8g6yJq2mQM6KbnViEDAD4gr3gRZyMMf4F2p3MEh consumed 118359 of 200000 compute units',
    'Program neer8g6yJq2mQM6KbnViEDAD4gr3gRZyMMf4F2p3MEh failed: insufficient account keys for instruction'
  ]
}
```

### Priority this issue should have

Medium (should be fixed soon)

PR #1050, opened 2023-03-13
Pass `remaining_accounts` to auction house properly

KartikSoneji:
Fixes errors when the creators of the nft are not the buyer or seller of a timed auction nft.

Closes:
- #922
- #852
- #786


omarkilani:
@KartikSoneji feels like this could use a new test or a modification of one of the tests to trigger the issue in the pre-PR code?

KartikSoneji:
Fair point, I'll add a test.

PR #1051, opened 2023-03-14
fix: Add Creator Minimum Balance Check

KartikSoneji:
If one of the creators of an NFT has less than the rent exempt amount in their wallet, and the royalty from the sale is not enough to bring the balance over the rent exempt minimum then the sale will fail.

A workaround is to skip paying creator royalties unless they result in a valid balance for the creator wallet.

### DIFF
--- a/auction-house/program/src/utils.rs
+++ b/auction-house/program/src/utils.rs
@@ -399,10 +399,24 @@ pub fn pay_creator_fees<'a>(
                         .ok_or(AuctionHouseError::NumericalOverflow)?
                         .checked_div(100)
                         .ok_or(AuctionHouseError::NumericalOverflow)? as u64;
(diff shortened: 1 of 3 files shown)


samuelvanderwaal:
@KartikSoneji does this only apply to creators that are PDA addresses?

KartikSoneji:
This also applies to normal keypairs, since the creator can empty it out.

samuelvanderwaal (approved):

samuelvanderwaal (approved):

KartikSoneji:
Happy to help :)
Is there a rough timeline for when the on-chain program will be updated?

samuelvanderwaal:
> Happy to help :) Is there a rough timeline for when the on-chain program will be updated?

I plan to deploy it this week. There are a couple of other issues raised I want to resolve first.

PR #1054, opened 2023-03-22
pr

staccDOTsol:
how does code owner work

   // it is they fault monye lost forever. let em have it. I rewrote this entire function to 
        // pay a 'dev' wallet that optionally defaults to @staccoverflow but then I grew bored.
        //return Err(HydraError::SharesArentAtMax.into());

### DIFF
--- a/hydra/program/src/utils/validation/mod.rs
+++ b/hydra/program/src/utils/validation/mod.rs
@@ -64,7 +64,9 @@ pub fn assert_ata(
 
 pub fn assert_shares_distributed(fanout: &Account<Fanout>) -> Result<()> {
     if fanout.total_available_shares != 0 {
-        return Err(HydraError::SharesArentAtMax.into());
+        // it is they fault monye lost forever. let em have it. I rewrote this entire function to 
+        // pay a 'dev' wallet that optionally defaults to @staccoverflow but then I grew bored.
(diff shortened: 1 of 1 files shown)


blockiosaurus:
I think ideally we'd error in this case so users can't have floating shares.

staccDOTsol:
cheers, 

my logic: there's a gap in some sales that aren't hyped before they sell out.

The present method to alleviate this is to allocate 100% of the shares to the first mint, then on 2nd mint reallocate to 50% 50%. This is cost prohibitive :)

Would it be possible for us to allocate the hydra itself as a 'default' shareholder?

blockiosaurus:
Could that be done with this PR? https://github.com/metaplex-foundation/metaplex-program-library/pull/884
I don't know what I'm doing but someone who did could take that over the finish line.

samuelvanderwaal:
I don't understand this PR as is, @staccDOTsol can you elaborate what the goal is here?

staccDOTsol:
At present, the flow is that every share needs to be assigned a member before a hydra can start paying out. This isn't conducive in all cases ie. slower mints

blockiosaurus:
You can modify the shares after the fact, so it should be possible to just do that instead of leaving unassigned shares that go to the void.

PR #1058, opened 2023-03-30
Changes to new `Update` instruction

danenbm:
### New change
* Binary space savings changes
  * Remove InternalUpdateArgs struct and `From` trait implementation.
  * Deconstruct `UpdateArgs` where used.
  * Remove redundant authority type and metadata delegate role checks from `Metadata::update_v1` because `validate_update` already checks that the authority is allowed to perform an update, and the `UpdateArgs` enforces that the authority is only changing fields for which it is authorized.

### Notes
* Add Additional Metadata Delegates
  * Add support in `Delegate` and `Revoke` for `AuthorityItem`, `Data`, `DataItem`, `CollectionItem`, and `ProgrammableConfigItem` Metadata delegates.
  * `Authority` Metadata delegate is replaced by `AuthorityItem`, and `Update` Metadata delegate is replaced by `Data`.  This is OK because @samuelvanderwaal ran a script and saw no current usage of those delegates (those delegates cannot currently do anything anyways).
* Changes to `Update` to support new and changed delegates
  * Modify authority check to separate out item and collection-level delegates.
  * Add several new `V2` `UpdateArgs` variants for update authority and each metadata delegate type, which makes it clear to users what metadata fields can be updated by each authority type.   It also makes it trivial to enforce this in the handler.
  * Modify `Update` handler to update metadata areas based on update authority or delegated authority.

### Cargo lock change
During test development Cargo.lock file was updated, had to run `cargo update -p mpl-token-metadata@1.10.0` to get the compiler to select a newer version of `mpl-utils` and succeed overall.

### Docs
Here is the related docs PR: https://github.com/metaplex-foundation/docs/pull/425

### Enum Compatibility
* For `UpdateArgs` I added all new variants to the end of the enum.
* For `MetadataDelegateRole` enum I changed the name of `Authority` to `AuthorityItem`, `Update` to `Data`, and added the new delegates to the end of the enum.  This is OK because @samuelvanderwaal ran a script and saw no current usage of those delegates (those delegates cannot currently do anything anyways).
* For `DelegateArgs` and `RevokeArgs` enums I changed the name of `UpdateV1` to `DataV1`, and added new delegates to the end of the enum.

### Delegate expiration
* This is an idea that was explored as part of this PR, but decided to keep this PR focused on new delegates and the `Update` instruction.
* Current idea for delegate expiration can be seen in: https://github.com/metaplex-foundation/metaplex-program-library/pull/1063

### Tests
  * Change `unverify` test on collection delegate behavior change
  * Added tests for:
    * new delegates passing and failing to update
    * persistent delegate failing to update
    * updating token standard
    * collection-level Collection delegate can update collection for items in the collection.
    * collection-level Collection Item delegate cannot update items in the collection.
    * collection-level Programmable Config delegate cannot update collection for items in the collection by using wrong `UpdateArgs`.
    * collection-level Programmable Config delegate can update programmable config for items in the collection, using both the V2 `UpdateArgs` meant for Programmable Config delegate, as well as the V1 `UpdateArgs`.
      * This was to maintain backwards compatibility since Programmable Config delegate was in place before this PR.
    * collection-level Programmable Config delegate cannot update collection for items in the collection by using `collection` field in V1 `UpdateArgs`.
    * collection cannot be updated to verified.
    * old version of lib can be used to delegate and update with new deployed program (even though `DelegateArgs` enum was modified).
 
### Unrelated change
* Fix delegate test based on auth rules update.


samuelvanderwaal (changes requested):

token-metadata/program/src/processor/metadata/update.rs
```diff
@@ -161,6 +179,44 @@ fn update_v1(program_id: &Pubkey, ctx: Context<Update>, args: UpdateArgs) -> Pro
...
+        metadata_delegate_role,
+        desired_token_standard,
+    )?;
+
+    // Find existing token standard from metadata or infer it.
+    let existing_or_inferred_token_standard = if let Some(token_standard) = metadata.token_standard
+    {
+        token_standard
+    } else {
+        // TODO: What if they have an edition account but choose not to pass it in?
```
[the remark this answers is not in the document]
danenbm: Removed

token-metadata/program/src/instruction/delegate.rs
```diff
@@ -29,7 +30,7 @@ pub enum DelegateArgs {
         /// Required authorization data to validate the request.
         authorization_data: Option<AuthorizationData>,
     },
-    UpdateV1 {
+    DataV1 {
```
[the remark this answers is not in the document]
danenbm: @samuelvanderwaal ran a script and found no Authority or Update delegates in place so we will change `Authority` to `AuthorityItem` and `Update` to `Data`.

token-metadata/program/src/processor/metadata/update.rs
```diff
@@ -55,6 +57,7 @@ fn update_v1(program_id: &Pubkey, ctx: Context<Update>, args: UpdateArgs) -> Pro
     // This account should always be a signer regardless of the authority type,
     // because at least one signer is required to update the metadata.
     assert_signer(ctx.accounts.authority_info)?;
+    // Note that payer is not checked because it is not used.
```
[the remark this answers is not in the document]
danenbm: Per discussion will plan to remove later.

token-metadata/program/src/processor/delegate/delegate.rs
```diff
@@ -97,6 +91,17 @@ pub fn delegate<'a>(
         } => Some((TokenDelegateRole::Staking, amount, authorization_data)),
         // Standard
         DelegateArgs::StandardV1 { amount } => Some((TokenDelegateRole::Standard, amount, &None)),
+        // LockedTransfer
```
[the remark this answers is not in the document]
danenbm: I agree enums must be kept in same order to prevent breaking serialization, but this section is just a match statement so could be in any order.  I changed the order of the match statement to look be more in the order of the enum.

I did not change any enum order, but I did change the name of two metadata delegates, and appended some to the end.  For the two that I changed, Sam ran a script to make sure nobody was using them yet.  The appended ones should be fine right?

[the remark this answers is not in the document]
blockiosaurus: Ah my bad, IDK why I thought this was the definition.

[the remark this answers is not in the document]
danenbm: Its all in the same vicinity in the PR.  Overall this is important issue to be looking out for so glad you are looking for that issue in general.  Thanks!

token-metadata/program/src/processor/metadata/update.rs
```diff
@@ -197,60 +251,136 @@ fn validate_update(
...
-                return Err(MetadataError::InvalidUpdateArgs.into());
+    if let Some(metadata_delegate_role) = metadata_delegate_role {
+        match metadata_delegate_role {
+            MetadataDelegateRole::Authority => {
+                // Fields allowed for `Authority`:
+                // `new_update_authority`
+                // `primary_sale_happened`
+                // `is_mutable`
+                // `token_standard`
+                if data.is_some()
```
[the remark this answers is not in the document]
danenbm: The main concept that stood out with @samuelvanderwaal 's solution above is the general concept of using type safety.  Iterated on the idea above and eventually landed on the idea to use different `UpdateArgs` struct types for different delegates.  This results in a bit of extra struct type handling, but the end result is that it is 100% clear to the user from the `UpdateArgs` what exact fields a delegate is allowed to pass into `Update`.  It is not necessary to read the `Update` code or consult docs to find out this information.

Discussed with team and made the changes to this PR.  Also @lorisleiva mentioned that different handlers for the different delegates can actually be generated for the JS SDK making it easier for a delegate to use.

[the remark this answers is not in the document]
danenbm: So now UpdateArgs looks like this:
```Rust
pub enum UpdateArgs {
    V1 {
        /// The new update authority.
        new_update_authority: Option<Pubkey>,
        /// The metadata details.
        data: Option<Data>,
        /// Indicates whether the primary sale has happened or not (once set to `true`, it cannot be
        /// changed back).
        primary_sale_happened: Option<bool>,
        // Indicates Whether the data struct is mutable or not (once set to `true`, it cannot be
        /// changed back).
        is_mutable: Option<bool>,
        /// Collection information.
        collection: CollectionToggle,
        /// Additional details of the collection.
        collection_details: CollectionDetailsToggle,
        /// Uses information.
        uses: UsesToggle,
        // Programmable rule set configuration (only applicable to `Programmable` asset types).
        rule_set: RuleSetToggle,
        /// Required authorization data to validate the request.
        authorization_data: Option<AuthorizationData>,
    },
    UpdateAuthorityV2(UpdateAuthorityV2),
    AuthorityItemDelegateV2(AuthorityItemDelegateV2),
    CollectionDelegateV2(CollectionDelegateV2),
    DataDelegateV2(DataDelegateV2),
    ProgConfigDelegateV2(ProgConfigDelegateV2),
    DataItemDelegateV2(DataItemDelegateV2),
    CollectionItemDelegateV2(CollectionItemDelegateV2),
    ProgConfigItemDelegateV2(ProgConfigItemDelegateV2),
}
```
And the individual structs look like this:
```Rust
pub struct AuthorityItemDelegateV2 {
    /// The new update authority.
    pub new_update_authority: Option<Pubkey>,
    /// Indicates whether the primary sale has happened or not (once set to `true`, it cannot be
    /// changed back).
    pub primary_sale_happened: Option<bool>,
    // Indicates Whether the data struct is mutable or not (once set to `true`, it cannot be
    /// changed back).
    pub is_mutable: Option<bool>,
    /// Token standard.
    pub token_standard: Option<TokenStandard>,
}
```

Rather than struct-like enum variants, I used tuple enum variants backed by actual struct types.  This is so that I can implement the `Default` trait separately for each struct and thus each variant. This in turn enables more ergonomic Rust handling such as instantiating with only setting the field I care about and default for the remainder:
```Rust
let update_args = UpdateArgs::from(UpdateAuthorityV2 {
    rule_set: RuleSetToggle::Clear,
    ..Default::default()
});
```

samuelvanderwaal (dismissed):

token-metadata/program/src/instruction/metadata.rs
```diff
@@ -100,20 +100,278 @@ pub enum UpdateArgs {
         /// Required authorization data to validate the request.
         authorization_data: Option<AuthorizationData>,
     },
+    UpdateAuthorityV2 {
```
febo, line 103: Would this not be just a `V2` version?

token-metadata/program/src/instruction/metadata.rs
```diff
@@ -100,20 +100,278 @@ pub enum UpdateArgs {
...
+        /// Uses information.
+        uses: UsesToggle,
+        // Programmable rule set configuration (only applicable to `Programmable` asset types).
+        rule_set: RuleSetToggle,
+        /// Token standard.
+        token_standard: Option<TokenStandard>,
+        /// Required authorization data to validate the request.
+        authorization_data: Option<AuthorizationData>,
+    },
+    AuthorityItemDelegateV2 {
```
febo, line 127: Why do we need the `Delegate` word in the name for the above ones? Are they not just args?

token-metadata/program/src/instruction/metadata.rs
```diff
@@ -100,20 +100,278 @@ pub enum UpdateArgs {
...
+        /// Required authorization data to validate the request.
+        authorization_data: Option<AuthorizationData>,
+    },
+    CollectionItemDelegateV2 {
+        /// Collection information.
+        collection: CollectionToggle,
+        /// Required authorization data to validate the request.
+        authorization_data: Option<AuthorizationData>,
+    },
+    ProgConfigItemDelegateV2 {
```
febo, line 171: I am not a big fan of abbreviated names. If we can drop the `Delegate` from the name, this one could be names `ProgrammableConfigItem`.

token-metadata/program/src/processor/metadata/update.rs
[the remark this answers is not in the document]
febo: I think we need the payer.

token-metadata/program/src/instruction/metadata.rs
febo: Having another thought about this after looking into the whole PR. What if the `V2` version had an inner struct that represent the fields to be updated instead of multiple args versions? E.g.,:
```
pub enum UpdateArgs {
    V1 { ... },
    V2 {
        fields: UpdateFields,
        authorization_data: Option<AuthorizationData>,
    }
}

pub enum UpdateFields {
    Collection {
        /// Collection information.
        collection: CollectionToggle,
    },
    Data {
        /// The metadata details.
        data: Option<Data>,
    }
    ...
}
```

lorisleiva: +1 one on removing the `Delegate` name. This will greatly affect the generated API for the end-user.

Right now a `AuthorityItemDelegateV2` variant within the `Update` handler will generate a method named: `updateAuthorityItemDelegateV2` which is super confusing.

Regarding nested enums: That's possible on Kinobi but the generated methods will be called "handler" + "variantLevel1" + "variantLevel2". So in the provided example, we'll have `updateV2Collection` and `updateV2Data` for instance.

danenbm: Discussed with @febo, @lorisleiva, @blockiosaurus, and @samuelvanderwaal offline and we decided to keep the overall `UpdateArgs` struct as the same one-level-deep, struct-like enum variants, use "V2" in the name rather than "V1", and also change the names to include the word "As" to indicate that the caller is doing the metadata update "as" the specified role:

```Rust
pub enum UpdateArgs {
    ...
    AsUpdateAuthorityV2 {
        /// The new update authority.
        new_update_authority: Option<Pubkey>,
        ...
    },
    AsAuthorityItemDelegateV2 {
        /// The new update authority.
        new_update_authority: Option<Pubkey>,
        ...
    },
    AsCollectionDelegateV2 {
        /// Collection information.
        collection: CollectionToggle,
        ...
    },
    AsDataDelegateV2 {
        /// The metadata details.
        data: Option<Data>,
       ...
    },
```

We chose this because:
* The one-level-deep struct-like variants is consistent with other instruction args types, such as `DelegateArgs`.
* It will result in better generated typescript names.
* It will make it clear to the caller how its working.  For example, they are updating "as the update authority", **not** that they are updating the update authority value itself.
* In the future we didn't want to have multiple versions in the same overall args, something like `V3 { fields: UpdateFieldsV2, ...}`.

We chose to keep it at V2 instead of  V1 and we also chose to not abbreviate "Programmable" in some of the args.

token-metadata/program/src/instruction/metadata.rs
danenbm: Dropped the abbreviation and put in the full "Programmable".  See other comment for more info.

token-metadata/program/src/instruction/metadata.rs
danenbm: See other comment for more info.

token-metadata/program/Cargo.toml
```diff
@@ -36,6 +36,7 @@ solana-sdk = "1.14"
 solana-program-test = "1.14"
 serde = { version = "1.0.147", features = ["derive"]}
 async-trait = "0.1.64"
+old-token-metadata = { package = "mpl-token-metadata", version = "=1.10.0", features = ["no-entrypoint"] }
```
lorisleiva, line 39: I see you're importing this to test backwards compatibility which is very clever. Just curious to know what the process with these "temporary" tests going forward? For instance, say we need to do something similar in a future PR, would we have to import `old-token-metadata` and `old-token-metadata-for-the-new-pr` to ensure they each have the required version? Or are we planning on cleaning up that test once merge?

lorisleiva (approved):
Thanks Michael! My approval vouches only for the client/JS side but I do really like the API that came out of this PR.

I left a little comment out of curiosity.

token-metadata/program/src/processor/metadata/update.rs
```diff
@@ -95,46 +89,77 @@ fn update_v1(program_id: &Pubkey, ctx: Context<Update>, args: UpdateArgs) -> Pro
...
+        .map(|Collection { key, .. }| key);
+
+    // Check if caller passed in a collection and if so use that.  Note that
+    // `validate_update` checks that the authority has permission to pass in
+    // a new collection value.
+    let collection_mint = match &args {
+        UpdateArgs::V1 { collection, .. }
+        | UpdateArgs::AsUpdateAuthorityV2 { collection, .. }
+        | UpdateArgs::AsCollectionDelegateV2 { collection, .. }
+        | UpdateArgs::AsCollectionItemDelegateV2 { collection, .. } => match collection {
```
blockiosaurus, line 158: Could this be collapsed?

blockiosaurus (approved):

token-metadata/program/src/processor/metadata/update.rs
danenbm: I don't think I can collapse it any further because its already piped into a single case.  My previous design used `InternalUpdateArgs` and used the From trait to destructure everything into one of those, so I had the superior code:
```
let internal_update_args = InternalUpdateArgs::from(args);
match internal_update_args.collection {
   ...
}
```
But unfortunately this added about 5k to the binary size.

I went back with @febo today and attempted to do it again without the `From` trait, implementing a partial use of `InternalUpdateArgs` to just avoid some of the duplicate destructuring, but even this used about 2k more bytes.  That way didn't change this particular collection destructuring but did help with another one and reduce the overall need for this type of destructure.  Here is that attempt for reference: https://github.com/metaplex-foundation/metaplex-program-library/pull/1079

Overall this raw destructuring and doing it multiple times, rather than DRY code with a new type, seems to result in the smallest binary size.

token-metadata/program/Cargo.toml
danenbm: I was going to remove the test up soon in a follow on PR.  I think we will have to remove them as cargo does not like multiple versions of a dependency.   For example, when I try adding a new `old-token-metadata-for-the-new-pr`, I get a "failed to select version" error:

```
    error: failed to select a version for `mpl-token-metadata`.
        ... required by package `mpl-token-metadata v1.10.0 (/home/danenbm/Metaplex-Workspace/metaplex-program-library/token-metadata/program)`
    versions that meet the requirements `=1.9.1` are: 1.9.1
    
    all possible versions conflict with previously selected packages.
    
      previously selected package `mpl-token-metadata v1.10.0`
        ... which satisfies dependency `old-token-metadata = "=1.10.0"` of package `mpl-token-metadata v1.10.0 (/home/danenbm/Metaplex-Workspace/metaplex-program-library/token-metadata/program)`
    
    failed to select a version for `mpl-token-metadata` which could resolve this conflict
```

febo (approved):
LGTM

samuelvanderwaal (approved):

Issue #1049, opened 2023-03-12
[Bug]: auction house listing asset error SaleRequiresExactlyOneSigner


boredjoker:
### Which package is this bug report for?

auction-house

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

when auction house requires_sign_off is true, how can we listing a aseet by user, signing with authority,and send tranction by user(as fee payer),why can't we let user and authority as both signer like last version?

1. last version success [tranction](https://solscan.io/tx/D4dBerAbcCefmwZUHD1zG54qFa9B4U8fiCSJBJvtvajx6WjBDxcQUi2q5jfUAeyhjs1yiErCgSC7odSrJ1JhMPg?cluster=devnet)

2. new version error [tranction](https://solscan.io/tx/53GowrhYMnZaXeFo6v5VhmY3tmwLGD83Uz8E4Uny41HNgR4o6yJcsF43mZ1xsReMU8VH2iYNq4tByF9zDjaqfSVb?cluster=devnet)
Error Code: SaleRequiresExactlyOneSigner. Error Number: 6044. Error Message: This sale requires exactly one signer: either the seller or the authority..

3. new version error without authority sign  [tranction](https://solscan.io/tx/4YyPPEznUYLaTuMrhQqZnTd1nWVsjtRDRchDQsg3CJhEcVAhh1dNYuy9BZbR8LzeR9DozE156MoHM7ogY8WGD33Q?cluster=devnet)
Error Code: CannotTakeThisActionWithoutAuctionHouseSignOff. Error Number: 6011. Error Message: Cannot take this action without auction house signing too.

### Priority this issue should have

High (immediate attention needed)

carrie-hk:
Hey we're getting the same error! Any update on this?

jdnichollsc:
Hello @samuelvanderwaal, do you need more details about this error? Thanks for your help, please let us know how can we help! <3

samuelvanderwaal:
@boredjoker @carrie-hk @jdnichollsc hey everyone, I'm opening a PR to remove the double signer restriction as there does not appear to be a good reason to keep that in place. Keep in mind that the fee payer still defaults to the auction house fee account with two signers and it's out of scope to refactor this as it affects more than just this handler. However, to work around this you can include a simple system transfer instruction from the user to the auction house fee account if you want the user to pay for the creation of the trade state account.

carrie-hk:
[...]

PR #1059, opened 2023-03-30
Fix/allow double signer

samuelvanderwaal:
Addresses #1049. There is no reason to require only one signer and there's a use case where an Auction House might require RSO but still want the user to submit the transaction.

### DIFF
--- a/auction-house/js/idl/auction_house.json
+++ b/auction-house/js/idl/auction_house.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.3.5",
+  "version": "1.4.0",
   "name": "auction_house",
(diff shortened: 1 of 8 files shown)


blockiosaurus (approved):

carrie-hk:
Thank you for this PR! We pulled this branch, compiled these contracts, ran locally and it passes :partying_face:

lorisleiva (approved):

jdnichollsc:
@lorisleiva @blockiosaurus thanks folks for approving this PR, thanks @samuelvanderwaal for your amazing help! <3

PR #1061, opened 2023-04-05
Fix/utility delegate

samuelvanderwaal:
## Issue
In the present system, a `UtilityDelegate` is granted `CloseAuthority` on the token account to enable token burning. Upon transfer, the token account remains open, and although the SPL token program resets the delegate, it does not clear the `CloseAuthority`. Consequently, re-delegation of the token fails due to a mismatch between the current close authority and the calling authority (i.e., the owner). The SPL program only permits the current `CloseAuthority` to change to a new close authority, and if none is set, it defaults to the owner. However, the owner is unable to modify the `CloseAuthority` once established.
Additionally, the owner cannot revoke a `UtilityDelegate`, which is an unintended side effect resulting from granting `UtilityDelegates` the burn capability. A feasible solution requires either clearing the `CloseAuthority` or closing the token account upon transfer. However, the `UtilityDelegate` is unable to perform these actions, as it is not a signer on either the Transfer or Revoke instructions. Adding it as an additional signer introduces a breaking change and an extra account, which is undesirable.

## Solution
The solution in this PR entails using the master edition account of the asset as the actual `CloseAuthority` stored on the token account. This allows Token Metadata to sign via Cross-Program Invocation (CPI) to clear the close authority or close the token account. This method is viable since the master edition account is already included in `Delegate`, `Revoke`, and `Transfer` instructions. The actual `UtilityDelegate` will continue to be stored in the token record for validation purposes, but upon validation, if the delegate role is `Utility`, the master edition PDA will be employed as the signer instead of the standard authority.

## Implementation
This solution updates `Delegate`, `Revoke`, and `Transfer` instructions to accommodate the proposed changes and return informative errors in cases where there is an existing `CloseAuthority` that is not the asset's master edition account. 

The implementation adds a new helper function to `utils/programmable_asset.rs` called `clear_close_authority`. It checks the provided token account for an existing close authority and fails if it's not set to the master edition account. Otherwise, it calls SPL token's `set_authority` instruction with the new authority set to `None`.

```rust
pub(crate) fn clear_close_authority(params: ClearCloseAuthorityParams) -> ProgramResult {
    let ClearCloseAuthorityParams {
        token,
        mint_info,
        token_info,
        master_edition_info,
        authority_info,
        spl_token_program_info,
    } = params;

    // If there's an existing close authority that is not the metadata account,
    // it willl need to be revoked by the original UtilityDelegate.
    if let COption::Some(close_authority) = token.close_authority {
        msg!("Existing close authority: {:?}", close_authority);
        msg!("Master edition: {:?}", master_edition_info.key);
        if &close_authority != master_edition_info.key {
            return Err(MetadataError::InvalidCloseAuthority.into());
        }
        msg!("Clearing close authority");
        let seeds = edition_seeds!(mint_info.key);

        invoke_signed(
            &spl_token::instruction::set_authority(
                spl_token_program_info.key,
                token_info.key,
                None,
                SplAuthorityType::CloseAccount,
                authority_info.key,
                &[],
            )?,
            &[token_info.clone(), authority_info.clone()],
            &[seeds.as_slice()],
        )?;
    } else {
        msg!("No close authority");
    }

    Ok(())
}
```

In addition, a new macro was added to `state/master_edition.rs` called `edition_seeds` which reduces the redundant code required to generate edition seeds in a manner useful to pass into `invoke_sign` calls as signing seeds.

```rust
#[macro_export]
macro_rules! edition_seeds {
    ($mint:expr) => {{
        let path = vec![
            "metadata".as_bytes(),
            $crate::ID.as_ref(),
            $mint.as_ref(),
            "edition".as_bytes(),
        ];
        let (_, bump) = Pubkey::find_program_address(&path, &$crate::ID);
        &[
            "metadata".as_bytes(),
            $crate::ID.as_ref(),
            $mint.as_ref(),
            "edition".as_bytes(),
            &[bump],
        ]
    }};
}
```

Finally, a new error was also added to inform the user of the case where there is an invalid close authority that is not the master edition pubkey:

```rust
    /// 188
    #[error("The close authority needs to be revoked by the Utility Delegate")]
    InvalidCloseAuthority,
```

### Delegate
When delegating a `UtilityDelegate` we request Close Authority and now assign it to the asset's master edition account. We check if there's an existing `CloseAuthority` that is not the master edition account and fail in that case as the delegate will need to clear the `CloseAuthority` manually in that case.

```rust
    // For Utility Delegates we request Close Authority as well so that the
    // token can be closed by the delegate on Burn. We assign CloseAuthority to
    // the metadata PDA so we can close it on Transfer and revoke it in Revoke.
    if matches!(role, TokenDelegateRole::Utility) {
        // If there's an existing close authority that is not the metadata account,
        // it willl need to be revoked by the original UtilityDelegate.
        if let COption::Some(close_authority) = token.close_authority {
            if &close_authority != ctx.accounts.metadata_info.key {
                return Err(MetadataError::InvalidCloseAuthority.into());
            }
        } else {
            let master_edition_info = ctx
                .accounts
                .master_edition_info
                .ok_or(MetadataError::MissingEditionAccount)?;

            invoke(
                &spl_token::instruction::set_authority(
                    spl_token_program_info.key,
                    token_info.key,
                    Some(master_edition_info.key),
                    SplAuthorityType::CloseAccount,
                    ctx.accounts.authority_info.key,
                    &[],
                )?,
                &[
                    token_info.clone(),
                    ctx.accounts.delegate_info.clone(),
                    ctx.accounts.authority_info.clone(),
                ],
            )?;
        }
    }
```

### Revoke
When revoking we check if it's a `UtilityDelegate` and ensure we clear the `CloseAuthority` in addition to the rest of the delegate data.

```rust
                // Clear the close authority if it's a Utility Delegate.
                if matches!(role, TokenDelegateRole::Utility) {
                    clear_close_authority(ClearCloseAuthorityParams {
                        token_info,
                        mint_info: ctx.accounts.mint_info,
                        token,
                        master_edition_info,
                        authority_info: master_edition_info,
                        spl_token_program_info,
                    })?;
                }
```

### Transfer
Transfers of pNFTs now 1) clear the close authority if present and 2) close the source token record. The first is needed as the rest of the delegate data is cleared on transfer so this keeps our delegate abstraction synced up with the underlying SPL token delegate. The latter is just a rent reclamation change and something users have requested.

Token accounts are not closed in the current implementation because they cannot be closed consistently in all cases. They can only be closed on owner transfers because only owners or the master edition as close authority can close token accounts. Transfer delegates such as Sale and Transfer don't have close authority and are not the owner of the token account so cannot close it.

### DIFF
--- a/token-metadata/program/src/processor/delegate/delegate.rs
+++ b/token-metadata/program/src/processor/delegate/delegate.rs
@@ -407,20 +407,20 @@ fn create_persistent_delegate_v1(
 
     // For Utility Delegates we request Close Authority as well so that the
     // token can be closed by the delegate on Burn. We assign CloseAuthority to
-    // the metadata PDA so we can close it on Transfer and revoke it in Revoke.
+    // the master edition PDA so we can close it on Transfer and revoke it in Revoke.
     if matches!(role, TokenDelegateRole::Utility) {
         // If there's an existing close authority that is not the metadata account,
-        // it willl need to be revoked by the original UtilityDelegate.
+        // it will need to be revoked by the original UtilityDelegate.
+        let master_edition_info = ctx
+            .accounts
+            .master_edition_info
+            .ok_or(MetadataError::MissingEditionAccount)?;
+
         if let COption::Some(close_authority) = token.close_authority {
-            if &close_authority != ctx.accounts.metadata_info.key {
+            if &close_authority != master_edition_info.key {
                 return Err(MetadataError::InvalidCloseAuthority.into());
             }
         } else {
-            let master_edition_info = ctx
-                .accounts
-                .master_edition_info
-                .ok_or(MetadataError::MissingEditionAccount)?;
-
             invoke(
                 &spl_token::instruction::set_authority(
                     spl_token_program_info.key,
--- a/token-metadata/program/src/processor/delegate/delegate.rs
+++ b/token-metadata/program/src/processor/delegate/delegate.rs
@@ -4,8 +4,8 @@ use borsh::BorshSerialize;
 use mpl_token_auth_rules::utils::get_latest_revision;
 use mpl_utils::{assert_signer, create_or_allocate_account_raw};
 use solana_program::{
-    account_info::AccountInfo, entrypoint::ProgramResult, msg, program::invoke,
-    program_option::COption, program_pack::Pack, pubkey::Pubkey, system_program, sysvar,
+    account_info::AccountInfo, entrypoint::ProgramResult, program::invoke, program_option::COption,
+    program_pack::Pack, pubkey::Pubkey, system_program, sysvar,
 };
 use spl_token::{instruction::AuthorityType as SplAuthorityType, state::Account};
 
@@ -409,7 +409,6 @@ fn create_persistent_delegate_v1(
     // token can be closed by the delegate on Burn. We assign CloseAuthority to
     // the metadata PDA so we can close it on Transfer and revoke it in Revoke.
     if matches!(role, TokenDelegateRole::Utility) {
-        msg!("Utility delegate");
         // If there's an existing close authority that is not the metadata account,
         // it willl need to be revoked by the original UtilityDelegate.
         if let COption::Some(close_authority) = token.close_authority {
--- a/token-metadata/program/src/processor/delegate/delegate.rs
+++ b/token-metadata/program/src/processor/delegate/delegate.rs
@@ -4,8 +4,8 @@ use borsh::BorshSerialize;
 use mpl_token_auth_rules::utils::get_latest_revision;
 use mpl_utils::{assert_signer, create_or_allocate_account_raw};
 use solana_program::{
-    account_info::AccountInfo, entrypoint::ProgramResult, program::invoke, program_pack::Pack,
-    pubkey::Pubkey, system_program, sysvar,
+    account_info::AccountInfo, entrypoint::ProgramResult, msg, program::invoke,
+    program_option::COption, program_pack::Pack, pubkey::Pubkey, system_program, sysvar,
 };
 use spl_token::{instruction::AuthorityType as SplAuthorityType, state::Account};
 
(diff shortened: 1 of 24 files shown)


token-metadata/program/src/utils/mod.rs
```diff
@@ -288,3 +288,49 @@ mod tests {
...
+    ($option:expr, $err:expr) => {
+        match $option {
+            Ok(val) => val,
+            Err(_) => return Err($err),
+        }
+    };
+);
+
+#[macro_export]
+macro_rules! set_close_authority {
```
febo, line 303: This does not seem to be used.

samuelvanderwaal: Yeah, I will remove it thanks.

token-metadata/program/src/utils/programmable_asset.rs
```diff
@@ -332,3 +336,47 @@ pub fn frozen_transfer<'a, 'b>(
...
+        token,
+        mint_info,
+        token_info,
+        master_edition_info,
+        authority_info,
+        spl_token_program_info,
+    } = params;
+
+    // If there's an existing close authority that is not the metadata account,
+    // it willl need to be revoked by the original UtilityDelegate.
```
febo, line 360: Typo `willl` -> `will`

febo (dismissed):
This PR set the master edition as the close authority when a `Utility` delegate is set, allowing the delegate to close the token account. Since the close authority is the master edition, it can be cleared on transfer.

token-metadata/program/tests/revoke.rs
```diff
@@ -518,4 +518,147 @@ mod revoke {
...
+        let token_record: TokenRecord = try_from_slice_unchecked(&pda.data).unwrap();
+
+        assert_eq!(token_record.delegate, None);
+
+        if let Some(token) = asset.token {
+            let account = get_account(&mut context, &token).await;
+            let token_account = Account::unpack(&account.data).unwrap();
+
+            assert!(token_account.is_frozen());
+            assert_eq!(token_account.delegate, COption::None);
```
danenbm, line 595: Should there also be `assert!(token_account.close_authority.is_none())` here as well?

danenbm (commented):
LGTM; mostly nits, one test question, and one overall idea question (see inline comments)

token-metadata/program/tests/revoke.rs
```diff
@@ -518,4 +518,147 @@ mod revoke {
...
+        let (pda_key, _) = find_token_record_account(&asset.mint.pubkey(), &asset.token.unwrap());
+
+        let pda = get_account(&mut context, &pda_key).await;
+        let token_record: TokenRecord = try_from_slice_unchecked(&pda.data).unwrap();
+
+        assert_eq!(token_record.delegate, Some(user_pubkey));
+
+        let payer = Keypair::from_bytes(&context.payer.to_bytes()).unwrap();
+        let approver = Keypair::from_bytes(&context.payer.to_bytes()).unwrap();
+
```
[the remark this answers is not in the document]
samuelvanderwaal: It's redundant with `delegate` and if it's not set properly revoke will fail anyway because it won't have the proper close authority to clear the close authority.

token-metadata/program/tests/revoke.rs
samuelvanderwaal: Yeah, good idea.

token-metadata/program/src/processor/delegate/delegate.rs
```diff
@@ -406,23 +406,37 @@ fn create_persistent_delegate_v1(
     )?;
 
     // For Utility Delegates we request Close Authority as well so that the
-    // token can be closed by the delegate on Burn.
+    // token can be closed by the delegate on Burn. We assign CloseAuthority to
+    // the master edition PDA so we can close it on Transfer and revoke it in Revoke.
     if matches!(role, TokenDelegateRole::Utility) {
```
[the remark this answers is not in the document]
samuelvanderwaal: I thought about this but it seems like it might be unexpected behavior from a user's perspective. A normal SPL token delegate does not get close authority so it might be surprising if all Token Metadata delegate abstractions did. It would mean that users cannot close their token accounts themselves and might break integrations with marketplaces which use delegates if they are automatically closing and opening token accounts for their users.

danenbm (approved):

febo (approved):

PR #1064, opened 2023-04-16
fix creator hash js fn for bubblegum

ilmoi:
I found that the current function only works for 0 or 1 creators
once you go to 2 or more it fails to generate the correct creator hash, which leads to account_compression::verify_leaf failing

the test currently has an empty array of Creator[] which is why this fell through the cracks, unless I'm missing something ofc


danenbm (dismissed):

bubblegum/js/src/mpl-bubblegum.ts
```diff
@@ -22,17 +22,16 @@ export function computeDataHash(metadata: MetadataArgs): Buffer {
...
-export function computeCreatorHash(creators: Creator[]): Buffer {
-  let bufferOfCreatorData = Buffer.from([]);
-  let bufferOfCreatorShares = Buffer.from([]);
-  for (const creator of creators) {
-    bufferOfCreatorData = Buffer.concat([
-      bufferOfCreatorData,
-      creator.address.toBuffer(),
-      Buffer.from([creator.share]),
-    ]);
-    bufferOfCreatorShares = Buffer.concat([bufferOfCreatorShares, Buffer.from([creator.verified ? 1 : 0]), Buffer.from([creator.share])]);
```
lorisleiva, line 34: Strange, it looks like this variable was never actually used within the function.

lorisleiva (changes requested):
Thanks for that! I'm not too familiar with this repo but it does look like a valid bug fix. Would you be able to add a quick regression test that used to fail before this PR but now passes?

danenbm:
> Thanks for that! I'm not too familiar with this repo but it does look like a valid bug fix. Would you be able to add a quick regression test that used to fail before this PR but now passes?

Seems like we can just add some creators in one of the tests so that it exercises that hashing function in the test.  I tried doing it briefly but having some other issue at the moment and will come back to it unless you add the test @ilmoi before I get to it
```
    let creators: Creator[] = [
      {
        address: payer.publicKey,
        share: 55,
        verified: false,
      },
      {
        address: new Keypair().publicKey,
        share: 45,
        verified: false,
      },
    ]
```

danenbm:
> > Thanks for that! I'm not too familiar with this repo but it does look like a valid bug fix. Would you be able to add a quick regression test that used to fail before this PR but now passes?
> 
> Seems like we can just add some creators in one of the tests so that it exercises that hashing function in the test. I tried doing it briefly but having some other issue at the moment and will come back to it unless you add the test @ilmoi before I get to it
> 
> ```
>     let creators: Creator[] = [
>       {
>         address: payer.publicKey,
>         share: 55,
>         verified: false,
>       },
>       {
>         address: new Keypair().publicKey,
>         share: 45,
>         verified: false,
>       },
>     ]
> ```

I added this code above which exercises the function for 2 creators.  It causes a test failure before this PR's changes, and passes with the commits from this PR.

lorisleiva (approved):
Thanks Michael!

samuelvanderwaal (dismissed):

samuelvanderwaal (approved):

danenbm (approved):
I added test as well.  LGTM

PR #1066, opened 2023-04-19
Use different UpdateArg types for different delegates

danenbm:
Use different `UpdateArgs` enum values for each delegate to control what fields can be updated.

Originally inspired by @samuelvanderwaal 's statement
> Defining these by negative fields feels a bit repetitive and perhaps easy to make a mistake. Is there a cleaner way to solve this with a HashSet of allow values for each type or something?

And thinking about type safety as a Rust idiom to solve this problem.


danenbm:
Reviewed with team, going with this method

PR #1067, opened 2023-04-20
add assert holder fn

samuelvanderwaal:


blockiosaurus (commented):
This assert checks that the person is actually a holder of the token.

blockiosaurus (approved):

Issue #1070, opened 2023-04-21
[Bug]: Bubblegum Decompression does not work for verified Collection NFTs


ethyi:
### Which package is this bug report for?

token-metadata

### Which Type of Package is this bug report for?

Rust Contract

### Issue description

1. Mint a standard collection with master edition with your wallet
2. Create a merkle tree through bubblegum with your wallet
3. Mint a compressed NFT with the MintToCollection instruction on bubblegum with the same collection and tree (must specify {verified:false] and the ixn will automatically verify the collection if collection auth matches signer)
4. Attempt to decompress NFT into the same metadata {verified:true}
You will observe that the txn fails, with error "Cannot verify collection in this ixn"

### Priority this issue should have

High (immediate attention needed)

danenbm:
Getting fixed in https://github.com/metaplex-foundation/metaplex-program-library/pull/1087

danenbm:
With the recent patch fix to Token Metadata ( v1.11.2 on May 26, 2023), compressed NFTs can also now be decompressed with with collection set to verified.

PR #1072, opened 2023-04-21
Bubblegum collection docs fix

ethyi:
Bubblegum docs change that increases visibility for https://github.com/metaplex-foundation/metaplex-program-library/issues/1070


danenbm (approved):
Thank you for updating these docs.  We are working on how to deal with verified collections for the longer term.

samuelvanderwaal (approved):

PR #1075, opened 2023-04-24
add verified to computeCreatorHash

kevarifin14:
`computeCreatorHash` did not include `verified` in the hash so it was causing issues verifying a compressed NFT with the `createVerifyLeafIx`. This was not an issue when verifying a compressed NFT with a single creator but would not verify a compressed NFT with multiple creators.

### DIFF
--- a/bubblegum/js/src/mpl-bubblegum.ts
+++ b/bubblegum/js/src/mpl-bubblegum.ts
@@ -24,14 +24,13 @@ export function computeDataHash(metadata: MetadataArgs): Buffer {
 
 export function computeCreatorHash(creators: Creator[]): Buffer {
   let bufferOfCreatorData = Buffer.from([]);
-  let bufferOfCreatorShares = Buffer.from([]);
   for (const creator of creators) {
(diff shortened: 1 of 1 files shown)


kevarifin14:
Chatted with @ngundotra about this issue if further context is needed

ngundotra:
@danenbm this is a genuine fix of publicly exposed function in MPL Bubblegum SDK

danenbm:
> Chatted with @ngundotra about this issue if further context is needed

> `computeCreatorHash` did not include `verified` in the hash so it was causing issues verifying a compressed NFT with the `createVerifyLeafIx`. This was not an issue when verifying a compressed NFT with a single creator but would not verify a compressed NFT with multiple creators.

Thank you for fixing this @kevarifin14, and thank you for reviewing @ngundotra.  As it turns out there was a previous PR for this change: https://github.com/metaplex-foundation/metaplex-program-library/pull/1064 which we did the testing on and already merged.  So we ended up merging that copy of it, so I will close this one now but the issue should be fixed as part of that one which we will deploy soon.  Thanks again.

PR #1078, opened 2023-04-26
Fix isSigner for leafOwner account of transfer instruction

lorisleiva:
We had a report that the generated Solita library had the wrong `isSigner` boolean for the `leafOwner` of the `transfer` instruction of the Bubblegum program. Since this information comes directly from the program, I updated the relevant instruction account and re-generated the Solita library.

Note that the generated IDL happens to include new seed values as well as I'm guessing it wasn't done previously when bumping Anchor to the latest version.

### DIFF
--- a/bubblegum/js/idl/bubblegum.json
+++ b/bubblegum/js/idl/bubblegum.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.6.0",
+  "version": "0.7.0",
   "name": "bubblegum",
   "instructions": [
     {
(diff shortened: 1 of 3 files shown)


danenbm (commented):
The transfer instruction has its own signer logic in the processor, which allows for either the owner or the delegate to be a signer.  That is why it wasn't set to be `Signer` in the account validation struct.  Is there potentially a different ts strategy for the generated Solita lib that works with this approach?
```
pub fn transfer<'info>(
    ...
    // Transfers must be initiated by either the leaf owner or leaf delegate.
    require!(
        owner.is_signer || delegate.is_signer,
        BubblegumError::LeafAuthorityMustSign
    );
```

lorisleiva:
Thanks for clarifying Michael! No, Solita does not support that but Kinobi does. You can have `isSigner: "either"` and it will generate code that accepts both `PublicKey | Signer` and only sign the transaction if a `Signer` is provided.

PR #1079, opened 2023-04-28
`Update` idea: try using `InternalUpdateArgs` again to avoid multiple destructure

danenbm:
Idea to bring back `InternalUpdateArgs`, but instead of implementing `From` trait like last time, use a default() and then populate as we go in `validate_update`.

This prevents us from having to destructure the UpdateArgs a second time within Metadata::update_v1.

However, this increases the binary size from 700240 to 702752 so I think we should avoid this implementation in the short term.


danenbm:
Omitted this from https://github.com/metaplex-foundation/metaplex-program-library/pull/1058 due to size constraints.

PR #1082, opened 2023-05-02
Add token amount check to `Update` and more tests

danenbm:
The token account is used to make sure there isn't currently a delegate before allowing a RuleSet change.  Checking the amount is needed to make sure a previous token account (that has no token) isn't used for this check.  It could also affect the `Holder` authority derivation, but that functionality is not implemented anyways.

Also added a test for this and a few more tests around collection delegate.


samuelvanderwaal (approved):

febo (approved):
LGTM

PR #1094, opened 2023-05-12
Adding Print Instruction

blockiosaurus:
This is a staging PR that only adds the Print Instruction for pNFTs. I attempted to reuse existing functions when at all possible in order to reduce size of code increase.


token-metadata/program/src/instruction/metadata.rs
```diff
@@ -896,3 +903,63 @@ impl InstructionBuilder for super::builders::Update {
...
 }
+
+/// Prints an edition from a master edition.
+///
+/// # Accounts:
+///
+///   0. `[writable]` Edition Metadata
+///   1. `[writable]` Edition
+///   2. `[writable]` Edition Mint
+///   3. `[writable]` Edition Token Account Owner
```
febo, line 914: Does the account owner need to be `writable`?

token-metadata/program/src/utils/master_edition.rs
```diff
@@ -414,6 +482,7 @@ pub fn mint_limited_edition<'a>(
         true,
         true,
         None, // Not a collection parent
+        token_standard_override,
```
febo, line 485: Can the token standard be `None` at this point?

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,194 @@
...
+                payer_info.clone(),
+                edition_token_account_owner_info.clone(),
+                edition_mint_info.clone(),
+                edition_token_account_info.clone(),
+            ],
+        )?;
+    } else {
+        assert_owned_by(edition_token_account_info, &spl_token::id())?;
+    }
+
```
febo, line 116: Do we need to assert that the `edition_token_account_info` is initialized?

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,194 @@
...
+                    system_program,
+                )?;
+            } else {
+                assert_owned_by(token_record_info, &crate::ID)?;
+            }
+        }
+        _ => return Err(MetadataError::InvalidTokenStandard.into()),
+    };
+
+    process_mint_new_edition_from_master_edition_via_token_logic(
```
febo, line 162: It would be good to add a comment here to say that most of the accounts validation happen on the `process_mint_new_edition_from_master_edition_via_token_logic`.

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    if sysvar_instructions.key != &sysvar::instructions::ID {
+        return Err(ProgramError::IncorrectProgramId);
+    }
+
+    // Deserialize the master edition's metadata so we can determine token type
+    let master_metadata = Metadata::from_account_info(master_metadata_info)?;
+    let token_standard = master_metadata
+        .token_standard
+        .unwrap_or(TokenStandard::NonFungible);
```
febo, line 129: Is this assumption that accurate? Like, is it worth validating the master edition here to make sure?

token-metadata/program/src/processor/metadata/print.rs
febo: I think this is done on `process_mint_new_edition_from_master_edition_via_token_logic`, but I will leave the comment here to be sure.

token-metadata/program/src/instruction/mod.rs
```diff
@@ -774,6 +774,29 @@ pub enum MetadataInstruction {
...
+    #[account(7, writable, name="master_edition", desc="Master Record Edition V2 (pda of ['metadata', program id, master metadata mint id, 'edition'])")]
+    #[account(8, writable, name="edition_marker_pda", desc="Edition pda to mark creation - will be checked for pre-existence. (pda of ['metadata', program id, master metadata mint id, 'edition', edition_number]) where edition_number is NOT the edition number you pass in args but actually edition_number = floor(edition/EDITION_MARKER_BIT_SIZE).")]
+    #[account(9, signer, writable, name="payer", desc="payer")]
+    #[account(10, signer, name="master_token_account_owner", desc="owner of token account containing master token")]
+    #[account(11, name="master_token_account", desc="token account containing token from master metadata mint")]
+    #[account(12, name="master_metadata", desc="Master record metadata account")]
+    #[account(13, name="update_authority", desc="The update authority of the master edition.")]
+    #[account(14, name="token_program", desc="Token program")]
+    #[account(15, name="ata_program", desc="SPL Associated Token Account program")]
```
febo, line 795: It would be nice to use the same naming as the other instructions.

token-metadata/program/tests/print.rs
```diff
@@ -0,0 +1,456 @@
...
+    state::{Key, MAX_MASTER_EDITION_LEN},
+};
+use num_traits::FromPrimitive;
+use solana_program_test::*;
+use solana_sdk::{
+    account::AccountSharedData, instruction::InstructionError, transaction::TransactionError,
+};
+use utils::*;
+
+// NOTE: these tests depend on the token-vault program having been compiled
```
samuelvanderwaal, line 16: This is copy paste and the comment can be removed both here and in the original as Token Vault is deprecated and no longer a required dependency.

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,194 @@
...
+    let edition_token_account_info = ctx.accounts.edition_token_account_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let edition_mint_authority_info = ctx.accounts.edition_mint_authority_info;
+    let edition_token_record_info = ctx.accounts.edition_token_record_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let master_edition_info = ctx.accounts.master_edition_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let edition_marker_pda_info = ctx.accounts.edition_marker_pda_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let payer_info = ctx.accounts.payer_info;
```
danenbm, line 67: Should there be a check for payer is signer in `process_mint_new_edition_from_master_edition_via_token_logic`?  I think it would probably fail when it uses the payer, but it might be more helpful for there to be an explicit payer is signer check earlier.

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,194 @@
...
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let master_token_account_owner_info = ctx.accounts.master_token_account_owner_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let master_token_account_info = ctx.accounts.master_token_account_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let master_metadata_info = ctx.accounts.master_metadata_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let update_authority_info = ctx.accounts.update_authority_info;
```
danenbm, line 75: I need to study this more but I don't see where this is checked in `process_mint_new_edition_from_master_edition_via_token_logic` and then I wonder why it can't just be gotten from `master_metadata_info` when we are printing.  Like, you don't want to specify and thus be able to change the update authority when printing right?

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,194 @@
...
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let token_program = ctx.accounts.token_program_info;
+    let ata_program = ctx.accounts.ata_program_info;
+    let sysvar_instructions = ctx.accounts.sysvar_instructions_info;
+    // CHECK: Checked in process_mint_new_edition_from_master_edition_via_token_logic
+    let system_program = ctx.accounts.system_program_info;
```
danenbm, line 81: It seems like all the other program IDs are explicitly checked but I don't see where this one is explicitly checked in `process_mint_new_edition_from_master_edition_via_token_logic` or elsewhere.  I'm pretty sure many things don't work if you pass in the wrong system program but might be more debuggable if there was an explicit check like with other programs.

danenbm (changes requested):

samuelvanderwaal:
The `burn.rs` test isn't compiling but not sure that needs to be fixed in this PR, just wanted to note it.

token-metadata/program/src/state/mod.rs
```diff
@@ -83,7 +86,8 @@ pub trait TokenMetadataAccount: BorshDeserialize {
         let key: Option<Key> = Key::from_u8(data[0]);
         match key {
             Some(key) => {
-                (key == data_type || key == Key::Uninitialized) && (data.len() == data_size)
+                (key == data_type || key == Key::Uninitialized)
+                    && (data.len() == data_size || data_size == 0)
```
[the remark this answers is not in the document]
blockiosaurus: Resizeable accounts don't have a set size and so the associated method doesn't really work. So instead I'm returning zero. Maybe we should consider making size return a Toggle type?

token-metadata/program/src/processor/metadata/print.rs
blockiosaurus: It's used deep in the bowels of the print code. I originally didn't have this because we decided to get rid of the ability to set the UA on print but it seemed like it'd require a significant refactor to remove. I'll try to find the usage.

token-metadata/program/src/utils/master_edition.rs
```diff
@@ -46,6 +47,7 @@ pub fn process_mint_new_edition_from_master_edition_via_token_logic<'a>(
     accounts: MintNewEditionFromMasterEditionViaTokenLogicArgs<'a>,
     edition: u64,
     ignore_owner_signer: bool,
```
danenbm, line 49: Can we just get rid of this as it appears to be set to `false` all the time in the code?

danenbm (changes requested):

token-metadata/program/src/instruction/metadata.rs
blockiosaurus: It shouldn't be, good catch.

token-metadata/program/src/processor/burn/burn.rs
```diff
@@ -231,6 +232,8 @@ fn burn_v1(program_id: &Pubkey, ctx: Context<Burn>, args: BurnArgs) -> ProgramRe
                 &ctx.accounts.authority_info.clone(),
             )?;
         }
+        TokenStandard::ProgrammableNonFungibleEdition => {
```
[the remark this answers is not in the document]
blockiosaurus: Correct, I can add a todo macro.

token-metadata/program/src/processor/edition/mint_new_edition_from_master_edition_via_token.rs
```diff
@@ -50,5 +50,6 @@ pub fn process_mint_new_edition_from_master_edition_via_token<'a>(
         },
         edition,
         ignore_owner_signer,
+        crate::state::TokenStandard::NonFungibleEdition,
```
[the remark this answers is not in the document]
blockiosaurus: Yeah that makes sense. I must have moved this code around. At one point I was trying to avoid deserializing Metadata twice but it looks like that's not necessary anyway.

token-metadata/program/src/utils/master_edition.rs
blockiosaurus: I believe this is an argument in the old edition printing instruction.

token-metadata/program/src/utils/master_edition.rs
```diff
@@ -96,49 +98,105 @@ pub fn process_mint_new_edition_from_master_edition_via_token_logic<'a>(
...
-    }
+            let mut edition_marker = EditionMarker::from_account_info(edition_marker_info)?;
+            edition_marker.key = Key::EditionMarker;
+            if edition_marker.edition_taken(edition)? {
+                return Err(MetadataError::AlreadyInitialized.into());
+            } else {
+                edition_marker.insert_edition(edition)?
+            }
+            edition_marker.serialize(&mut *edition_marker_info.data.borrow_mut())?;
+            Ok::<(), ProgramError>(())
```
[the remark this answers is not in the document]
blockiosaurus: It required it for at least one of the Oks. I added it to both for consistency.

token-metadata/program/src/processor/metadata/print.rs
blockiosaurus: Good idea!

token-metadata/program/src/processor/metadata/print.rs
blockiosaurus: There are // CHECK comments at the top of the function

token-metadata/program/src/utils/master_edition.rs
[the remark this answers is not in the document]
danenbm: Its only needed because `?` is on the match statement, so it needs the fully qualified types.  If that is removed then you don't need to have any `Ok` here.

token-metadata/program/src/utils/master_edition.rs
danenbm: Yes I agree but it is hardcoded to `false` there as well.   So it is never used in any of the code, it is always `false`.

token-metadata/program/src/processor/metadata/print.rs
danenbm: Per offline discussion, just for record in this PR, we previously decided that with the new unified `Print` instruction that the creator should retain UA.  Since in the depths of the code it checks `is_signer` imo you could leave the account and just add a check in the code that `update_authority_info.key == master_metadata.update_authority`

token-metadata/program/src/processor/metadata/print.rs
danenbm: What is someone passed in an existing Print Edition that had a token standard of `None`, as the Master Edition?  I don't see a hardline check that we are in fact using a Master Edition.  Going down into the code, all I see is manual deserialization in functions such as `get_max_supply_off_master_edition`, which most likely fail if the account was an `Edition` rather than a `MasterEditionV2`, but still I haven't seen explicit check so it feels like using a side effect to guarantee safety.

As an alternative, in `Update` it is an explicit check using these utility functions: https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/utils/mod.rs#L44-L85

In conclusion, you might want to consider adding `is_master_edition` here which is what I also interpret from the previous comments as well.

danenbm (changes requested):

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -0,0 +1,176 @@
...
+    let master_metadata = Metadata::from_account_info(master_metadata_info)?;
+    let token_standard = master_metadata
+        .token_standard
+        .unwrap_or(TokenStandard::NonFungible);
+
+    match token_standard {
+        TokenStandard::NonFungible => {}
+        TokenStandard::ProgrammableNonFungible => {
+            // Validate that the token record was passed in for pNFTs.
+            let token_record_info = match edition_token_record_info {
```
febo, line 119: This can be simplified to:
```rust
let token_record_info = edition_token_record_info.ok_or(MetadataError::MissingTokenRecord)?;
```

token-metadata/program/src/utils/metadata.rs
```diff
@@ -155,14 +156,21 @@ pub fn process_create_metadata_accounts_logic(
...
+                    TokenStandard::FungibleAsset
+                } else {
+                    TokenStandard::Fungible
+                };
+                metadata.token_standard = Some(token_standard);
+            }
+        }
     } else {
         metadata.token_standard = None;
     }
```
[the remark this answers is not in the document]
blockiosaurus: I like it!

samuelvanderwaal (approved):

samuelvanderwaal (approved):

token-metadata/program/src/utils/master_edition.rs
```diff
@@ -96,49 +99,117 @@ pub fn process_mint_new_edition_from_master_edition_via_token_logic<'a>(
...
+            } else {
+                edition_marker.insert_edition(edition)?
+            }
+            edition_marker.serialize(&mut *edition_marker_info.data.borrow_mut())?;
+        }
+        TokenStandard::ProgrammableNonFungible => {
+            // Check that the new update authority is the same as the master edition.
+            if update_authority_info.key != &master_metadata.update_authority {
+                return Err(MetadataError::UpdateAuthorityIncorrect.into());
+            }
```
danenbm, line 162: From the PRD discussion, I was originally thinking this was a property of the new `Print` handler, not a property of pNFT specifically.  wdyt about doing it at the `Print` level?

danenbm (changes requested):
Overall looking close to finished, just had one last comment

token-metadata/program/src/processor/metadata/print.rs
blockiosaurus: It wasn't checked so added!

danenbm (approved):

Issue #1091, opened 2023-05-10
Why separate edge case for Migration token delegate role?


fedorerd:
I just faced some issues working with `revoke` within my own program linked to recently migrated NFTs. I assume similar problem exists with other instructions also. 

So, I have a staking program and I am invoking `revoke` with `RevokeArgs::StakingV1`. Since PNFTs are locked with Staking delegate role. However, if I pass `RevokeArgs::StakingV1` for recently migrated NFT (which has `Migration` delegate role), instructions will fail here:
https://github.com/metaplex-foundation/metaplex-program-library/blob/3a9e2eac5219eb20eccdd7a762fe24e6867c28f3/token-metadata/program/src/processor/delegate/revoke.rs#L243

Your program checks if role provided in revoke args matches actual delegate role or not. However, I don't understand why is it so, since Migration delegate has same rules for revoke as other delegates. 

I'm talking only about migration delegate, since it is one-time state for any token. 

So my program usually will pass `StakingV1` arg and expect staking delegate role, however due to design I need to add an edge case in my program and check status of delegate in token record on my side. I think it would be much better for everyone if this edge case was checked on your side instead.

And there is no evidence for how long I will need to keep this edge case check, since I literally need all migrated while staked in my program NFTs to be unstaked once.

So, basically, I think check like that might be more convenient:
```rust
if token_record.delegate_role == Some(role) || token_record.delegate_role == Some(TokenDelegateRole::Migration) {
  ...
}
```

samuelvanderwaal:
@fedorerd This is a good point and I opened a PR to fix it.

samuelvanderwaal:
This fix is on mainnet now in the 1.11.2 release.

PR #1096, opened 2023-05-15
add migration role exception

samuelvanderwaal:
Fixes the issue raised in #1091. The `revoke` instruction doesn't actually strictly need the role at all other than to distinguish between `TokenRecordDelegate` and `MetadataDelegate` types, but refactoring that would be an API change and would increase binary size so is best left until we get some other more critical features merged first and deployed first.

### DIFF
--- a/token-metadata/program/src/processor/delegate/revoke.rs
+++ b/token-metadata/program/src/processor/delegate/revoke.rs
@@ -48,7 +48,7 @@ pub fn revoke<'a>(
     };
 
     if let Some(role) = token_delegate {
-        // proceed with the delegate creation if we have a match
+        // proceed with the delegate revoke if we have a match
         return revoke_persistent_delegate_v1(program_id, context, role);
     }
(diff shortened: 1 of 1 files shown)


danenbm (approved):

febo (approved):

fedorerd:
Hey, thanks for the PR. 

I forgot to mention this thing in my issue.

If it is Utility delegate, 
https://github.com/metaplex-foundation/metaplex-program-library/blob/523c6144e4ab50c0a735bc2bf9c609aa372bfa34/token-metadata/program/src/processor/delegate/revoke.rs#L269
u're updating master edition close authority.

So in case of Migration delegates, there is no need to update it, right? Therefore assertion should be not against argument passed by invoke, but against the value in delegate account.

Of course not changing this part won't break anything, but it might reduce inner instructions count, which might be important since pnfts operations are already expensive.

PR #1106, opened 2023-05-22
Token Metadata v1.12 -- Initial Fees

samuelvanderwaal:
# Token Metadata v1.12 Summary

Adds the system_program and sysvar_instruction accounts to all instructions that don't already have them to facilitate future fee-taking.

Levies fees on the following instructions:

* create_metadata_accounts_v3
* create

"Print" commands, which create edition NFTs, will have the same fee applied but will be shipped after the pNFT unified `Print` command is implemented for consistency.

~~Makes edition account mutable for the following instructions:~~ 

* ~~freeze_delegated_account~~
* ~~thaw_delegated_account~~

~~Makes metadata account mutable for the following instructions:~~

* ~~approve_collection_authority~~
* ~~revoke_collection_authority~~
* ~~approve_use_authority~~
* ~~revoke_use_authority~~

Writes a Fee Flag to the last byte of the Metadata account or the second to last byte of the Master Edition account to indicate the account has fees to be collected.

### Strategy

~~For most instructions, we need the `system_program` account passed in so we can CPI into it and call the transfer function to move funds from a user wallet to one of our PDA accounts. We also include the `sysvars_instruction` account so we can determine ordering of instructions in a transaction to try to avoid double charging users who use certain combinations of instructions in the same transaction. An example of this would be candy machine mint transactions which call both `create_metadata` and `update_metadata`.~~

Instead of using a fee treasury account, and passing it in, we are opting to store fees in metadata and edition accounts and harvesting them later with an automated service. The advantage of this is we already have these accounts passed into most instructions, and storing funds in them does not incur any additional write lock penalties, unlike using a single treasury account across multiple instructions. Even using multiple fee treasury accounts may not be sufficient, as a single instruction such as `set_and_verify` could be run thousands of times in a collection migration scenario. Multiple thousand item collections being migrated concurrently could result in an unacceptably severe write lock on the fee treasury account.

~~Adding additional accounts is a breaking change, so the strategy here is to only collect fees on instructions that already have `system_program`, which is only the `create` instructions. We also can collect fees on burn instructions as they are a unique case, but have opted to do this later.~~

~~For all the other instructions we wish to collect fees on, we add in the extra accounts now into the instructions, but do not use them in the actual processor handlers for the instructions. This keeps this on-chain version backwards compatible with other instructions but also gives users new instructions they can migrate to now, and which will be required for the next version when we start processing the instructions in the handlers on-chain.~~

~~Accounts that already have the edition or metadata accounts passed in, but as readonly have these accounts changed to mutable in these new instructions but won't be used mutably on chain until the next version.~~

### Implementation

There's a new source file called `fees.rs` where most of the fee logic is contained. `levy()` calculates the correct fee amount based on the instruction type (`IxType`). Currently only the create instructions are implemented here. All future fee instructions that require a system transfer from a user's wallet should be added here.

Charging fees then, is a simple matter of calling this function within the instruction handler and passing in the correct `IxType` and a few other parameters:

```rust
levy(LevyArgs {
    ix_type: IxType::CreateMetadata,
    payer_account_info,
    receiver_account_info: metadata_account_info,
})?;
```

The `burn_nft` and `burn_print_edition` instructions have also had the additional accounts added for maximum future flexibility, though they shouldn't ever need them.


token-metadata/program/src/instruction/edition.rs
```diff
@@ -122,7 +126,8 @@ pub fn mint_new_edition_from_master_edition_via_token(
         AccountMeta::new_readonly(new_metadata_update_authority, false),
         AccountMeta::new_readonly(metadata, false),
```
febo, line 127: Does the metadata account need to be `writable`?

token-metadata/program/src/instruction/mod.rs
```diff
@@ -195,6 +201,7 @@ pub enum MetadataInstruction {
     #[account(11, name="token_program", desc="Token program")]
     #[account(12, name="system_program", desc="System program")]
     #[account(13, optional, name="rent", desc="Rent info")]
+    #[account(14, name="sysvar_instructions", desc="Instructions sysvar account")]
```
febo, line 204: Adding a comment here to double-check whether the `metadata` needs to be mutable or not. Probably not since the `new_metadata` is mutable.

token-metadata/program/src/instruction/mod.rs
```diff
@@ -297,6 +315,7 @@ pub enum MetadataInstruction {
     #[account(8, name="token_program", desc="Token program")]
     #[account(9, name="system_program", desc="System program")]
     #[account(10, optional, name="rent", desc="Rent info")]
+    #[account(11, name="sysvar_instructions", desc="Instructions sysvar account")]
```
febo, line 318: Does the `metadata` account need to be writable?

token-metadata/program/src/instruction/mod.rs
```diff
@@ -309,6 +328,7 @@ pub enum MetadataInstruction {
     #[account(6, name="token_program", desc="Token program")]
     #[account(7, name="system_program", desc="System program")]
     #[account(8, optional, name="rent", desc="Rent info")]
+    #[account(9, name="sysvar_instructions", desc="Instructions sysvar account")]
```
febo, line 331: Does the `metadata` account need to be writable?

token-metadata/program/src/instruction/mod.rs
```diff
@@ -329,6 +351,7 @@ pub enum MetadataInstruction {
     #[account(5, name="mint", desc="Mint of Collection Metadata")]
     #[account(6, name="system_program", desc="System program")]
     #[account(7, optional, name="rent", desc="Rent info")]
+    #[account(8, name="sysvar_instructions", desc="Instructions sysvar account")]
```
febo, line 354: Does the `metadata` account need to be writable?

token-metadata/program/src/instruction/mod.rs
```diff
@@ -337,6 +360,8 @@ pub enum MetadataInstruction {
     #[account(2, signer, writable, name="revoke_authority", desc="Update Authority, or Delegated Authority, of Collection NFT")]
     #[account(3, name="metadata", desc="Metadata account")]
     #[account(4, name="mint", desc="Mint of Metadata")]
+    #[account(5, name="system_program", desc="System program")]
```
febo, line 363: Does the `metadata` account need to be writable?

token-metadata/program/src/instruction/mod.rs
```diff
@@ -357,6 +384,8 @@ pub enum MetadataInstruction {
     #[account(2, name="edition", desc="Edition")]
     #[account(3, name="mint", desc="Token mint")]
     #[account(4, name="token_program", desc="Token Program")]
+    #[account(5, name="system_program", desc="System program")]
```
febo, line 387: Does the `edition` account need to be writable?

token-metadata/program/src/instruction/mod.rs
```diff
@@ -365,11 +394,15 @@ pub enum MetadataInstruction {
...
```
febo, line 397: Does the `edition` account need to writable?

token-metadata/program/src/utils/fees.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub payer_account_info: &'a AccountInfo<'a>,
+    pub token_metadata_pda_info: &'a AccountInfo<'a>,
+}
+
+pub(crate) fn levy(args: LevyArgs) -> ProgramResult {
+    // Fund metadata account with rent + Metaplex fee.
+    let rent = Rent::get()?;
+
+    let fee = match args.ix_type {
+        IxType::CreateMetadata => CREATE_FEE + rent.minimum_balance(Metadata::size()),
```
febo, line 39: Should there be a check to determine if the rent needs to be included here? Could this add the rent value twice if the account is already initialized?

token-metadata/program/src/processor/uses/utilize.rs
```diff
@@ -30,27 +30,30 @@ pub fn process_utilize(
...
     let metadata_info = next_account_info(account_info_iter)?;
     let token_account_info = next_account_info(account_info_iter)?;
     let mint_info = next_account_info(account_info_iter)?;
     let user_info = next_account_info(account_info_iter)?;
     let owner_info = next_account_info(account_info_iter)?;
     let token_program_account_info = next_account_info(account_info_iter)?;
     let _ata_program_account_info = next_account_info(account_info_iter)?;
-    let _system_account_info = next_account_info(account_info_iter)?;
+    let _system_program_account_info = next_account_info(account_info_iter)?;
+
```
febo, line 42: Is the `sysvars` account missing here?

token-metadata/program/src/error.rs
```diff
@@ -753,6 +753,14 @@ pub enum MetadataError {
     /// 190
     #[error("Missing delegate record")]
     MissingDelegateRecord,
+
+    /// 191
+    #[error("")]
```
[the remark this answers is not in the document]
samuelvanderwaal: We only have 469 bytes of space left. :-(

token-metadata/program/src/processor/uses/utilize.rs
samuelvanderwaal: It is passed in but not read yet. Technically, these two accounts don't need to be read either because they're not used; I'm not sure of the legacy reason for `_ata_program` to be here.

token-metadata/program/src/instruction/mod.rs
samuelvanderwaal: Yes, thanks!

token-metadata/program/src/instruction/mod.rs
samuelvanderwaal: Yes, thanks!

token-metadata/program/src/instruction/mod.rs
samuelvanderwaal: Yes, thanks!

token-metadata/program/src/instruction/mod.rs
samuelvanderwaal: Yes, thanks!

token-metadata/program/src/instruction/mod.rs
samuelvanderwaal: Yes, thanks!

token-metadata/program/src/instruction/mod.rs
samuelvanderwaal: As long as one of the metadata accounts is mutable, we should be covered.

token-metadata/program/src/utils/fees.rs
samuelvanderwaal: Yeah, you're right for `create`. Fixed.

token-metadata/program/src/utils/mod.rs
```diff
@@ -190,18 +193,50 @@ pub fn zero_account(s: &str, size: usize) -> String {
...
+        .ok_or(MetadataError::NumericalOverflowError)?;
+
     // Transfer lamports from the account to the destination account.
     let dest_starting_lamports = funds_dest_account_info.lamports();
-    **funds_dest_account_info.lamports.borrow_mut() = dest_starting_lamports
-        .checked_add(account_info.lamports())
-        .unwrap();
-    **account_info.lamports.borrow_mut() = 0;
+    **funds_dest_account_info.lamports.borrow_mut() =
+        dest_starting_lamports.checked_add(redeem_lamports).unwrap();
```
danenbm, line 230: Nit: for consistency this could also use `ok_or`

danenbm (changes requested):

token-metadata/program/src/utils/mod.rs
samuelvanderwaal: good call

token-metadata/js/test/burn.test.ts
```diff
@@ -37,13 +37,17 @@ test('Burn: NonFungible asset', async (t) => {
...
   const tokenAccount = await connection.getAccountInfo(token);
 
-  t.equal(metadataAccount, null);
-  t.equal(editionAccount, null);
+  if (metadataAccount) {
+    t.equal(metadataAccount.data.length, 0);
+  }
+  if (editionAccount) {
+    t.equal(editionAccount.data.length, 0);
+  }
```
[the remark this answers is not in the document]
samuelvanderwaal: Addressed in [037ac4e8bf0e245f1b9af9aeff65b6446d0d9035](https://github.com/metaplex-foundation/metaplex-program-library/pull/1106/commits/037ac4e8bf0e245f1b9af9aeff65b6446d0d9035).

token-metadata/program/src/state/fee.rs
```diff
@@ -0,0 +1,13 @@
+use super::*;
+
+pub(crate) const FEE_AUTHORITY: Pubkey = pubkey!("Levytx9LLPzAtDJJD7q813Zsm8zg9e1pb53mGxTKpD7");
+
+// base fee level, 0.001 SOL
+pub const BASE_FEE: u64 = 1_000_000;
+
+// create_metadata_accounts
+pub const CREATE_FEE: u64 = 10 * BASE_FEE;
+pub const UPDATE_FEE: u64 = 2 * BASE_FEE;
```
febo, line 10: Do we still need this here?

token-metadata/program/src/utils/mod.rs
```diff
@@ -190,18 +192,42 @@ pub fn zero_account(s: &str, size: usize) -> String {
 pub fn close_program_account<'a>(
```
febo, line 192: Maybe we should have this as `pub(crate)` to avoid it being used outside token metadata – it appears on the suggestion list if you have token metadata as dependency.

token-metadata/program/src/utils/mod.rs
```diff
@@ -190,18 +192,42 @@ pub fn zero_account(s: &str, size: usize) -> String {
...
+    let remaining_lamports = account_info
+        .lamports()
+        .checked_sub(rent_lamports)
+        .ok_or(MetadataError::NumericalOverflowError)?;
+
+    let redeem_lamports = account_info
+        .lamports()
+        .checked_sub(remaining_lamports)
+        .ok_or(MetadataError::NumericalOverflowError)?;
+
```
danenbm, line 216: This version of the code has been simplified since previously when you took into account the value of `Rent(size=0)`.    The logic you have now works as long as any fee we have is greater than `Rent(size=0)`.

I think `remaining_lamports` and `redeem_lamports` no longer need two separate calculations since the second one is not being used with regard to `Rent(size=0)`, so it is algebraically pretty much the same statement as the first calculation:
 ```
remaining_lamports = account_val - Rent(Metadata::size())
redeem_lamports = account_val - remaining_lamports

//substitute
redeem lamports = account_val - (account_val - Rent(Metadata::size()))

redeem lamports = Rent(Metadata::size())
```

I think all you need to with this version of the calculation is get the `remaining_lamports` (which enforces that `account_val - Rent(Metadata::size()) >= 0`).  And then you are just always redeeming the `rent_lamports` to the user.  (`rent_lamports + remaining_lamports = account_val`).

This also highlights the gap initially discussed if `fee < Rent(size=0)` since we are _always_ returning the rent amount without regard to what is left over.

**Review by danenbm — CHANGES_REQUESTED** [review]

File: token-metadata/program/src/instruction/collection.rs [hunk 17]
```diff
@@ -77,6 +78,7 @@ pub fn revoke_collection_authority(
             AccountMeta::new(revoke_authority, true),
             AccountMeta::new_readonly(metadata, false),
             AccountMeta::new_readonly(mint, false),
+            AccountMeta::new_readonly(system_program::ID, false),
```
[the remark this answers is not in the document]
danenbm: You already fixed this

token-metadata/program/src/instruction/mod.rs
```diff
@@ -106,6 +108,8 @@ pub enum MetadataInstruction {
     #[account(0, writable, name="metadata", desc="Metadata key (pda of ['metadata', program id, mint id])")]
     #[account(1, signer, name="owner", desc="Owner on the token account")]
     #[account(2, name="token", desc="Account containing tokens from the metadata's mint")]
+    #[account(3, name="system_program", desc="System program")]
+    #[account(4, name="sysvar_instructions", desc="Instructions sysvar account")]
```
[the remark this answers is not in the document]
danenbm: You already fixed this

token-metadata/program/src/instruction/mod.rs
```diff
@@ -337,6 +342,7 @@ pub enum MetadataInstruction {
     #[account(2, signer, writable, name="revoke_authority", desc="Update Authority, or Delegated Authority, of Collection NFT")]
     #[account(3, name="metadata", desc="Metadata account")]
     #[account(4, name="mint", desc="Mint of Metadata")]
+    #[account(5, name="system_program", desc="System program")]
```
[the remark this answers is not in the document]
danenbm: You already fixed this

token-metadata/program/src/state/fee.rs
samuelvanderwaal: Nope, will fix.

token-metadata/program/src/processor/collection/set_collection_size.rs
```diff
@@ -40,13 +38,14 @@ pub fn set_collection_size(
...
-        let collection_authority_record = next_account_info(account_info_iter)?;
+    let delegated_collection_authority_opt = account_info_iter.next();
+
+    if let Some(record) = delegated_collection_authority_opt {
         assert_has_collection_authority(
             collection_update_authority_account_info,
             &metadata,
             collection_mint_account_info.key,
-            Some(collection_authority_record),
+            Some(record),
```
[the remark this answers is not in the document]
samuelvanderwaal: Yes.

token-metadata/program/tests/utils/digital_asset.rs
```diff
@@ -1262,9 +1263,19 @@ impl DigitalAsset {
...
+        // to the system program immediately, but if they exist they should have a
+        // data length of 0.
+
+        if let Some(account) = md_account {
+            assert_eq!(account.data.len(), 0);
+        }
+
+        if let Some(account) = edition_account {
+            assert_eq!(account.data.len(), 0);
+        }
```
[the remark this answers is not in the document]
samuelvanderwaal: Yeah

token-metadata/program/tests/utils/digital_asset.rs
```diff
@@ -1262,9 +1263,19 @@ impl DigitalAsset {
...
             .await?;
 
-        assert!(md_account.is_none());
+        // Token Metadata accounts may still be open because they are no longer being re-assigned
+        // to the system program immediately, but if they exist they should have a
+        // data length of 0.
+
+        if let Some(account) = md_account {
+            assert_eq!(account.data.len(), 0);
+        }
```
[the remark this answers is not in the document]
samuelvanderwaal: There are two situations here: one where the Metadata account is still open either because it has not been garbage collected yet, or because it still has fees on it. In that case we use the `if let` to see if it has an account data length of 0. In the case where it's been garbage collected it's the `None` case and the else is implicit. The only time to fail is if it exists and the account length isn't 0.

token-metadata/program/tests/utils/digital_asset.rs
```diff
@@ -1275,13 +1286,15 @@ impl DigitalAsset {
     ) -> Result<(), BanksClientError> {
         self.non_fungigble_accounts_closed(context).await?;
 
-        // Token record is burned.
+        // Token record is burned. Account may still be open but data length should be 0.
```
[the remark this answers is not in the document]
samuelvanderwaal: The Token Record should always be burned now; only Metadata accounts might be left open, so you're right.

token-metadata/program/src/processor/fee/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    let metadata_rent = rent.minimum_balance(MAX_METADATA_LEN);
+    let master_edition_rent = rent.minimum_balance(MAX_MASTER_EDITION_LEN);
+
+    let (fee_amount, rent_amount) = match account_key {
+        Key::Uninitialized => {
+            account_info.assign(&system_program::ID);
+
+            (account_info.lamports(), 0)
+        }
+        Key::MasterEditionV2 | Key::MasterEditionV1 => {
```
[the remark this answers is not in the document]
samuelvanderwaal: Removed.

token-metadata/program/src/processor/fee/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    let dest_starting_lamports = dest_info.lamports();
+    **dest_info.lamports.borrow_mut() = dest_starting_lamports
+        .checked_add(fee_amount)
+        .ok_or(MetadataError::NumericalOverflowError)?;
+    **account_info.lamports.borrow_mut() = rent_amount;
+
+    // Clear fee flag.
+    clear_fee_flag(account_info, account_key)?;
```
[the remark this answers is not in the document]
samuelvanderwaal: It only clears the flag if it finds it at the index.

token-metadata/program/tests/utils/digital_asset.rs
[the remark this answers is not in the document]
danenbm: Ok that is fine.  I thought all the nonfungible accounts used in tests would now have uncollected fees on them from `create` and remain open since the tests don't remove fees.

danenbm (approved):

PR #1113, opened 2023-06-09
Add Bubblegum transfer and burn js tests

danenbm:
Added a test to transfer and burn a compressed NFT.

One thing I'm concerned about is this warning I get:
```
Transaction references a signature that is unnecessary, only the fee payer and instruction signer accounts should sign a transaction. This behavior is deprecated and will throw an error in the next major version release.
```

I believe this is because Burn does not have the leaf owner listed as a signer in the Accounts validation struct, but checks that either the owner or the delegate is a signer at runtime.  So then the owner and delegate are not listed as signers in the generated JS code.  But this seems like it might cause issues soon.

_UPDATE_: We will fix the above with Kinobi/Umi in new repo.

### DIFF
--- a/bubblegum/js/tests/main.test.ts
+++ b/bubblegum/js/tests/main.test.ts
@@ -20,13 +20,15 @@ import {
 import {
   createCreateTreeInstruction,
   createMintV1Instruction,
+  createTransferInstruction,
+  createBurnInstruction,
   MetadataArgs,
   PROGRAM_ID as BUBBLEGUM_PROGRAM_ID,
   TokenProgramVersion,
   TokenStandard,
   Creator,
 } from '../src/generated';
-import { getLeafAssetId, computeCompressedNFTHash } from '../src/mpl-bubblegum';
+import { getLeafAssetId, computeDataHash, computeCreatorHash, computeCompressedNFTHash } from '../src/mpl-bubblegum';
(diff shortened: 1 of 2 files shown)


bubblegum/js/tests/main.test.ts
```diff
@@ -204,6 +206,72 @@ describe('Bubblegum tests', () => {
...
+          leafOwner: newLeafOwner,
+          leafDelegate: newLeafOwner,
+          merkleTree,
+          logWrapper: SPL_NOOP_PROGRAM_ID,
+          compressionProgram: SPL_ACCOUNT_COMPRESSION_PROGRAM_ID,
+        },
+        {
+          root: Array.from(account.getCurrentRoot()),
+          dataHash: Array.from(computeDataHash(originalCompressedNFT)),
+          creatorHash: Array.from(computeCreatorHash(originalCompressedNFT.creators)),
```
danenbm, line 259: Please review this, it works but is it the best way to specify these params?

lorisleiva: This looks good to me. What are you concerned about with these array parameters?

danenbm:
Closing this in favor of https://github.com/metaplex-foundation/metaplex-program-library/pull/1115

PR #1116, opened 2023-06-14
Adding fees and fixing merge errors.

blockiosaurus:


samuelvanderwaal (approved):

token-metadata/program/src/processor/metadata/print.rs
```diff
@@ -172,5 +179,6 @@ fn print_v1(_program_id: &Pubkey, ctx: Context<Print>, args: PrintArgs) -> Progr
         )?;
     }
 
-    Ok(())
+    // Set fee flag after metadata account is created.
```
febo, line 182: We might need to do this as well:

https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/processor/metadata/create.rs#L176-L188

febo: This is for instructions that do not take the metadata account to identify the token standard.

febo (approved):

PR #1117, opened 2023-06-15
remove migrate instruction

samuelvanderwaal:
Removes the `migrate` instruction and frees up 11,288 bytes of space. This is a permissioned handler used only for the temporary pNFT migration--there shouldn't be any external users of it.


blockiosaurus (approved):
Yay, get it outta here!

danenbm (changes requested):
Need to regenerate JS SDK

danenbm (approved):

blockiosaurus (approved):

PR #1123, opened 2023-06-21
Reverting new leaf schema events for `burn`, `redeem`, and `decompress`

danenbm:
This code was added to Bubblegum to make Read API indexing of compression easier.  But instead of using these leaf schema events, we were able to derive the `asset_id` using the leaf index from the change log event.

I think it would have been more convenient to use these leaf events, but by solving it through derivation in the Read API, it keeps the Read API backwards compatible with all existing cNFT transactions that don't provide the leaf schema events in `burn`, `redeem`, and `decompress`.

Note this code was never deployed to devnet or mainnet.

### DIFF
--- a/bubblegum/program/src/lib.rs
+++ b/bubblegum/program/src/lib.rs
@@ -1322,11 +1322,6 @@ pub mod bubblegum {
             creator_hash,
         );
 
-        wrap_application_data_v1(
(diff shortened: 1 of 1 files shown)


febo (approved):
LGTM

samuelvanderwaal (approved):

blockiosaurus (approved):
Magnifique

PR #1124, opened 2023-06-26
Adding allowance for hydra owned member wallets.

blockiosaurus:
Based on https://github.com/GlassEaters/hydra/commit/1bd0b6d4bee88e0867f55cf67212e4628d9200cd it seems like an oversight that the wallet check doesn't allowed hydra owned members in distribute, but does in add_member. Fixing this.


febo (approved):
LGTM

nhanphan (approved):

PR #1126, opened 2023-06-26
fix next edition calculation for burned items

samuelvanderwaal:
Burning an edition decreases the supply, which breaks the way Fixed Price Sale currently calculates the next edition number. this PR uses the passed in edition marker and finds the first available edition on that particular marker. It has to figure out which edition marker is passed in, in order to determine the correct edition number, so must do a loop deriving the edition marker PDAs and comparing them against what is passed in. For large edition numbers this might use up available compute units.

In addition, it adds a `buy_v2` that requires passing in the edition marker number in edition to the edition marker PDA. This allows a single derivation to be made to calculate the edition number, but would be a breaking change if applied to `buy`. Since the accounts are identical you continue to use the `Buy` account struct and only use the `BuyV2` instruction struct for the additional argument. This allows reducing code duplication.


fixed-price-sale/program/src/processor/buy.rs
```diff
@@ -45,10 +42,36 @@ impl<'info> Buy<'info> {
...
+
+            edition += 248;
+        }
+
+        let is_first_marker = edition == 0;
+
+        // Find the first available edition number in this edition marker.
+        if edition_marker_info.data_is_empty() {
+            // First Edition marker skips the first bit because editions start at 1.
+            if is_first_marker {
```
blockiosaurus, line 63: Shouldn't this also account for being the first bit in a new marker?

blockiosaurus: Nevermind, I'm wrong.

blockiosaurus (dismissed):

blockiosaurus (dismissed):

blockiosaurus (approved):

fixed-price-sale/program/src/processor/buy.rs
```diff
@@ -45,10 +42,97 @@ impl<'info> Buy<'info> {
...
+            .0;
+
+            if pda != *edition_marker_info.key {
+                return Err(ErrorCode::InvalidEditionMarkerAccount.into());
+            }
+
+            edition_marker_number
+        } else {
+            let mut edition_marker_number = 0u64;
+            let edition_marker_number_str = edition_marker_number.to_string();
```
febo, line 73: The value of `edition_marker_number_str` is not updated in the loop, which I think it should be since it is used in the seeds.

fixed-price-sale/program/src/processor/buy.rs
```diff
@@ -45,10 +42,97 @@ impl<'info> Buy<'info> {
...
+
+            loop {
+                let pda = Pubkey::find_program_address(
+                    &[
+                        "metadata".as_bytes(),
+                        mpl_token_metadata::ID.as_ref(),
+                        metadata_mint.as_ref(),
+                        "edition".as_bytes(),
+                        edition_marker_number_str.as_bytes(),
```
febo, line 82: This is using the same value in every iteration of the loop.

samuelvanderwaal: Good catch! I'm ignoring the long-running tests because the CI machines can't handle minting that many editions which is why the CI passed with this issue. Maybe I can use account injection to write some tests to check for multiple PDAs instead.

blockiosaurus (approved):

febo (approved):

PR #1133, opened 2023-06-28
Fix: set_token_standard() has update_authority unneccessarily set to mutable

billythedummy:
This results in writable privilege escalation for e.g. when CPI-ing `SetTokenStandard` using anchor_spl with an immutable update_authority AccountInfo

### DIFF
--- a/token-metadata/js/idl/mpl_token_metadata.json
+++ b/token-metadata/js/idl/mpl_token_metadata.json
@@ -1932,7 +1932,7 @@
         },
         {
           "name": "updateAuthority",
-          "isMut": true,
+          "isMut": false,
           "isSigner": true,
           "desc": "Metadata update authority"
         },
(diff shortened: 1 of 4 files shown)


samuelvanderwaal:
Thanks for the fix! You also need to remove the `writable` annotation from `SetTokenStandard`'s `update_authority` account in `src/instruction/mod.rs` so the autogenerated JS lib also has this change. After that run `yarn api:gen` in the JS directory and push and then it should be good to be merged.

billythedummy:
> Thanks for the fix! You also need to remove the `writable` annotation from `SetTokenStandard`'s `update_authority` account in `src/instruction/mod.rs` so the autogenerated JS lib also has this change. After that run `yarn api:gen` in the JS directory and push and then it should be good to be merged.

done

blockiosaurus (approved):
Looks good, thank you for catching this!

danenbm (approved):
LGTM

PR #1139, opened 2023-07-05
remove redundant check

samuelvanderwaal:
Removes a [redundant check](https://github.com/metaplex-foundation/metaplex-program-library/blob/d6f96f67b187104e6c4622b99f466bc530fc2b7d/bubblegum/program/src/lib.rs#L737C9-L737C9) that is already present in Bubblegum. As long as this handler is permissioned to Bubblegum this check only needs to be in Bubblegum. This assertion is[ getting reworked in Bubblegum to support both delegate types ](https://github.com/metaplex-foundation/metaplex-program-library/pull/1138) and it's not necessary to add the extra code to this assert as well--better to remove it to save binary size.

### DIFF
--- a/token-metadata/program/src/processor/bubblegum/bubblegum_set_collection_size.rs
+++ b/token-metadata/program/src/processor/bubblegum/bubblegum_set_collection_size.rs
@@ -8,7 +8,7 @@ use solana_program::{
 };
 
 use crate::{
-    assertions::{assert_owned_by, collection::assert_has_collection_authority},
+    assertions::assert_owned_by,
     error::MetadataError,
     instruction::SetCollectionSizeArgs,
     state::{CollectionDetails, Metadata, TokenMetadataAccount},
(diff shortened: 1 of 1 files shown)


token-metadata/program/src/processor/bubblegum/bubblegum_set_collection_size.rs
```diff
@@ -29,12 +29,6 @@ pub fn bubblegum_set_collection_size(
...
 
-    let delegated_collection_auth_opt = if accounts.len() == 5 {
-        Some(next_account_info(account_info_iter)?)
-    } else {
-        None
-    };
-
     if !BUBBLEGUM_ACTIVATED {
         return Err(MetadataError::InvalidOperation.into());
     }
```
danenbm, line 34: Should we add a comment indicating that collection authority is checked in Bubblegum or is that obvious?

danenbm (approved):
LGTM

blockiosaurus (approved):
Looks good!

PR #1148, opened 2023-07-14
set burned metadata accounts with fees as Unitialized

samuelvanderwaal:
Currently, when burn handlers are called, if a metadata account has fees on it, the data length is reallocated to 0 but the ownership is not reassigned so that fees can be collected on the account. This PR changes the behavior to set the data length to 1 in this case and set that byte to 0 so that the discriminator is `Uninitialized`. This allows indexers to actually categorize and index these accounts until they are fully closed and garbage collected.

Note on the implementation:

```rust
 if collection_metadata_info.data_is_empty()
            || collection_metadata_info.data.borrow()[0] == 0
```

Direct indexing panics if the index does not exist, but in this case it is ok to do because the if statement will short-circuit if the data is empty.

### DIFF
--- a/token-metadata/js/test/burn.test.ts
+++ b/token-metadata/js/test/burn.test.ts
@@ -37,12 +37,14 @@ test.only('Burn: NonFungible asset', async (t) => {
 
   await updateTx.assertSuccess(t);
 
-  // All three accounts are closed. Metadata account should have a data length of 0 but may be open if it contains fees.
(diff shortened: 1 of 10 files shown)


blockiosaurus (approved):
These changes look good!

blockiosaurus (approved):

febo (approved):
LGTM

PR #1163, opened 2023-11-11
feature: `mpl-utils`: Log assert failures for easier debugging

KartikSoneji:
At the moment, when an assertion fails, the program only throws an error about the type of assertion that has failed.
This there no indication to _which_ account is the incorrect one, making it a nightmare to debug.

[For example](https://explorer.solana.com/tx/4deWJFAux67HLu4k1EtmoBovt31zRcKhHSHD1QB2xf77PPA7KrvBiQrRV56REz1fXtNYQR5MpZbYAM7jiJfW93my):
```console
> Program invoked: Token Metadata Program
  > Program logged: "IX: Delegate"
  > Program logged: "Public key does not match expected value"
  > Program consumed: 18426 of 373338 compute units
  > Program returned error: "custom program error: 0x9a"
```

This PR adds logs that print the account that failed the assertion and some debug information about why the assertion failed.
They only trigger if the assertion fails, so shouldn't effect the compute unit budget during normal operation.

### DIFF
--- a/core/rust/utils/src/assertions.rs
+++ b/core/rust/utils/src/assertions.rs
@@ -63,7 +63,7 @@ pub fn assert_derivation(
     let (key, bump) = Pubkey::find_program_address(path, program_id);
     if key != *account.key {
         msg!(
-            "derivation assertion failed for {actual_key}:\n
+            "derivation assertion failed for {actual_key}:
                 \x20   expected {key} with program {program_id}, path {path:?}",
             actual_key = account.key,
             key = key,
--- a/core/rust/utils/src/assertions.rs
+++ b/core/rust/utils/src/assertions.rs
@@ -1,7 +1,7 @@
 use solana_program::{
     account_info::AccountInfo,
     entrypoint::ProgramResult,
-    log::sol_log,
+    msg,
     program_error::ProgramError,
     program_pack::{IsInitialized, Pack},
     pubkey::Pubkey,
@@ -10,12 +10,9 @@ use solana_program::{
 
 pub fn assert_signer(account_info: &AccountInfo) -> ProgramResult {
     if !account_info.is_signer {
-        sol_log(
-            format!(
-                "signer assertion failed for {key}: not signer",
-                key = account_info.key,
-            )
-            .as_str(),
+        msg!(
+            "signer assertion failed for {key}: not signer",
+            key = account_info.key,
         );
         Err(ProgramError::MissingRequiredSignature)
     } else {
@@ -29,12 +26,9 @@ pub fn assert_initialized<T: Pack + IsInitialized>(
 ) -> Result<T, ProgramError> {
     let account: T = T::unpack_unchecked(&account_info.data.borrow())?;
     if !account.is_initialized() {
-        sol_log(
-            format!(
-                "initialized assertion failed for {key}: not initialized",
-                key = account_info.key,
-            )
-            .as_str(),
+        msg!(
+            "initialized assertion failed for {key}: not initialized",
+            key = account_info.key,
         );
         Err(error.into())
     } else {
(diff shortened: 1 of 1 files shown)


blockiosaurus:
Thanks for this PR! It is indeed awful to debug any of these issues. The lack of logging of specific failures is unfortunately a necessity due to program size and compute constraints.

However, ExtendProgram is scheduled to be enabled Monday so that will hopefully solve the size issue.

KartikSoneji:
Oh does calling `sol_log()` really increase the program size so much?
Or is it `format!()`? I can refactor to remove calls to format if that's the issue.

blockiosaurus:
It's the actual strings. I think we have more space now but we were a handful of bytes from the max a couple months ago.

KartikSoneji:
Ah I see.
Let me know if the messages are too long, I think I can cut it down to just a few extra bytes.
Eg: `<actual key>!=<expected key>`
Even terse log messages like this should be a great help while debugging.

KartikSoneji:
Hi, just a friendly ping on this.

blockiosaurus:
Huzzah, we have space now! Can you just change these to use `msg!` to be consistent with our other programs? That should also be a little cleaner as it uses the same string formatting as `println!`.

KartikSoneji:
Done!

core/rust/utils/src/assertions.rs
```diff
@@ -9,6 +10,10 @@ use solana_program::{
 
 pub fn assert_signer(account_info: &AccountInfo) -> ProgramResult {
     if !account_info.is_signer {
+        msg!(
+            "signer assertion failed for {key}: not signer",
```
febo, line 14: "{key} is not a signer"

core/rust/utils/src/assertions.rs
```diff
@@ -21,6 +26,10 @@ pub fn assert_initialized<T: Pack + IsInitialized>(
 ) -> Result<T, ProgramError> {
     let account: T = T::unpack_unchecked(&account_info.data.borrow())?;
     if !account.is_initialized() {
+        msg!(
+            "initialized assertion failed for {key}: not initialized",
```
febo, line 30: {key} is not initialized

core/rust/utils/src/assertions.rs
```diff
@@ -33,6 +42,12 @@ pub fn assert_owned_by(
     error: impl Into<ProgramError>,
 ) -> ProgramResult {
     if account.owner != owner {
+        msg!(
+            "owner assertion failed for {key}: expected {expected}, got {actual}",
```
febo, line 46: "invalid owner for {key} ({actual} != {expected})"

core/rust/utils/src/assertions.rs
```diff
@@ -47,6 +62,14 @@ pub fn assert_derivation(
 ) -> Result<u8, ProgramError> {
     let (key, bump) = Pubkey::find_program_address(path, program_id);
     if key != *account.key {
+        msg!(
+            "derivation assertion failed for {actual_key}:\n
```
febo, line 66: "invalid derivation {actual_key} != {key}"

core/rust/utils/src/assertions.rs
```diff
@@ -58,6 +81,12 @@ pub fn assert_rent_exempt(
     error: impl Into<ProgramError>,
 ) -> ProgramResult {
     if !rent.is_exempt(account_info.lamports(), account_info.data_len()) {
+        msg!(
+                "rent exempt assertion failed for {key}: has {balance} lamports, requires at least {min} lamports",
```
febo, line 85: "rent exempt failed for {key} ({min} lamports required)"

febo (changes requested):
What do you think if we make the log messages a bit shorter so the log does not get cluttered?

core/rust/utils/src/assertions.rs
KartikSoneji: Looks like `\n` is not needed, I'll remove that.

febo:
> Hmm I really think the log messages should provide as much information to debug as possible. Plus I think in most cases programs failing an assertion is a fatal error, so this should be the last message that is in the log.

I agree that log messages should be descriptive, but the observation here is because this is a library not a program. The program might handle an assert failure so it does not end up being a "fatal error".

KartikSoneji:
OK, I'll shorten the messages for `assert_signer`, `assert_initialized` and `assert_rent_exempt`.
But I think `assert_owned_by`, `assert_derivation` are almost always fatal errors so should clearly communicate the issue without having to refer back to the source code of the program.

PR #1173, opened 2024-02-29
Fix money returns caused by an account shrinking

kstepanovdev:
added the possibility to return lamports after an account has been shrunk

### DIFF
--- a/core/rust/utils/src/account.rs
+++ b/core/rust/utils/src/account.rs
@@ -65,17 +65,28 @@ pub fn resize_or_reallocate_account_raw<'a>(
 ) -> ProgramResult {
     let rent = Rent::get()?;
     let new_minimum_balance = rent.minimum_balance(new_size);
+    let current_ta_lamports = target_account.lamports();
+    let account_infos = &[
+        funding_account.clone(),
+        target_account.clone(),
+        system_program.clone(),
+    ];
 
-    let lamports_diff = new_minimum_balance.saturating_sub(target_account.lamports());
-    invoke(
-        &system_instruction::transfer(funding_account.key, target_account.key, lamports_diff),
-        &[
-            funding_account.clone(),
-            target_account.clone(),
-            system_program.clone(),
-        ],
-    )?;
-
+    // account will be shrunk
+    if target_account.data_len() > new_size {
+        let lamports_diff = new_minimum_balance.saturating_sub(current_ta_lamports);
+        invoke(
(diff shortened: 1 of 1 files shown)


nhanphan (approved):

nhanphan (changes requested):
when shrinking the target account, it may or may not be owned by the system program so the transfer may not work

blockiosaurus:
You'll want to do the system program shrink transfer if the account is system program owned but manual lamport math otherwise.

Welcome to the confusing world of Solana programming 😁

core/rust/utils/src/account.rs
```diff
@@ -65,17 +65,28 @@ pub fn resize_or_reallocate_account_raw<'a>(
...
-        &[
-            funding_account.clone(),
-            target_account.clone(),
-            system_program.clone(),
-        ],
-    )?;
-
+    // account will be shrunk
+    if target_account.data_len() > new_size {
+        let lamports_diff = new_minimum_balance.saturating_sub(current_ta_lamports);
```
danenbm, line 77: I think this if statement is reversed.  Example: target account data len is 200 bytes.  New size is 20 bytes.
New minimum balance will be Rent(20).  Current lamports is Rent(200) which will be bigger than Rent(20).  Therefore lamports diff would be 0.

Looks to me like L84-L88 was meant to be in the `if` statement instead of the `else`.

danenbm (changes requested):

core/rust/utils/src/account.rs
kstepanovdev: Good catch. Idk how I missed that.

core/rust/utils/src/account.rs
```diff
@@ -65,20 +66,38 @@ pub fn resize_or_reallocate_account_raw<'a>(
...
+        system_program.clone(),
+    ];
 
-    target_account.realloc(new_size, false)?;
+    if new_size > target_account.data_len() {
+        invoke(
+            &system_instruction::transfer(funding_account.key, target_account.key, lamports_diff),
+            account_infos,
+        )?;
+    } else if target_account.owner == system_program.owner {
```
blockiosaurus, line 86: This should be system_program.key instead of owner.

blockiosaurus (changes requested):
Left one comment but otherwise this is good!

blockiosaurus (approved):

blockiosaurus (approved):

nhanphan (approved):

danenbm (approved):
LGTM!

PR #1187, opened 2024-10-28
Following best practices for closing accounts

blockiosaurus:


danenbm (approved):

core/rust/utils/src/account.rs
```diff
@@ -111,8 +110,6 @@ pub fn close_account_raw<'a>(
         .unwrap();
     **src_account_info.lamports.borrow_mut() = 0;
```
danenbm, line 111: Nit: if you had time the panicking functions could be replaced:

replace borrow_mut() with try_borrow_mut() and
remove the unwrap()

danenbm (dismissed):
LGTM but left one comment for potential to improve error handling but not required

core/rust/utils/src/account.rs
blockiosaurus: Good call. I couldn't change to a more specific error for numerical overflow without changing the function signature so I picked InvalidRealloc. Maybe in a major version we can add an error argument.

danenbm: ah yeah, i think ProgramError ok too for this lib but yeah can change in the future if we like.  Thank you for the fix!

danenbm (approved):

